PR #1, opened 2022-03-25
Request to rename/avoid MicroState

IAlibay:
Microstates is a stat mech loaded term - asking as someone that still has to deal with profs regularly, can we just avoid it 🙃 

Here I propose ChemicalState. The idea being that ChemicalState(s) are real states that are joined by various AlchemicalStates along a free energy network.

That being said, if the intent is to truly represent a very specific microstate of a system (which didn't seem the case from this week's F@H alchemy call), then that's a different issue...


dotsdl (approved):
I love this name choice, actually! Thank you @IAlibay!

PR #2, opened 2022-03-25
solvent implementation

richardjgowers:


gufe/solventcomponent.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, smiles='O', ions=None, concentration=None):
+        """
+        Parameters
+        ----------
+        smiles : str, optional
+          smiles of the solvent, default 'O' (water)
+        ions : list of str, optional
+          ions in the system, default None
+        concentration : float, optional
+          ionic concentration required, default None
```
IAlibay, line 23: Define units please.

IAlibay (changes requested):
Some quick changes, but for the most part this looks sufficient to me. Just need to make sure we don't buy into a specific way of doings things where possible.

gufe/solventcomponent.py
```diff
@@ -0,0 +1,88 @@
...
+                 ions: Tuple[str, ...] = None,
+                 neutralize: bool = True,
+                 concentration: unit.Quantity = None):
+        """
+        Parameters
+        ----------
+        smiles : str, optional
+          smiles of the solvent, default 'O' (water)
+        ions : list of str, optional
+          ions in the system, default `None`
```
dwhswenson, line 29: Is this a list of all individual ions, or is this expected to be one positive and one negative? (If so, maybe have `position_ion` and `negative_ion`?) Taking this as a list is confusing me, because I'm trying to figure out what list length 1 or list length 3 might mean.

gufe/solventcomponent.py
```diff
@@ -0,0 +1,88 @@
...
+        neutralize : bool, optional
+          if the net charge on the chemical state is neutralized by the ions in this
+          solvent component.  Default `True`
+        concentration : openff-units.unit.Quantity, optional
```
dwhswenson, line 33: Is this concentration in excess of the neutralizing ions? If not, then what is it concentration of?

dwhswenson (commented):
Sorry to review this after merge, but there are some points I'm unclear on as I try to figure out how to use this. I think some quick docstring additions will clear it up for me.

PR #3, opened 2022-03-25
Added initial attributes and methods for ChemicalState

dotsdl:
Included `__init__` for core set of attributes for `ChemicalState` from working session with @richardjgowers. Assumes that `ChemicalState` components are all RDKit molecules, and that that the `ChemicalState` is immutable upon instantiation.

### DIFF
--- a/gufe/chemicalstate.py
+++ b/gufe/chemicalstate.py
@@ -1,7 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
-from typing import Dict
+from typing import Dict, Optional
 
 import numpy as np
 from frozendict import frozendict
@@ -31,8 +31,8 @@ class ChemicalState:
     def __init__(
             self,
             components: Dict[str, RDKitMol],
-            box_vectors: np.ndarray,
-            identifier: str = None,
+            box_vectors: Optional[np.ndarray] = None,
+            identifier: Optional[str] = None,
             ):
         """Create a node for an alchemical network.
 
@@ -52,25 +52,36 @@ class ChemicalState:
             `AlchemicalNetwork`
 
         """
-        self.components = frozendict(components)
-        self.box_vectors = box_vectors
-        self.identifier = identifier
+        self._components = frozendict(components)
+        self._box_vectors = box_vectors
+        self._identifier = identifier
 
     def __hash__(self):
-        return hash((self.components, self.identifier))
+        return hash((self._components, self._identifier, self._box_vectors))
+
+    @property
+    def components(self):
+        return self._components
+
+    @property
+    def box_vectors(self):
+        return np.array(self._box_vectors)
+
+    @property
+    def identifier(self):
+        return self._identifier
+
+    @classmethod
+    def as_protein_ligand_solvent(cls):
+        """
+
+        """
+        ...
 
 
-class ProteinLigandSolventMicrostate(ChemicalState):
-    """
-
-    """
-    ...
-
-
-class LigandSolvent(ChemicalState):
-    """
-
-    """
-    ...
-
-    pass
+    @classmethod
+    def as_ligand_solvent(cls):
+        """
+    
+        """
+        ...
--- a/gufe/chemicalstate.py
+++ b/gufe/chemicalstate.py
@@ -53,8 +53,12 @@ class ChemicalState:
 
         """
         self._components = frozendict(components)
-        self._box_vectors = box_vectors
         self._identifier = identifier
+
+        if box_vectors is None:
+            self._box_vectors = np.array([np.nan]*9)
+        else:
+            self._box_vectors = box_vectors
 
     def __hash__(self):
         return hash((self._components, self._identifier, self._box_vectors))
--- a/gufe/chemicalstate.py
+++ b/gufe/chemicalstate.py
@@ -4,11 +4,12 @@
 from typing import Dict, Optional
 
 import numpy as np
+from openff.toolkit.utils.serialization import Serializable
 
 from .component import Component
 
 
-class ChemicalState:
+class ChemicalState(Serializable):
     """A node of an alchemical network.
 
     Attributes
@@ -62,10 +63,29 @@ class ChemicalState:
     def __hash__(self):
         return hash(
             (
-                tuple(self._components.items()),
+                tuple(sorted(self._components.items())),
                 self._box_vectors,
                 self._identifier,
             )
+        )
+
+    def to_dict(self):
+        return {
+            "components": {
+                key: value.to_dict() for key, value in self.components.items()
+            },
+            "box_vectors": self.box_vectors.tolist(),
+            "identifier": self.identifier,
+        }
+
+    @classmethod
+    def from_dict(cls, d):
+        return cls(
+            components={
+                key: Component.from_dict(value) for key, value in d["components"]
+            },
+            box_vectors=np.array(d["box_vectors"]),
+            identifier=d["identifier"],
         )
 
     @property
--- a/gufe/chemicalstate.py
+++ b/gufe/chemicalstate.py
@@ -16,12 +16,12 @@ class ChemicalState:
     Attributes
     ----------
     components
-        The molecular representation of the chemical state, including connectivity and
-        coordinates. This is a frozendict with user-defined labels as keys, RDKit
-        molecules as values.
+        The molecular representation of the chemical state, including
+        connectivity and coordinates. This is a frozendict with user-defined
+        labels as keys, RDKit molecules as values.
     box_vectors
-        Numpy array indicating shape and size of unit cell for the system.
-        May be a partial definition to allow for variability on certain dimensions.
+        Numpy array indicating shape and size of unit cell for the system. May
+        be a partial definition to allow for variability on certain dimensions.
     identifier
         Optional identifier for the chemical state; used as part of the
         (hashable) graph node itself when the chemical state is added to an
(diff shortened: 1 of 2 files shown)


gufe/chemicalstate.py
```diff
@@ -1,6 +1,77 @@
...
+        """
+        self.components = frozendict(components)
+        self.box_vectors = box_vectors
+        self.identifier = identifier
+
+    def __hash__(self):
+        return hash((self.components, self.identifier))
+
+
+class ProteinLigandSolventMicrostate(ChemicalState):
```
IAlibay, line 64: How about calling this ComplexState and then the alternative LigandState?

I think the idea was that gas phase is just a special case of Vacuum solvent, but that will probably need more discussions.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+    identifier
+        Optional identifier for the chemical state; used as part of the
+        (hashable) graph node itself when the chemical state is added to an
+        `AlchemicalNetwork`
+
+    """
+
+    def __init__(
+            self,
+            components: Dict[str, RDKitMol],
```
richardjgowers, line 33: This is nearly what we had before but with a couple of differences. By having the key (hash) of the molecule separate there's a chance it'll get misplaced, and/or you'll end up passing around the combination of (label, Molecule) together; so we ended up lumping them together in the LigandComponent class. This has a side effect that you don't have to commit to any particular underlying representation, it's currently rdkit but once OFF is stable we could migrate if it was easier. Then it also makes it easier to attach a to_dict method to each *Component and everything is trivially serialisable. So I fought for a while to not reimplement Molecule but that's why we did in the end. Thoughts @dotsdl ?

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+        identifier
+            Optional identifier for the chemical state; used as part of the
+            (hashable) graph node itself when the chemical state is added to an
+            `AlchemicalNetwork`
+
+        """
+        self.components = frozendict(components)
+        self.box_vectors = box_vectors
+        self.identifier = identifier
+
```
richardjgowers, line 58: Could lift the OFF serializable mixin and define to_dict here.

gufe/chemicalstate.py
richardjgowers: I'd just make these classmethods which are a guided way (& sanity checks) to make the class, unless there's some methods that only this child class will have.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+            May be a partial definition to allow for variability on certain
+            dimensions.
+        identifier
+            Optional identifier for the chemical state; used as part of the
+            (hashable) graph node itself when the chemical state is added to an
+            `AlchemicalNetwork`
+
+        """
+        self.components = frozendict(components)
+        self.box_vectors = box_vectors
```
richardjgowers, line 56: I'd have box_vectors be optional, then if `None` make a box of `np.array([np.nan]*9])`.  Where we're using `np.nan` as an sentinel empty value indicator.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+            Numpy array indicating shape and size of unit cell for the system.
+            May be a partial definition to allow for variability on certain
+            dimensions.
+        identifier
+            Optional identifier for the chemical state; used as part of the
+            (hashable) graph node itself when the chemical state is added to an
+            `AlchemicalNetwork`
+
+        """
+        self.components = frozendict(components)
```
richardjgowers, line 55: This class needs to be immutable, so we might have to have these all as `self._attribute` and have a property accessor that doesn't implement a `setter`.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+        self.box_vectors = box_vectors
+        self.identifier = identifier
+
+    def __hash__(self):
+        return hash((self.components, self.identifier))
```
IAlibay, line 60: Just a reminder here, a hash on the protein will need to return non-standard residues (migrating this discussion from the main openfe repo)

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+            Optional identifier for the chemical state; used as part of the
+            (hashable) graph node itself when the chemical state is added to an
+            `AlchemicalNetwork`
+
+        """
+        self.components = frozendict(components)
+        self.box_vectors = box_vectors
+        self.identifier = identifier
+
+    def __hash__(self):
```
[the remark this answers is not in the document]
IAlibay: So there's maybe a point to be made here that box size is more of a method thing than a state thing. I'm thinking the case of an enthalpy calculation where you end up with a big box for your ligand (because that's how the method works), but you might want to use the same set of nodes that you use for an ABFE?

Possibly worth discussing further.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,76 @@
...
+        Optional identifier for the chemical state; used as part of the
+        (hashable) graph node itself when the chemical state is added to an
+        `AlchemicalNetwork`
+
+    """
+
+    def __init__(
+            self,
+            components: Dict[str, RDKitMol],
+            box_vectors: np.ndarray,
```
IAlibay, line 34: So one of the discussions we had on the openfe side of things is that "box" might want to be optional, particularly accounting for the following cases:

1. Vacuum boxes
2. Boxes which need to be generated from other Boxes (i.e. mol A to mol B RBFEs essentially need to have the same number of waters / boxes, which may be more / fewer than mol B to mol C).

There's probably special considerations for membranes that I'm not thinking of right now.

gufe/chemicalstate.py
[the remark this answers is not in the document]
dotsdl: Agree with @IAlibay; is there a compelling case for the box vectors being part of the chemical state definition, or does this fall more on the side of the method being applied? 

I think we had a discussion where box vectors were hard to separate from a chemical system itself for the case of e.g. a protein embedded in a membrane, so it may be crucial for some systems but less so for others.

Perhaps a policy of: providing `box_vectors` is discretionary, but it will be part of the node hash and a transformation between two nodes will use `box_vectors` as defined and where possible in executing the protocol.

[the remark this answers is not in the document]
IAlibay: Sounds reasonable to me - do we need to go further here and allow for users to explicitly define "frozen states"? I.e. cases where you absolutely cannot deviate from a state definition. In which case that takes the highest priority in defining neighbouring transformations.

It does mean that network solvers will need to trace out all available paths, giving priority to pre-defined states, but I don't think we can avoid that anyways (since we'll need to make sure that all alchemical states can be created prior to launching a campaign).

gufe/chemicalstate.py
dotsdl: From today's call: we decided to proceed with the `dict` input with user-defined keys as labels for now, as these *may* be needed by an assigned edge protocol between two `ChemicalState`s to determine which components in the states map to one another (it may be possible for the protocol to discern this based on which components are different between the two states, but this may get complicated quickly for some protocols / systems). We may drop this `dict` down to a `tuple` of `Component`s later if we deem keys unnecessary here.

gufe/chemicalstate.py
```diff
@@ -1,6 +1,98 @@
...
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
+from typing import Dict, Optional
+
+import numpy as np
+
+from .component import Component
+
+
 class ChemicalState:
```
jchodera, line 11: This class represents a particular composition of components, so names like `ChemicalSystem`, `ComponentCollection`, or `ChemicalComponents` might be more appropriate. This is not a strong recommendation, however.

It's worth noting that this is very similar to the [`openff.toolkit.topology.Topology`](https://open-forcefield-toolkit.readthedocs.io/en/0.10.3/api/generated/openff.toolkit.topology.Topology.html#openff.toolkit.topology.Topology) class, which is meant to hold multiple chemical components whose chemical identities, positions, and potentially periodic cell are uniquely specified. Essentially all of the information you seek to store is already represented in a much more robust manner in `Topology`, and if the path to upgrade to `Topology` is desired, it's worth building in consistency between a current minimal interface and the `Topology` API. The one thing that is missing is labels/tags/identifiers for each component, but it's worth noting the `Molecule.name` field is already available to name the components.

I'd check with @j-wags about what expected changes are coming now that biopolymer support is being added. I don't imagine the `Molecule` object will experience significant API-breaking changes, however.

gufe/chemicalstate.py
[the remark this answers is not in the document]
dotsdl: At the risk of putting too much machinery into the edges, I think we can allow edge protocols to decide how to handle box vectors? As @richardjgowers suggested, many protocols wouldn't be tolerant of differing box vectors for the systems it connects, so these could fail quickly if the box vectors provided are not the same (or sufficiently the same).

Does this answer your question @IAlibay?

gufe/chemicalstate.py
dotsdl: Ah, we'll definitely need to solve this. @IAlibay can you raise an issue on this repo and link to the `openfe` issue that gives more detail on the problem?

gufe/chemicalstate.py
dotsdl: Agreed, I think having it as an optional node property makes sense, even if for vacuum boxes and boxes for RBFEs in water boxes they make more sense as an edge property. I think membrane proteins or any system where size/shape of the box is a key property is where these will have to be on the nodes.

dotsdl:
@richardjgowers I think I've addressed all the feedback above. Please merge if you're satisfied! Tests dependent on #5 unfortunately.

gufe/chemicalstate.py
dotsdl: @jchodera I believe hashability, and therefore immutability, where the major concerns with using `openff.toolkit.topology.Topology` directly as the `ChemicalState`, since our aim is to use the `ChemicalState` *itself* as a node in a `networkx`-based `AlchemicalNetwork`.

I think we recognize that both `Topology` and `ChemicalState` perform similar roles, but they serve different contexts: `Topology` is designed for the application of forcefield parameters on a system, while `ChemicalState` is designed for the hashability needs of the graph representation in an `AlchemicalNetwork`. For many protocols we anticipate generating `Topology`s from these hashable representations, but this sits downstream in our use of `ChemicalState`.

gufe/chemicalstate.py
richardjgowers: See #7

PR #5, opened 2022-03-28
Added Components parent class

richardjgowers:
with 4 base methods.

### DIFF
--- a/gufe/component.py
+++ b/gufe/component.py
@@ -20,5 +20,5 @@ class Component(abc.ABC):
 
     @classmethod
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


gufe/component.py
```diff
@@ -0,0 +1,24 @@
...
+    def __eq__(self, other):
+        pass
+
+    @abc.abstractmethod
+    def to_dict(self) -> dict:
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def from_dict(cls):
```
dwhswenson, line 23: this needs a `dct: dict` parameter

PR #10, opened 2022-03-30
Added formal charge to components, chemicalstate

dotsdl:
Adds `formal_charge` property to `ChemicalState`, all defined `Component` classes so far.

Renamed `ChemicalState` to `ChemicalSystem` (fixes: #11)


richardjgowers (commented):
This only hitch with this method is I'm not sure what the solvent component charge is. It's technically anything, and won't know its charge until it sees the State, (and I don't want a gross reference to the State in SolventComponent).

IAlibay:
> This only hitch with this method is I'm not sure what the solvent component charge is. It's technically anything, and won't know its charge until it sees the State, (and I don't want a gross reference to the State in SolventComponent).

Return the solvent charge if the solvent is properly defined (i.e. you passed in a solvated system), otherwise return NaN / None / other undefined value? Technically it's undefined at this stage and that's ok. Internally unless pre-defined we assume neutrality until ion addition happens.

gufe/chemicalstate.py
```diff
@@ -101,9 +101,9 @@ def identifier(self):
         return self._identifier
 
     @property
-    def charge(self):
-        """Total charge for the ChemicalState."""
-        return sum([component.charge for component in self._components.values()])
+    def formal_charge(self):
+        """Formal charge for the ChemicalState."""
+        return sum([component.formal_charge for component in self._components.values()])
```
richardjgowers, line 106: I think this needs to check if there's a neutralizing solvent, if so return 0 else return the sum ignoring nans.  So...

```python
if any(hasattr('neutralize', comp) and comp.neutralize
         for comp in self._components.values()):
    return 0
```

Which is a tad ugly tbh, but not sure how else to work around solvents.

richardjgowers (changes requested):
This one got missed: https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/solventcomponent.py#L59

I think this one needs to return `np.nan` since it's not really defined what the charge is (and nans break maths nicely if naively used), it's more like instructions of whether to neutralize or not...

richardjgowers:
@dotsdl I merged main into this and added the proposed `SolventComponent` formal charge implementation

gufe/chemicalstate.py
IAlibay: What happens here if a box of water & ions are passed through? (say a membrane protein where you need to define the water component because otherwise most solvation routines are too dumb and might add water inside the membrane?)

gufe/chemicalstate.py
```diff
@@ -101,9 +101,9 @@ def identifier(self):
         return self._identifier
 
     @property
-    def charge(self):
-        """Total charge for the ChemicalState."""
-        return sum([component.charge for component in self._components.values()])
+    def formal_charge(self):
```
IAlibay, line 104: Total charge?

gufe/component.py
```diff
@@ -22,3 +22,9 @@ def to_dict(self) -> dict:
     @abc.abstractmethod
     def from_dict(cls, d: dict):
         pass
+
+    @property
+    @abc.abstractmethod
+    def formal_charge(self):
```
IAlibay, line 28: Total charge?

gufe/smallmoleculecomponent.py
```diff
@@ -205,3 +205,7 @@ def _from_sdf_supplier(cls, supp):
             raise RuntimeError(f"SDF contains more than 1 molecule")
 
         return cls(rdkit=mol)  # name is obtained automatically
+
+    @property
+    def formal_charge(self):
```
IAlibay, line 210: As above, we only really care about the total charge of the component right?

gufe/solventcomponent.py
```diff
@@ -65,6 +66,10 @@ def concentration(self) -> unit.Quantity:
         """Concentration of ions in the solvent state"""
         return self._concentration
 
+    def formal_charge(self) -> int:
+        """Solvents don't have a formal charge defined so this returns nan"""
+        return math.nan
```
dwhswenson, line 71: mypy doesn't like this -- `math.nan` is a float, not an int

dwhswenson: `None` would make typing easier but math harder; OTOH, maybe we want to fail fast if user adds an unknown charge, rather than make them hunt down the source of a `nan` in their code?

gufe/chemicalstate.py
richardjgowers: Hmm or it could be that solvents look neutral and the edge gets to decide their actual charge? Like how box info is sometimes undefined.

IAlibay: It depends on how we do that tbh. I mentioned somewhere that we might want to do a "frozen solvent component" where we tell edges they absolutely cannot change anything about the solvent.

Also does it matter if the edge does something else than what the components on the node currently look like? Isn't the whole point of this method so that users & edges know what the node components look like and then use that to do further things? (parameterisation, charge corrections, etc...)

gufe/solventcomponent.py
IAlibay: I would advocate for `None` as a "there is nothing here" - `NaN` feels like there is an indication of something existing but it's just wrong?

dotsdl: I think I'm leaning more toward `None`, despite the backflips we get to do in handling in e.g. `sum`. Solves the `mypy` issue and agree with @dwhswenson and @IAlibay notes above.

gufe/chemicalstate.py
dotsdl: I think we can pursue a policy where edge protocols don't do anything in processing that would contradict the node definition. This means that if e.g. box vectors are fully specified in two nodes, an edge protocol *must* use those, and if it *can't* then it must fail upon being defined. 

In other words, node properties define a policy, and protocols should produce systems to be computed in line with the policies of the nodes they connect. If the policies are too strict for the protocol to work with, it should throw up its hands immediately on definition; if the policies are loose enough, it has that much ground to work with in downstream processing to produce computable MD systems.

Does this make sense?

richardjgowers: Yep that's a good definition.

gufe/smallmoleculecomponent.py
dwhswenson: +1 -- I think total charge is the better term here.

gufe/solventcomponent.py
richardjgowers: Ok went with None

richardjgowers (approved):

IAlibay (approved):
lgtm!

PR #12, opened 2022-03-30
started tests for chemicalstate

richardjgowers:


gufe/tests/test_chemicalstate.py
```diff
@@ -0,0 +1,51 @@
...
+
+import pytest
+from openff.toolkit.topology import Molecule
+
+import gufe
+
+
+@pytest.fixture
+def prot_comp(PDB_181L_path):
+    yield gufe.ProteinComponent.from_pdbfile(PDB_181L_path)
```
IAlibay, line 12: Can you replace this with a fully prepared protein with hydrogens please? Ping me if you want me to generate it, should only take a couple of minutes (I usually use the protein plus webserver + pymol for fixing any issues)

gufe/tests/test_chemicalstate.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.fixture
+def solv_comp():
+    yield gufe.SolventComponent(ions=('K', 'Cl'))
```
IAlibay, line 17: Some joke about chemical engineers preferring KCl over NaCl, but actually just to point out - NaCl might be a more common use case.

gufe/tests/test_chemicalstate.py
```diff
@@ -0,0 +1,51 @@
...
+def test_hash_and_eq(prot_comp, solv_comp, lig_comp):
+    c1 = gufe.ChemicalState({'protein': prot_comp,
+                             'solvent': solv_comp,
+                             'ligand': lig_comp})
+
+    c2 = gufe.ChemicalState({'solvent': solv_comp,
+                             'ligand': lig_comp,
+                             'protein': prot_comp})
+
+    assert c1 == c2
```
IAlibay, line 50: I'm not sure why codecov is complaining that it's not covering the equality method 😕

gufe/tests/test_chemicalstate.py
```diff
@@ -0,0 +1,51 @@
...
+def solv_comp():
+    yield gufe.SolventComponent(ions=('K', 'Cl'))
+
+
+@pytest.fixture
+def lig_comp(benzene_modifications):
+    yield gufe.SmallMoleculeComponent.from_rdkit(benzene_modifications['toluene'])
+
+
+def test_construction(prot_comp, solv_comp, lig_comp):
```
IAlibay, line 25: Could you make this "test_complex_construction" and then add a test for at least the ligand in water case  (just so we have the very minimum of what we need to get everything working)

IAlibay (changes requested):
mostly looking good, just some tiny comments / suggested changes

gufe/tests/test_chemicalstate.py
dwhswenson: Tests are incomplete -- there aren't tests for all the ways this could be not equal.

Is covering that intended to be a part of this PR? I was holding off on review until there were more tests.

IAlibay: Ah ok yeah makes sense, didn't see that line 77 was counting as covered.

gufe/tests/test_chemicalstate.py
richardjgowers: Yeah please, ideally it'll be pretty close to what we actually use

gufe/tests/test_chemicalstate.py
richardjgowers: done

gufe/tests/test_chemicalstate.py
IAlibay: Cool I'll ping it across in a few, just need to check the coordinates match with the benzene modifications.

IAlibay: @richardjgowers - I've added a new PDB and CIF files (181L-prepped.*) that contains 181L with protons, waters within 10 A of the positions where the benzene ligand would be, ACE and NME caps, and two Cl- ions which were part of the crystallographic artifacts.

Couple of things:

1. The ions aren't structural but were kept so that we had a direct test case for the presence of ions in the "protein" component.

2. The cap isn't necessary on one of the termini, but also good to have so we know it won't be rejected in real life.

Long term we should also add in a multi-chain test case too.

richardjgowers: @IAlibay can you just overwrite the old pdb&cif?

IAlibay: sure

IAlibay: Alright took some CI fixing, but done.

IAlibay (approved):
I did some changes to prevent the removal of hydrogens, since that's a lot more true to life. Overall lgtm but I'd be especially interested to hear what @dwhswenson thinks of the `__eq__` stuff.

gufe/chemicalstate.py
```diff
@@ -60,11 +60,27 @@ def __init__(
         else:
             self._box_vectors = box_vectors
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        if self.identifier != other.identifier:
```
dwhswenson, line 66: I think the only reason to have all these early exits is if we think equality checking is likely to be a significant cost. In that case, I'd go with the direct `._identifier`, rather than invoking he overhead of the property. (And similar with other things: `_box_vectors`, `_components`.)

gufe/chemicalstate.py
```diff
@@ -60,11 +60,27 @@ def __init__(
...
+            return False
+        if not np.array_equal(self.box_vectors, other.box_vectors,
+                              equal_nan=True):  # nan usually compares to false
+            return False
+        if self.components.keys() != other.components.keys():
+            return False
+        for k in self.components:
+            if self.components[k] != other.components[k]:
+                return False
```
dwhswenson, line 75: Is this going to be faster than Python's built-in `dict` equality? I suspect that has been pretty well optimized. Is there some other reason to separately test keys and values?

gufe/chemicalstate.py
```diff
@@ -60,11 +60,27 @@ def __init__(
...
+        return True
+
     def __hash__(self):
         return hash(
             (
                 tuple(sorted(self._components.items())),
-                self._box_vectors,
+                self._box_vectors.tobytes(),
                 self._identifier,
             )
```
dwhswenson, line 85: If performance of hashing is a concern, we might want to cache the hash and the use that as the first equality test.

dwhswenson (commented):
Haha... @IAlibay is right about the main thing I have questions about.

I don't mind early returns in `__eq__` if it's a performance concern. But if we're concerned about performance, there are a few other things (noted in comments) that will probably make this faster. Also, I'm not sure whether a sequence of if-statements with early returns would be faster than using `and`'s built-in short-circuit logic.

Anyway, basically looks good to me; suggestions are mainly if we're trying to focus on performance-critical things now.

dwhswenson (approved):
Pushed my little changes to `__eq__` (after chatting with @richardjgowers in Slack earlier). As far as I'm concerned, this is ready to merge -- will leave it open in case there's an evening review from someone else, but @IAlibay or @richardjgowers, feel free to merge when you start your morning.

dotsdl (approved):
Looks great @richardjgowers! Went through all tests, only made small edits.

mikemhenry (approved):
LGTM! We might want to look into adding hypothesis https://hypothesis.readthedocs.io/en/latest/index.html sometime to our CI so that we can explore (or sample hehe) all of the different combination of things we support

dwhswenson:
Hey, before I go to bed I now see that we're unanimous! So I'll merge it before Irfan or Richard wake up! 😁

PR #13, opened 2022-03-31
Add initial implementations of Transformation, AlchemicalNetwork, Protocol; stubs for Executor components

dotsdl:
Adds the following components:
- [x] `Transformation`
- [x] `Protocol`
- [x] `AlchemicalNetwork`

Also adds stubs for `Executor` components; these were necessary to reason about `Protocol` structure.

## Checklist
- [x] add `AlchemicalNetwork.__init__` with state held in `networkx` graph
- [x] add initial set of unit tests


gufe/transformation.py
```diff
@@ -0,0 +1,51 @@
...
+    # these could be expensive operations
+    def dg(self, estimator=None):
+        """Free energy difference of transformation based on given estimator,
+        using only data available for this edge.
+
+        """
+        ...
+
+    def ddg(self, estimator=None):
+        ...
```
richardjgowers, line 36: We're thinking of splitting pre/post computation data structures.  Maybe `dg` and `ddg` are best as methods of the edges of results network?

gufe/network.py
```diff
@@ -0,0 +1,27 @@
+"""
```
IAlibay, line 1: We had a minimal comment we were adding to all our `openfe` files, maybe the same thing can be done here? 

https://github.com/OpenFreeEnergy/openfe/blob/721abe4c4d1668cc6ab01e02bbcc22787156e92e/openfe/setup/ligand_network_planning.py#L1-L2

gufe/network.py
```diff
@@ -0,0 +1,27 @@
...
+
+"""
+
+from typing import FrozenSet, Iterable
+
+import networkx as nx
+
+from .chemicalstate import ChemicalState
```
IAlibay, line 9: Will need to be updated to ChemicalSystem or whatever the new namechange is.

gufe/transformation.py
IAlibay: That or I would say that we should just conflate them into a `results` method. My reasoning here is that if `Transformation` is meant to be the parent class to several downstream classes, then it should be up to children classes to implement various estimators - not everything might fit to every method?

gufe/transformation.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def chemicalstate_start(self):
```
IAlibay, line 38: What is meant by these?

gufe/transformation.py
```diff
@@ -0,0 +1,51 @@
+from .protocols import Protocol
+
+
+class Transformation:
```
IAlibay, line 4: Is this meant to be the parent class for all edges?

If so, given all the user stories from F@H alchemy, would it make sense not to call it `Transformation`, are there any cases where we wouldn't be doing a transformation between two distinct alchemical states but still using this parent class?

gufe/transformation.py
richardjgowers: I think a reference to the “first” state/node were moving between

IAlibay: I see, maybe being silly but wouldn't this be set in `init`?

richardjgowers: Yeah this is just a getter, so could be a property for example

gufe/network.py
dotsdl: Got it!

gufe/transformation.py
dotsdl: Yeah, these are just getters for the `ChemicalState`s at either end of the `Transformation`. Simplified naming down to `start` and `end`.

gufe/transformation.py
dotsdl: Might try and draft this in now. Basically just a `Transformation` with only one state, not two.

dotsdl:
So, we have objects for nodes and edges, but now we need to decide how state is structured in `AlchemicalNetwork`.
We have some options here.
1. Don't store state as `networkx` graph at all; simply hold on to all nodes and edges fed to `AlchemicalNetwork` as objects already (e.g. `Transformation`, `ChemicalSystem`). Only yield a `networkx.MultiDiGraph` on-demand as a method call.
2. Store all state as `networkx` graph; on `AlchemicalNetwork` init, convert all `ChemicalSystem` nodes and `Transformation` edges into a `MultiDiGraph` representation. Can use `ChemicalSystem`s directly as nodes, but `Transformations` will have to be dangled off edges in the graph as attributes.
3. Both 1 and 2, with duplicated state. Tolerable so long as object is immutable, but may present pain in serialization/deserialization roundtripping.
    - effectively works as a cache, with data in two forms most convenient for different purposes.
    - may also be the worst of both worlds

dwhswenson:
> * Don't store state as `networkx` graph at all; simply hold on to all nodes and edges fed to `AlchemicalNetwork` as objects already (e.g. `Transformation`, `ChemicalSystem`). Only yield a `networkx.MultiDiGraph` on-demand as a method call.

This is basically what we did the [OpenFE ligand network](https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/network.py). Advantage is that we can force things to be immutable, and we return a frozen NetworkX graph. It also gives us access to custom serialization for our objects.

dwhswenson:
Actually, I guess we sort-of do 3; but only because we cache the nx graph after first creation.

dotsdl:
Working notes from today, with some questions I could use some help resolving:

1. `name` or `identifier` for user-settable identifier?
    - want this the same across all object, yes?
2. should `ChemicalSystem` and `Transformation` have tags, key-value pairs?
    - basically, easy queryables for e.g. subgraph generation?
3. for serialization, reasonable to decree that all objects must have a `.to_dict` method?
    - is there a "gold standard" serialization approach we should identify? GraphML better than JSON? Something else?
    - need a registry for `Component`, `Protocol` subclasses to allow roundtripping for non-pickle formats
4. Although there may be conceptual value in separating `Transformation` from `Protocol`, not really any clear practical value
    - consider collapsing these abstractions together
        - subclasses of `Transformation`, `NonTransformation` are our protocols
        - `openfe` already does this

dwhswenson:
> * is there a "gold standard" serialization approach we should identify? GraphML better than JSON? Something else?

On this one point (too late at night for a more complete response -- and this is before even reviewing the code in detail), but I would say the folowing:

GraphML is a good serialization format for graphs. Its benefits are wide support, with the ability to recognize the same, repeated, node as the same object. That isn't guaranteed in naïve JSON representations. One of the advantages of GraphML is that it allows arbitrary node/edge attributes. (Personally, not a fan of XML, but GraphML is well supported by other programs.)

GraphML attributes can be stored as arbitrary strings. My recommendation would be for these strings to be JSON, which gives a good ability to serialize node/edge information. It becomes a kind of weird mix of XML and JSON, but there are good tools to handle each in most programming languages, so I think client code has the ability to work freely and serialize objects in this format.

dotsdl:
> > * Don't store state as `networkx` graph at all; simply hold on to all nodes and edges fed to `AlchemicalNetwork` as objects already (e.g. `Transformation`, `ChemicalSystem`). Only yield a `networkx.MultiDiGraph` on-demand as a method call.
> 
> This is basically what we did the [OpenFE ligand network](https://github.com/OpenFreeEnergy/openfe/blob/main/openfe/setup/network.py). Advantage is that we can force things to be immutable, and we return a frozen NetworkX graph. It also gives us access to custom serialization for our objects (EDIT: by serializing our nodes and edges to strings and making that into a NetworkX graph, which is then serialized using NetworkX.)

Adopted your approach; agree it is quite good. :grin:

gufe/protocols/protocol.py
```diff
@@ -0,0 +1,89 @@
...
+        ...
+
+    @abc.abstractmethod
+    def is_complete(self) -> bool:
+        """Check if the results of this workload already exist"""
+        ...
+
+    @abc.abstractmethod
+    def run(self, 
+            start: ChemicalSystem,
```
jchodera, line 65: Perhaps `initial` and `final` would be more consistent than `start` and `end` for transformations?

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
+from typing import Optional, Dict, Union, List
+
+from pydantic import BaseModel, validator
+from openff.units import unit
+
+
+class SystemSettings(BaseModel):
```
jchodera, line 7: It will be critical to sync up with the Open Force Field Initiative folks to come up with a self-consistent division between:

1. Choices that impact the fully interacting potential energy function (which OpenFF is aiming to fully encode in the force field definition)
2. Choices that impact the free energy difference computed for a transformation (atom mapping, alchemical modification functional form, thermodynamic state at which the calculation is performed)
3. Choices that impact the efficiency of the calculation (integrators, barostats, collision rates, alchemical protocols, alchemical methods, alchemical intermediates, hydrogen mass)

Right now, you're mixing several of these together. This is highly problematic, and will lead directly to an inability to fulfill the primary mission of OpenFE (enabling free energy calculations with OpenFF force fields) because OpenFF is encoding all options that impact the fully interacting system potential function _within the force field definition_. Therefore, enabling options that impact the potential energy definition (such as `nonbonded_method`, `nonbonded_cutoff`, `constraints`, `rigid_water`) will mean you are failing to actually perform free energy calculations with the specified OpenFF force fields.

I'd be happy to help clarify these issues, but I think we need a design document that clearly separates these items.

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+    nonbonded_method = 'PME'
+    nonbonded_cutoff = 1.0 * unit.nanometer
+    constraints: Union[str, None] = 'HBonds'  # Usually use HBonds
+    rigid_water = True
+    remove_com = True  # Probably want False here
+    hydrogen_mass: Union[float, None] = None
+
+
+class TopologySettings(BaseModel):
+    """Settings for creating Topologies for each component
```
jchodera, line 38: Can this be more clearly defined? I'm not clear on what this class is supposed to represent from the brief doctstring.

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+    constraints : str
+        Which bonds and angles should be constrained. Default None.
+    rigid_water : bool
+        Whether to apply rigid constraints to water molecules. Default True.
+    hydrogen_mass : float
+        How much mass to repartition to hydrogen. Default None, no
+        repartitioning will occur.
+    """
+    class Config:
+        arbitrary_types_allowed = True
```
richardjgowers, line 27: Just a note, this is required by pydantic to allow strange types.  In the future I think we could write validators for `openff.units` objects, then register these with pydantic somehow and not have to have this setting

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+    ----
+    * We can probably just detect the solvent model from the force field
+      defn. In that case we wouldn't have to have ``solvent_model`` here.
+    """
+    # mapping of component name to forcefield path(s)
+    forcefield: Dict[str, Union[List[str], str]]
+    solvent_model = 'tip3p'
+
+
+class AlchemicalSettings(BaseModel):
```
jchodera, line 58: It looks like this is a dumping ground for random `perses` alchemical configuration parameters. Perhaps we could meet specifically to discuss how to come up with a better object model to more flexibly represent alchemical protocols, alchemical intermediates, and the like? We've been meaning to do this for a while, and I think it can be done flexibly, extensibly, and elegantly. I fear that just dumping everything into an object like this with attributes that will need to be supported is going to be problematic in the medium- and long terms.

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+    softcore_electrostatics = True
+    softcore_alpha = 0.85
+    softcore_electrostatics_alpha = 0.3
+    softcore_sigma_Q = 1.0
+    interpolate_old_and_new_14s = False
+    flatten_torsions = False
+
+
+class EngineSettings(BaseModel):
+    """OpenMM MD engine settings
```
jchodera, line 114: This is general MM engine settings, not OpenMM specific?

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+               'online_analysis_minimum_iterations', 'gamma0')
+    def must_be_positive(cls, v):
+        if v < 0:
+            errmsg = ("Online analysis target error, minimum iteration "
+                      "and SAMS gamm0 must be 0 or positive values")
+            raise ValueError(errmsg)
+        return v
+
+
+class BarostatSettings(BaseModel):
```
jchodera, line 199: I think this is a problematic design, since it convolves parameters that define the alchemical state and hence impact the free energy (`pressure`) with parameters that simply control the efficiency of sampling (`frequency`).

I highly recommend instead following something like what `openmmtools` does and defining a [`ThermodynamicState`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/states.py#L385) that encodes the information about thermodynamic parameters that will impact the thermodynamic ensemble sampled (and hence the free energy)---such as `temperature` and optionally `pressure`, `pH`, etc. 

This should be separated from settings that simply impact the efficiency of the sampling.

gufe/protocols/protocol_settings.py
```diff
@@ -0,0 +1,381 @@
...
+      Number of attempts to restart from Context if there are NaNs in the
+      energies after integration. Default 20.
+    constraint_tolerance : float
+      Tolerance for the constraint solver. Default 1e-6.
+    """
+    class Config:
+        arbitrary_types_allowed = True
+
+    timestep = 2 * unit.femtosecond
+    temperature = 298.15 * unit.kelvin
```
jchodera, line 267: As with `BarostatSettings`, this convolutes thermodynamic parameters that impact the free energy of interest (`temperature`) with a mess of settings that impact the efficiency of sampling and stability of sampling. These should be separated, grouping temperature, pressure, etc. into the thermodynamic state information.

gufe/protocols/protocol.py
```diff
@@ -0,0 +1,89 @@
...
+
+    """
+    ...
+
+    def __init__(
+            self,
+            *,
+            settings: ProtocolSettings = None
+        ):
+        """
```
richardjgowers, line 36: If the two `ChemicalSystems` aren't here, then this class becomes a factory for simulations.  I think being able to serialize the factory separately is a good idea for extending a network with the same protocol (else where is the factory stored).

In which case, I think `run()` isn't run, but instead `prepare(Transformation(initial, final, mapping)) -> PreparedProtocol` and `PreparedProtocol` has the `run()` and `is_complete()` etc.

gufe/protocols/protocol_settings.py
jchodera: I'd also highly recommend considering we adopt some flexible definition of how the sampling is done, perhaps in a hierarchical manner: It's most important to specify first (1) _what_ calculation do you want to perform (openff-2.0.0 at 300 K and 1 atm), then (2) how you computed it, at a high level (e.g. replica exchange, nonequilibrium switching), then (3) how exactly it is performed can be left to best practices defaults, auto-tuned, or specified explicitly if this specification can actually be implemented by the engine. But I imagine a great deal of work will be trying to automate best practices and auto-tuning for each calculation type.

More broadly, in openmmtools, for the lowest level (3), we have embraced a [general MCMC framework](https://openmmtools.readthedocs.io/en/latest/mcmc.html) that allows us to compose or chain together MCMC samplers that sample from the same alchemical state in order to efficiently and flexibly sample from within an alchemical state. This is a much more general design than specifying all the details of a single integrator, since a calculation may want to compose compatible types of sampling strategies.

gufe/protocols/protocol.py
```diff
@@ -0,0 +1,89 @@
...
+        ...
+
+    @abc.abstractmethod
+    def is_complete(self) -> bool:
+        """Check if the results of this workload already exist"""
+        ...
+
+    @abc.abstractmethod
+    def run(self,
```
jchodera, line 64: If this is intended to be the actual class that can be `run()` to accumulate data from which a free energy difference can be estimated, it would be useful to structure this class so that it can meet the needs we've defined:
* It must be serializable so we can store it to disk, transmit it over the wire, or restore it later
* It must enable us to assess how complete the calculation is, rather than just check if it is fully complete
* It must enable us to assess the current free energy estimate and the estimates of statistical fluctuation if we want to be able to adaptively run calculations
* It must enable us to extend the calculation 

I honestly don't think this structure makes sense as-is, though. You probably need to generate some kind of simulation object that is constructed from asking to perform a particular kind of transformation that would then provide an API for all of the above queries.

gufe/protocols/protocol_settings.py
richardjgowers: Yep, we had this discussion internally and had a good argument over whether e.g. the cutoff is free from the force field etc, there's room for improvement here.  

While we need to support openff force fields, I'm not sure we can support them at the exclusion of other force fields, so there might have to be variations on these settings where the parameters are stated separately (such as is happening here).

gufe/protocols/protocol_settings.py
IAlibay: Starting with a point of curiosity - I realise that at the limit of infinite sampling the MC frequency has no impact on how the thermodynamic state of the system, but given we are likely running short simulations and often non-equilibrium ones, does the change in relaxation time not affect the sampled free energy?

A thought on the splitting of parameters - I can understand & support the idea of separating out the "things that affect the ensemble", but at what limit can we differentiate between algorithmic details that can have an impact and broad ensemble parameters? (e.g. the choice of `gamma_ln` for Langevin integrators or the splitting sequence, etc...)

jchodera: There will be many things that impact the the finite-time behavior of simulations---up to and including the exact choice of random numbers you generate. But it simply isn't sensible to convolve (1) parameters that impact the infinite-time limit of the calculation (thermodynamic state, potential function) with (2) parameters that impact the rate at which bias and variance (and hence total error to the asumptotic limit) diminish. There is a very good reason for this: 

First, it's possible to independently optimize and tune (or even auto-tune adaptively on the fly) the simulation strategy to make the rate of convergence to the asymptotic limit, and hence make the calculation more efficient. That makes for a very useful separation of effort, since work can be carried out independently on optimizing the efficiency of protocols, alchemical methods, engines, etc. 

Second, if you attempt to optimize solely for accuracy of finite-time calculations _to experiment directly_, you will never get anywhere. Any choice you make (force field, calculation strategy, collision rate) will impact these finite-time calculations in uncontrollable and unknowable ways. We will never make any progress in improving either the force field, the protocol, or the calculation strategy because there is absolutely no way to separate any of these effects. Progress becomes completely impossible because all modifications are a random walk through calculation space.

gufe/protocols/protocol_settings.py
jchodera: There is nothing in my argument that says you should support OpenFF to the exclusion of other force fields.
My only point is that you _must_ support OpenFF force fields, which means that you will need to bundle all of the parameters that impact the potential energy for a fully interacting system together, separated from other parameters that impact simulation efficiency.

One way to do this would be to recognize that an OpenFF force field will bundle all of the parameters that impact the potential energy of a fully-interacting system into a single object. For other force fields, you can use a container object that similarly bundles everything together, explicitly specifying all the parameters left undefined by the choice of force field. These objects can then be used interchangeably to specify which "complete" force field should be used, and the choices of auxiliary parameters for traditional force fields like GAFF can subsequently be optimized to define best practices.

jchodera: > Yep, we had this discussion internally and had a good argument over whether e.g. the cutoff is free from the force field etc, there's room for improvement here.

Might I also suggest that internal discussions about how to represent force fields should likely also include a representative from OpenFF? There seems to be a great deal of accumulated knowledge within OpenFF that is not making its way into these discussions at the moment.

richardjgowers:
@dotsdl I think adding the settings definition in here is maybe a step too far and/or removed from the API for Protocol.  We probably need to hash out the taxonomy of settings (which @jchodera has started) with a wider audience in something like a google doc and they try and encode this in nested pydantic structures later.

gufe/protocols/protocol_settings.py
richardjgowers: > There is nothing in my argument that says you should support OpenFF to the exclusion of other force fields. My only point is that you _must_ support OpenFF force fields, which means that you will need to bundle all of the parameters that impact the potential energy for a fully interacting system together, separated from other parameters that impact simulation efficiency.
> 

Yep sorry misread this a bit at first.

jchodera: Another great cross-cutting discussion to have might be "which force fields should be supported?" and "how can we unify force field support across simulation engines?". There are a number of tools that have been developed that attempt to band-aid this situation for now, like [simtk.openmm.app.ForceField](http://docs.openmm.org/latest/userguide/application/02_running_sims.html#a-first-example), [`openmmforcefields.generators.SystemGenerator`](https://github.com/openmm/openmmforcefields#automating-force-field-management-with-systemgenerator), [OpenFF Interchange](https://openff-interchange.readthedocs.io/), and now [espaloma](http://github.com/choderalab/espaloma). If we know what you'd like to support, there's likely some things that can be done to make this easy that OpenFE need not solely support.

gufe/protocols/protocol.py
dotsdl: Sounds good; done!

dotsdl:
> @dotsdl I think adding the settings definition in here is maybe a step too far and/or removed from the API for Protocol. We probably need to hash out the taxonomy of settings (which @jchodera has started) with a wider audience in something like a google doc and they try and encode this in nested pydantic structures later.

Understood, and agreed. I will remove `protocol_settings` from this PR.

gufe/protocols/protocol.py
dotsdl: Indeed, here are my notes when I separated things out this way:

2. Benefits of separating `Transformation` from `Protocol`
    - can have a single `Protocol` object instance serving many `Transformation`s
        - can save on memory for very large networks, serialization cost
        - may be worthwhile making them function as singletons when the hash is identical
    - only two types of `Transformation`: `Transformation` and `NonTransformation`
        - we allow for a zoo of `Protocol`s
    - can execute a `Protocol` in isolation, without any graph/network machinery at all
        - define `ChemicalSystem`s, `AtomMapping` if appropriate, settings, and fire it off with 
          ```
          Protocol(settings).run(start, end, mapping)
          ```
          
         - very similar to `sklearn`-style estimator classes
         - should we adopt a similar approach to `sklearn` for results?

dotsdl: Will have to think some more on `Protocol` structure to accommodate @jchodera's [points here](https://github.com/OpenFreeEnergy/gufe/pull/13#discussion_r853994681). I think you're right that a `Protocol` with an `initial`, `final`, and `mapping` at this point is a job that's ready to be executed, and it is what would need to be serializable to be shipped to a worker process, local or remote.

`PreparedProtocol.run` could have an `async=True` kwarg that proceeds with execution in a separate thread or process, with its methods able to yield result status from file outputs or stream buffers to the main thread/process. We'll need to determine what operations we support (stop, start, extend?).

gufe/network.py
```diff
@@ -0,0 +1,182 @@
...
+    @property
+    def client(self) -> Client:
+        return self._client
+
+    @client.setter
+    def client(self, client: Client):
+        # TODO add validation checks
+        self._client = client
+
+    def estimate(self, transformations: Iterable[Transformation], estimator: str = None) -> Tuple[Tuple[float],Tuple[float],Tuple[float]]:
```
richardjgowers, line 99: The estimate methods in general seem like something that could be pushed down onto `Transformation`s, what's wrong with this pattern

```python
for edge in network.filter(QUERY):
    estimate = edge.estimate(args)
```

gufe/protocols/protocol.py
```diff
@@ -0,0 +1,101 @@
...
+
+        """
+        ...
+
+    @abc.abstractmethod
+    def create(self, 
+            initial: ChemicalSystem, 
+            final: ChemicalSystem,
+            mapping: Optional[Mapping] = None,
+            **kwargs
```
richardjgowers, line 67: I think this would likely benefit from a `Settings` kwarg that could update the base parameters, unless that happens elsewhere.  A use case I'm thinking of is extending a simulation but with updated sampling settings that you've tuned with the previous simulation

gufe/transformations/transformation.py
```diff
@@ -0,0 +1,332 @@
...
+
+        Parameters
+        ----------
+        settings
+            Override level 3 settings with a ProtocolSettings.
+
+        """
+        # TODO: need an update path for level-3 settings overrides
+        if settings is not None:
+            self.protocol.__class__(settings=settings)
```
richardjgowers, line 156: Related to the comment above wrt settings, but I think this is mutating the transformation which might break the Network (depending on assumptions on how edges hash/behave)

gufe/protocols/protocolunit.py
```diff
@@ -0,0 +1,85 @@
...
+        block : bool
+            If `True`, block until execution completes; otherwise run in its own thread.
+
+        """
+        if block:
+            dep_results = [dep.estimate() for dep in self._dependencies]
+
+            self._status = "RUNNING"
+            self._execute(dep_results)
+            self._status = "COMPLETE"
```
richardjgowers, line 37: At the risk of one class too many, I think a status class could be handy, could just be a named tuple with fields like status enum(like), datetime started, datetime finished, host id, which then gets stored alongside the produced artifacts(?) as provenance.

gufe/executors/executor.py
```diff
@@ -0,0 +1,55 @@
+import datetime
+from typing import Iterable, Union
+
+from .results import ResultStoreClient
+
+
+class Executor:
+    """Destination for `AlchemicalNetwork` submission and result requests via `Client`.
+    
+    The `Executor` exposes a RESTful API for incoming `Client` requests.
```
richardjgowers, line 10: Are we committed to restful?  Pure python would also be nice at times

gufe/executors/executor.py
```diff
@@ -0,0 +1,55 @@
...
+
+    """
+    def __init__(self,
+            results: ResultStoreClient,
+            state: str,
+            ):
+        ...
+
+
+    def export_state(datetime: datetime.datetime):
```
richardjgowers, line 37: Is this meant to be `-> datetime`? I'm not sure how datetime is an argument

gufe/protocols/protocolunit.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def status(self):
+        if self._status == "WAITING":
+            # check dependencies; if all COMPLETE, change to READY
+            if set(dep.status for dep in self._dependencies) == {"COMPLETE"}:
+                self._status = 'READY'
+
+        return self._status
+
+    def get_artifacts(self) -> Dict[str, Path]:
```
richardjgowers, line 91: This might be better as `Dict[str, Pathlike]` where Pathlike just yields bytes on `read()` but might not actually be a file

gufe/protocols/protocolunit.py
```diff
@@ -0,0 +1,94 @@
...
+
+        """
+        return self._result
+
+    @property
+    def status(self):
+        if self._status == "WAITING":
+            # check dependencies; if all COMPLETE, change to READY
+            if set(dep.status for dep in self._dependencies) == {"COMPLETE"}:
+                self._status = 'READY'
```
richardjgowers, line 87: Is this pseudocode or literal?

gufe/protocols/results.py
```diff
@@ -0,0 +1,46 @@
...
+    name: str  # name of the `ProtocolUnit` that produced this `ProtocolUnitResult`
+    data: Any  # should likely be fleshed out, currently a free-for-all
+
+    def __init__(self, 
+            dependencies: Optional[List["ProtocolUnitResult"]] = None,
+            **data):
+
+        super().__init__(**data)
+
+        self._uuid = str(uuid.uuid4())
```
richardjgowers, line 30: So in a world where we know the protocol id, run id, gen id etc, do we really need uuids?  Because yes they're nice for generating a unique thing, but they're (by design) awful to interpret, and we have enough knowledge to generate a useful identifier.

gufe/protocols/protocoldag.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def name(self):
+        return self._name
+
+    def to_dask(self):
+        """Produce a `dask`-executable DAG from this `ProtocolDAG` as a `dask.Delayed` object.
+
+        """
+        ...
```
richardjgowers, line 46: Can we keep scheduler specific out of this?  I.e. can we put this logic in `DaskScheduler.submit(thing)`

richardjgowers (commented):
Might be nice to cherry-pick Network out of here, some of this is clashing with what @dwhswenson is doing with results storage

gufe/protocols/protocoldag.py
richardjgowers: Unless this is literally the object that dask consumes, in which case that's ok.  Not sure if we're implementing the scheduler interfaces or not.

gufe/executors/executor.py
dotsdl: Fixed!

gufe/protocols/protocolunit.py
dotsdl: Literal! This actually works, and we exercise it in `test_protocol.py`.

gufe/protocols/results.py
dotsdl: Yes, but since we want to be able to deduplicate `ProtocolUnit`s across `ProtocolDAG`s within a `Scheduler`, it makes less sense generally to speak in terms of the source transformation that it came from, since there may be many such transformations that generated that particular `ProtocolUnit`. Giving the `ProtocolUnitResult`a unique "stamp" allows for unambiguous reference to its `_depdendencies`, which also have their own stamp, even once serialized and shipped to a `ResultStore`.

I do agree that generally UUIDs are to be avoided where possible, but in this case I think they are the least bad option. Even a hash of `ProtocolUnit.data` may not be that reasonable, since it may include references to rather large files and so may not even be a complete, differentiating hash across instances.

gufe/protocols/protocoldag.py
dotsdl: This was more of a stub for an idea, where we can produce fully-executable `dask` graphs that can be run with `dask`'s own schedulers (local and `distributed`). I can remove it from this PR and consider it again later.

dotsdl:
Ready for review! Seeking to merge when satisfied.

gufe/protocols/protocolunit.py
richardjgowers: Actually I think I hate this idea after all.

gufe/protocols/protocolunit.py
```diff
@@ -0,0 +1,92 @@
...
+    def result(self) -> ProtocolUnitResult:
+        """Return `ProtocolUnitResult` for this `ProtocolUnit`.
+
+        Requires `status` == "COMPLETE"; exception raised otherwise.
+
+        """
+        return self._result
+
+    @property
+    def status(self):
```
richardjgowers, line 83: I started sketching out what the state diagram for this would look like if this class didn't store the state in an attribute:
- if there's completed output then its FINISHED, either SUCCESS or ERROR
- if there's something in the scheduler for this, it's RUNNING
- if it has unresolved dependencies, it's WAITING
- else it's READY

```mermaid
  graph TD;
      WAITING-->READY;
      READY-->RUNNING;
      RUNNING-->FAILED;
      RUNNING-->COMPLETED;
```

This would require the Unit to see a) the results store to determine if output exists, b) the scheduler to determine if a job of this type exists...

Then I realised that that's not really the job of this object, it's just meant to encode how to run the thing, and all this knowledge of status should be external to this class, so I'd remove `status` on Unit.  Similarly I'd remove the `result()` and `get_artifacts()` methods, these also require knowledge of state (knowing if/where the Unit was ran).  Instead just make a Worker responsible for capturing the output of `execute()` and store the tuple of `(Unit, Result)` somewhere.  So all in all, this is treading on the toes of the scheduler too much and should get stripped down.

dotsdl:
Ready to go @richardjgowers! Got tests passing again after DAG restructuring from yesterday.

dotsdl:
@IAlibay blackified changes in the PR against `main` using `darker --revision main gufe`.

richardjgowers (approved):

gufe/transformations/transformation.py
```diff
@@ -0,0 +1,263 @@
...
+        -------
+        ProtocolResult
+            Aggregated results from many `ProtocolDAGResult`s from a given `Protocol`.
+
+        """
+        return self.protocol.gather(protocol_dag_results=protocol_dag_results)
+
+
+# we subclass `Transformation` here for typing simplicity
+class NonTransformation(Transformation):
```
IAlibay, line 169: So throwing my 2 cent here, I don't really like the `Transformation` and `NonTransformation` naming (especially the latter), it's strangely non-intuitive (especially if you have nontransformation sitting in transformations, I honestly can just see users sitting there being confused).

Is there an alternative we could use here? The concept of AlchemicalPaths (through states) or just calling them Experiments? Even being more descriptive with AlchemicalTransform(ation) would probably ease some of the likely pain points users are going to feel here.

gufe/tests/test_transformation.py
```diff
@@ -0,0 +1,114 @@
...
+            mapping=None,
+        )
+        assert absolute_transformation == identical
+
+    def test_dict_roundtrip(self):
+        # TODO: need registration of `Protocol`s for this to work
+        ...
+
+
+class TestNonTransformation:
```
IAlibay, line 74: From a cursory look these two classes seem like they could be squashed to the one base class with just ntnf renamed to something that's declared as a class attribute?

gufe/protocols/protocoldag.py
```diff
@@ -0,0 +1,82 @@
...
+    def execute(self) -> ProtocolDAGResult:
+        """Execute the full DAG in-serial, in process."""
+
+        # operate on a copy, since we'll add ProtocolUnitResults as node attributes
+        graph = self._graph.copy(as_view=False)
+
+        completed: Set[ProtocolUnit] = set()
+
+        while len(completed) != len(self._graph.nodes):
+            for pu in graph.nodes:
```
IAlibay, line 57: Pedantic but for the sake of sanity can we make `pu` a bit more explicitly named here? Someone is going to eventually mess up when looking at `pu` and `pur`

gufe/protocols/protocolunit.py
```diff
@@ -0,0 +1,78 @@
...
+        settings: Optional["ProtocolSettings"] = None,  # type: ignore
+        name: Optional[str] = None,
+        **kwargs
+    ):
+
+        self._settings = settings
+        self._kwargs = kwargs
+        self._name = name
+
+    def __hash__(self):
```
IAlibay, line 30: Should there be an `__eq__` to deduplicate units? (or maybe deduplication isn't happening at this level?)

gufe/tests/test_network.py
```diff
@@ -0,0 +1,98 @@
...
+    styrene,
+    prot_comp,
+    solv_comp,
+):
+
+    variants = [toluene, phenol, benzonitrile, anisole, benzaldehyde, styrene]
+
+    # define the solvent chemical systems and transformations between benzene and the others
+    solvated_ligands = {}
+    solvated_ligand_transformtions = {}
```
IAlibay, line 29: here and everywhere else `transformtions` -> `transformations`? (i.e. do we really save anything from a single letter?)

IAlibay (changes requested):

richardjgowers:
@dotsdl I pushed some stuff up

gufe/tests/test_network.py
richardjgowers: I changed this

gufe/protocols/protocoldag.py
richardjgowers: reworked this loop to be much simpler

gufe/transformations/transformation.py
richardjgowers: Experiment sounds like a bench thing.

What a `Transformation` is the application of a `Protocol` onto specific chemical systems.  It is also used as the edge on alchemical Graphs.

You could call it `ParametrisedProtocol` but I feel like we've already got too many names revolving around Protocol.

I don't hate `AlchemicalTransformation`

gufe/protocols/protocoldag.py
```diff
@@ -0,0 +1,64 @@
...
+    def graph(self):
+        return self._graph
+
+    def execute(self) -> ProtocolDAGResult:
+        """Execute the full DAG in-serial, in process."""
+        # operate on a copy, since we'll add ProtocolUnitResults as node attributes
+        graph = self._graph.copy(as_view=False)
+
+        # iterate in DAG order
+        for unit in reversed(list(nx.topological_sort(self._graph))):
```
richardjgowers, line 54: So we're having to reverse the provided iterator for DAGs from Networkx.  Not a drama, but it might be worth revisiting if we've got our DAG definition the right way round.

gufe/transformations/transformation.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class Transformation(Serializable):
+    """An edge of an alchemical network.
+
+    Connects two `ChemicalSystem`s, with directionality.
+
+    Attributes
+    ----------
+    initial : ChemicalSystem
```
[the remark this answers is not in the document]
dotsdl: I'm not married to `initial` and `final`, but I believe I'm going from @jchodera's suggestion from earlier on this PR.

gufe/protocols/protocoldag.py
dotsdl: Apologies @richardjgowers, from our discussion Friday we had decided to encode DAGs as `ProtocolUnit`s pointing to their dependents, but traversal appeared easier to go the other direction (`ProtocolUnit`s point to their dependencies). 

I should have mentioned this in a comment here that I reversed it in the tests, but I think you noticed this.

dotsdl:
> I've got an off by one error on the dependencies in one of the tests. I think I'll change the test to be more meaningful (sum of integers) so it'll be easier to debug.

The problem appears to be that `nx.descendants` [gives *all* reachable nodes of the given node](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.descendants.html#networkx.algorithms.dag.descendants), not just nearest neighbors. We only want immediate dependencies included in `dependency_results`, but for `FinishUnit` we end up getting all previous `ProtocolUnit`s in there.

gufe/transformations/transformation.py
dotsdl: Agree `NonTransformation` is a bit obtuse. I think it's also totally descriptive without stacking in any additional, unintended meaning. One could even imagine it as a lambda protocol that changes a system to itself, effectively just doing dynamics and no alchemistry.

I'm fine with `Transformation` being renamed to `AlchemicalTransformation`, which goes with our `AlchemicalNetwork`.

gufe/tests/test_transformation.py
dotsdl: Yes, there is some potential for consolidation, but there are some differences in `test_equality`.

dotsdl:
Fixed tests with change to use only immediate successors instead of all descendants.

gufe/protocols/protocoldag.py
richardjgowers: Either is fine, you're either saying "A leads to B" or "B requires A", I just don't want to be going against the grain with how everyone else seems to phrase it.  Can leave this as-is, but if we keep finding counter examples we might have to switch polarity.

gufe/transformations/transformation.py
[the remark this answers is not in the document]
richardjgowers: We'll keep A/B and have a great renaming once we're happy with the objects.  I think we're still determining how many objects/layers there are right now.

gufe/transformations/transformation.py
richardjgowers: I'll kick this down the road to the "great renaming" event we can have once we've got all the objects.

richardjgowers (approved):
Ok looks good enough to start trying it out now, I think it'll be easier to see how it fits once we try it on.  @IAlibay want to merge?

IAlibay (approved):
lgtm!

@richardjgowers please go ahead with the merge, I'm on the move right now.

dotsdl:
Thanks all! Very excited to build on this!

PR #15, opened 2022-04-01
Make `ChemicalSystem` an `abc.Mapping`

dwhswenson:
In starting to write code that might use this, I'm finding that `chemical_system.components['thing']` is coming up over and over again. I propose making `ChemicalSystem` an `abc.Mapping`, so that we can use it as `chemical_system['thing']`.


richardjgowers (approved):
No reason not to I guess

PR #16, opened 2022-04-03
added SmallMolecule to_dict/from_dict behaviour

richardjgowers:


gufe/smallmoleculecomponent.py
```diff
@@ -124,11 +125,15 @@ def __eq__(self, other):
         return hash(self) == hash(other)
 
     def to_dict(self) -> dict:
-        raise NotImplementedError()
+        """Serialize to dict representation"""
+        d = self.to_openff().to_dict()
+        return d
```
richardjgowers, line 130: One thing I'm not sure of is maybe we should tag this dict with a version tag, and then the from_dict keeps track... so

```python
# in to_dict
d = self.to_openff().to_dict()
d['gufe_version'] = 1  # i.e. our first schema, if/when we update our to_dict behaviour, increment this number

# in from_dict
if d['gufe_version'] == 1:
    # current stuff
elif d['gufe_version'] == 2:
   # future stuff, etc
```

Alternatively we enjoy our project youth and allow this to break horribly as its not in production yet.

IAlibay: So whilst we're not close to API stable, I don't think adding functionality that takes advantage of versioning is worth it. That being said, it might be good to have it there early just purely for our sanity.

gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -182,4 +183,32 @@ def test_sorting(ethane, alt_ethane):
...
+        d = phenol.to_dict()
+
+        assert isinstance(d, dict)
+
+    def test_deserialize_roundtrip(self, toluene, phenol):
+        # TODO: Currently roundtripping via openff adds Hydrogens even when
+        #       they weren't in the original input molecule.
+        roundtrip = SmallMoleculeComponent.from_dict(phenol.to_dict())
+
+        assert roundtrip == phenol
```
IAlibay, line 200: Our hash is purely smiles, could you check coordinates too please? (just in case)

IAlibay (changes requested):
Just the one thing, it would be good to make sure that coordinates are preserved on roundtrip (since they are important at this stage).

gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -200,6 +200,14 @@ def test_deserialize_roundtrip(self, toluene, phenol):
         assert roundtrip == phenol
         assert roundtrip != toluene
 
+        # check the coordinates, these aren't in the hash
+        # but are important to preserve
+        pos1 = phenol.to_openff().conformers
+        pos2 = roundtrip.to_openff().conformers
+        assert len(pos1) == len(pos2)
+        for x, y in zip(pos1, pos2):
+            assert (x == y).all()
```
IAlibay, line 209: wouldn't an assert_allclose work here?

IAlibay (approved):
lgtm - I'm happy to use raise an issue for https://github.com/OpenFreeEnergy/gufe/pull/16/files#r841249213 if we don't want to deal with it here.

richardjgowers:
@IAlibay yeah was just thinking forwards for how we'll deal with that in the future

PR #17, opened 2022-04-04
WIP: nasty hack, optionally load a PDBFile in from_pdb for now

richardjgowers:
Optionally, makes `from_pdbfile` populate a secret `_openmm_rep` attribute, which holds a `PDBFile`.  Not a huge fan, but I think this is the best solution for now.

### DIFF
--- a/gufe/proteincomponent.py
+++ b/gufe/proteincomponent.py
@@ -7,7 +7,6 @@ except ImportError:
     HAS_OPENMM = False
 else:
     HAS_OPENMM = True
-from typing import Optional
 
 from gufe import Component
 from gufe.custom_typing import RDKitMol, OEMol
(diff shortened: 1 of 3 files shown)


gufe/proteincomponent.py
```diff
@@ -1,6 +1,12 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
 from rdkit import Chem
+try:
```
IAlibay, line 4: This is more of a thought, but if there isn't a package for optional importing that works maybe we should consider making one at some point (would be useful for MDA too).

gufe/proteincomponent.py
```diff
@@ -39,6 +52,9 @@ def from_rdkit(cls, rdkit: RDKitMol, name=""):
     def to_rdkit(self) -> RDKitMol:
         return Chem.Mol(self._rdkit)
 
+    def to_openmm_PDBFile(self):
+        return self._openmm_rep
```
IAlibay, line 56:
```suggestion
        if HAS_OPENMM:
            return self._openmm_rep
        else:
            raise ImportError("OpenMM module is not available")
```

Possible suggestion here, I feel like returning `None` could be dangerous / misleading. You might as well break early if someone wants to do something they can't do?

IAlibay (changes requested):
Just needs a test

gufe/proteincomponent.py
```diff
@@ -61,11 +79,11 @@ def from_dict(cls, d: dict):
...
-        return hash((self.name, Chem.MolToSequence(self._rdkit)))
+        return hash(self.name)
 
     def __eq__(self, other):
         return hash(self) == hash(other)
 
     @property
     def total_charge(self):
-        return Chem.GetFormalCharge(self._rdkit)
+        return 0
```
IAlibay, line 89: There must be a way to get this, gimme 5.

IAlibay: Nope won't work, in that cases return `None`? The total charge is just undefine for now.

```suggestion
        return None
```

IAlibay (approved):
The total charge thing is more infuriating to look at than the previous rdkit version, but I can live with it 🤣 

Just the one suggestion for 0 -> `None` for charge, up to you though.

PR #18, opened 2022-04-04
Tightening up `SolventComponent`

richardjgowers:
- if given, must give exactly two ions, one position one negative
- force use of kwargs to initialize
- rename concentation to ion_concentration
- added usage example
- allows ions=None rather than empty tuple

### DIFF
--- a/gufe/solventcomponent.py
+++ b/gufe/solventcomponent.py
@@ -1,5 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
+from __future__ import annotations
+
 from openff.units import unit
 from typing import Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


dwhswenson (dismissed):
Can we change `ions` to `positive_ion, negative_ion`? Or we need to provide those as properties. Otherwise downstream needs to maintain the list of what elements give positive/negative ions to extract that. Changing the args also means we can be explicit in the defaults in the signature, which will help users.

Can we add something to the docstring of `ion_concentration` to clarify that this is the concentration *after* neutralizing ions have been added? (I believe that it correct for what is intended, but it's worth making that clear to users.)

Otherwise looks good; thanks for adding the clarification!

IAlibay (approved):
There's 5.5h until the board meeting and this is necessary to get the openmm protocol working (which I need in order to do the actual notebook...). Please merge.

PR #21, opened 2022-05-12
Raw data storage

dwhswenson:
Code written while I was thinking through some of the second round of storage ideas. See also [Google doc](https://docs.google.com/document/d/1wcTy4TqBIGbxdt9p2reVTY4J5iKAzF673DQsZeZ8vg8) with the new thoughts.


gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+                raise ChangedExternalResourceError(
+                    msg + " To allow this, set ExternalStorage."
+                    "allow_changed = True"
+                )
+            else:
+                warnings.warn(msg)
+
+    def get_sha2(self, location: str):
+        filelike = self._as_filelike(location, bytes_mode=True)
+        sha2 = hashlib.sha256()
```
richardjgowers, line 44: So sha256 is better, but md5 is faster.  We're not really worried about collisions or security, so can we pick speed here?

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+        # TODO: this can be probably be improved by chunking
+        sha2.update(filelike.read())
```
richardjgowers, line 46: Probably a future thing, but would it be worth also storing last modified time as a quicker hash, then actually compute sha if that fails?

richardjgowers, line 46: Probably a future thing, but would it be worth also storing last modified time as a quicker hash, then actually compute sha if that fails?

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return (str(path), self.get_sha2(path))
+
+    def _as_path(self, location):
+        return self.root_dir / pathlib.Path(location)
+
+    def _as_filename(self, location):
+        return str(self._as_path(location))
+
+    def _as_filelike(self, location, bytes_mode: bool = True):
```
richardjgowers, line 96: So encoding is a pain, but treating everything as ascii/bytes is probably a good way to avoid that headache (in this domain), and force people to know their encodings on getting their bytes back

richardjgowers, line 96: So encoding is a pain, but treating everything as ascii/bytes is probably a good way to avoid that headache (in this domain), and force people to know their encodings on getting their bytes back

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+        # we'd like to not need to include the as_filename method, but for
+        # now some consumers of this may not work with filelike
+        self.validate(location, sha2)
+        return self._as_filename(location)
+
+    def as_filelike(self, location, sha2, bytes_mode: bool = True):
+        self.validate(location, sha2)
+        return self._as_filelike(location, bytes_mode)
+
+    def create(self, location, byte_data):
```
richardjgowers, line 59: rename to store?

richardjgowers, line 59: rename to store?

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+        raise NotImplementedError()
+
```
richardjgowers, line 61: Add a `delete` method in the abstract?

richardjgowers, line 61: Add a `delete` method in the abstract?

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+        sha2.update(filelike.read())
+        return sha2.hexdigest()
+
+    def as_filename(self, location, sha2) -> str:
+        # we'd like to not need to include the as_filename method, but for
+        # now some consumers of this may not work with filelike
+        self.validate(location, sha2)
+        return self._as_filename(location)
+
+    def as_filelike(self, location, sha2, bytes_mode: bool = True):
```
richardjgowers, line 55: rename to `load_as_filelike`?

richardjgowers, line 55: rename to `load_as_filelike`?

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,113 @@
...
+            if not self.allow_changed:
+                # NOTE: having it here instead of in a DataLoader means that
+                # you can only change it for the whole system, instead of
+                # for a specific object
+                raise ChangedExternalResourceError(
+                    msg + " To allow this, set ExternalStorage."
+                    "allow_changed = True"
+                )
+            else:
+                warnings.warn(msg)
```
richardjgowers, line 40: Not sure what the purpose of this is?  If I know the location and sha of something and I want to check that it got stored correctly?

richardjgowers, line 40: Not sure what the purpose of this is?  If I know the location and sha of something and I want to check that it got stored correctly?

richardjgowers (commented):
Looking good, put up the tests so I can see usage examples

richardjgowers:
@dwhswenson wouldn't returning a context manager sort this out?  So the usage pattern would be:

```python
with data_store.as_filelike('foo.txt', ...) as f:
    data = f.read()
```

We used to have an `ExternalResource` class where the `__enter__` could live right?

In my experience, `__del__` is a bad idea to play with, usually a context manager is enough

dwhswenson:
@richardjgowers : That can already be done -- filelikes (at least, the ones in Python 3+) are context managers. The issue is that users can also use it (without error) as:

```python
file = file_storage.as_filelike("foo.txt")
file.read()
```

but then the responsibility is on the user to know to close. This is just like `open` normally is, but my point is that when I see a non-`with` `open` statement, I immediately see that I need a close (just like I know I need a `free` after I see a `malloc`). We need to tell users that they should see `as_filelike` the same way.

I had been hoping for some automatic way to handle this, but I don't think there is one. The only way we can scope the resource acquisition/release is by asking users to it as a context manager.

richardjgowers:
If we wrap the resource inside a context class (see below), then they **must** use a `with` statement else they get nothing out.  So I think this fixes the problem, by only allowing access within a `with` statement.

```python
class ForcedContext:
    def __init__(self, fn):
        self.fn = fn
        
    def __enter__(self):
        self._fhandle = open(self.fn)
        
        return self._fhandle
    
    def __exit__(self, exc_type, exc_value, traceback):
        self._fhandle.close()

with ForcedContext('foo.txt') as f:
    print(f.read())  # works

ForcedContext('foo.txt').read()  # doesn't work
```

dwhswenson:
That works; thanks!

gufe/storage/externalresource.py
dwhswenson: Yep (both on reasonable possibility, and on hold for future). Anticipating this, I changed the method name to `get_metadata`. Currently only returns the hexdigest string, but in the future that might be a (named?) tuple of metadata.

There may be some backend-specific weirdness (like if some backend doesn't preserve last-modified) but this should be backend-specific, and we could use `None` as a sentinel value for that.

gufe/storage/externalresource.py
dwhswenson: Done.

gufe/storage/externalresource.py
dwhswenson: Renamed to `load_stream` to match with others in the stack. Also renamed `as_filename` to `get_filename` (the `as_*` names were left over from an implementation where each "file" was wrapped by an object of ours, in which case you could get the filename representation or the actual byte stream.)

gufe/storage/externalresource.py
dwhswenson: This is the place where we actually validate the metadata, e.g., to ensure that when we reload the file, it hasn't changed without our knowledge.

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,161 @@
...
+    def store(self, location, byte_data):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def exists(self, location):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def delete(self, location):
+        raise NotImplementedError()
```
richardjgowers, line 82: I was tempted to put the docs here, but then annoyingly you'd have to rewrite the docs when you reimplemented this.  Would it be worth making `_delete` the `abc.abstractmethod` to allow the docs for `delete` to be consistent?  Or are individual implementations going to need their own docs?

dwhswenson: I agree that `_delete` is the better way. Will implement.

This should be seen as an adapter that allows us to abstract the details of storage and provide a limited API for functionality that all backends must support. The only reason not to have `delete` define the API and then call `_delete` would be if the call signature or behavior would depend on the specific subclass. But allowing differences in subclass behavior would break the whole point of having an API with interchangeable backends.

gufe/storage/externalresource.py
dwhswenson: On this particular point: Is it worth wrapping the metadata in that namedtuple already? Right now it would only have one field (`md5`), but it would be easier to extend it in the future (in ways that might be more easily backward compatible).

The argument against it is YAGNI. Keep in mind that other metadata (last modified, etc.) will probably only be used as a fast-failure approach. A successful check will still need to calculate the md5. And I'd expect that most of the time, the file passes the metadata check. So adding paths for fast failure doesn't seem like something we should spend significant time on.

richardjgowers (approved):

gufe/storage/externalresource.py
```diff
@@ -0,0 +1,301 @@
...
+    def _get_filename(self, location):
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def _load_stream(self, location):
+        raise NotImplementedError()
+
+
+# TODO: this should use pydantic to check init inputs
+class FileStorage(ExternalStorage):
```
richardjgowers, line 204: Might be better to split this out into another file, same with MemoryStorage below

gufe/storage/metadatastore.py
```diff
@@ -0,0 +1,59 @@
+import json
```
richardjgowers, line 1: need copyright headers

gufe/storage/resultsclient.py
```diff
@@ -0,0 +1,141 @@
...
+        # across sessions
+        return str(hash(item))
+
+    def __getitem__(self, item):
+        hash_item = self._to_path_component(item)
+
+        if hash_item not in self._cache:
+            self._cache[hash_item] = self._load_next_level(item)
+
+        return self._cache[hash_item]
```
richardjgowers, line 39: I remember from discussion we talked about the possibility of this checking if the file existed and only proceeding to `load_next_level` if `item` was a dir

gufe/storage/resultstore.py
```diff
@@ -0,0 +1,63 @@
+import warnings
+
+from typing import ClassVar
+
+from gufe.storage.errors import (
+    MissingExternalResourceError, ChangedExternalResourceError
+)
+
+
+class ResultStore:
```
richardjgowers, line 10: This class name is very similar to ResultsClient and might be better renamed to something else

dotsdl (approved):
Looks great @dwhswenson ! I made some changes that I don't think you'd object to, but please feel free to correct me if you dislike what I've done here. 

Feel free to merge if you're satisfied and we can build on the foundation you've built here!

gufe/storage/resultsclient.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class _ResultContainer:
+    """
+    Abstract class, represents all data under some level of the heirarchy.
+    """
+    def __init__(self, parent, path_component):
+        self.parent = parent
+        self._path_component = self._to_path_component(path_component)
+        self._cache = {}
```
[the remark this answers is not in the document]
dwhswenson: The "subdirectories" are cached. The use case I was thinking was: Say you get a bunch of paths as strings. You want to iterate over then and do something with each. Rather than instantiate another object for each "subdirectory" (and possibly have to worry about memory cleanup) we ensure uniqueness of each "directory" object by caching it and searching the cache first. This should *not* cache file handlers/contexts returned for the direction.

As for when it happens in the `__getitem__`.

I thought I had some use case where this was potentially a significant issue, although honestly I'm not sure now. It might be a YAGNI.

gufe/storage/resultstore.py
dwhswenson: Changed this to `ResultServer`

gufe/storage/resultsclient.py
dwhswenson: Added this. Gotta say, I feel like I have a pretty good design when adding functionality like this is as easy as it was.

dwhswenson:
A few final changes made:

* renamed `ResultStore` -> `ResultServer`
* made `_iter_contents` the abstract method, to enforce default `prefix=""` and to guard docstring from ABC
* made a `Metadata` class, instead of the `dict` that @dotsdl had made. This gives us better typing, and I'd rather be a bit more strict in the metadata fields (avoid typos causing problems)
* reverted [a7e089f](https://github.com/OpenFreeEnergy/gufe/pull/21/commits/a7e089fd81bd226fdc7ecfe7b9d8efe0f5f44238). We'd really like to remove `get_filename` from the API altogether, but the premise is that some consumers may *require* filenames instead of filelikes for analysis. Ideally, we fix it in those projects (as @dotsdl did for Alchemlyb), but dumping to a temporary file may be the only solution in some cases. We would not expect it to be any easier to pass arbitrary URIs to those tools.
* Added the ability to get a file handler context from the `__div__` and `__getitem__` conveniences in `ResultClient`. Note that this only works for things that have been explicitly stored there (i.e., things with metadata about them stored as well).

@dotsdl @richardjgowers: Feel free to merge if these changes look good to you.

dotsdl:
Excellent work @dwhswenson! Apologies, meant to give more explicit thumbs up. Excited to build on this foundation!

PR #23, opened 2022-05-24
(WIP) Stable object hashes

### LINKED ISSUE #22: Hash needs to work across python sessions
https://docs.python.org/3/reference/datamodel.html#object.__hash__

Probably need to go to md5

richardjgowers:
Fixes #22 

added str methods to components
- can just call `str(c)` to get something useful out
- for SmallMoleculeComponent this is the sdf, for proteins likely the mmcif?  Is this useful or should we be more explicit

added stable hash method to components
- currently this uses blake2 as it was fastest, but maybe using md5 allows command line generation too (blake2 probably isn't built into terminals)
- this is an int representation of the first 8 bytes of the (blake2) hash.  This should also allow a foreign package to calculate the same hash, which might be useful...

### DIFF
--- a/gufe/component.py
+++ b/gufe/component.py
@@ -2,17 +2,36 @@
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
 import abc
+import hashlib
 from typing import Union
 
 
 class Component(abc.ABC):
     """Base class for members of a ChemicalSystem"""
-    @abc.abstractmethod
+
+    # defining this here doesn't actually allow subclasses to inherit, see:
+    #  see https://bugs.python.org/issue1549
     def __hash__(self):
-        pass
+        # in python hash(str) isn't deterministic (by design)
+        # to provide a stable hash across sessions:
+        # generate the string representation of this component
+        # (the definition of this string left to individual components)
+        s = str(self)
+
+        # pass this through a (fast) stable hash algorithm
+        hasher = hashlib.blake2b()
+        hasher.update(s.encode('utf-8'))
+        dig = hasher.digest()
+
+        # return big endian int representation of the first 8 bytes
+        return int.from_bytes(dig[:8], byteorder='big')
 
     def __lt__(self, other):
         return hash(self) < hash(other)
+
+    @abc.abstractmethod
+    def __str__(self):
+        ...
 
     @abc.abstractmethod
     def __eq__(self, other):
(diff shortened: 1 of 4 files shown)


dwhswenson:
A couple of first thoughts; short of a full review:

> * for SmallMoleculeComponent this is the sdf, for proteins likely the mmcif?  Is this useful or should we be more explicit

I'm not sure about this decision -- I would lean toward name (or something closer to a repr) for the str function. I'm very much in support of some widely-used method name for a longer string representation, though.

In OPS, we made this mistake with our path ensemble objects. The `__str__` for them is a long string that uniquely identifies the ensemble (except for possible ordering differences). The problem is that any time you're print-debugging (which is what we expect from a lot of users/client code writers) you get inundated with this huge string. It's useful to have that string available, but over the years, I have found that having it available at `__str__` was more inconvenient than useful.

For example, it's pretty natural to do something like:

```python
for component in components:
    print(component, value_calculated_from(component))
```

It's a headache to remember to `repr(component)` or `component.name` there.


> * this is an int representation of the first 8 bytes of the (blake2) hash.  This should also allow a foreign package to calculate the same hash, which might be useful...

I think we should distinguish between hash for session-independent identification and hash for Python's dict/set purposes. Hash for identification should be the full hash (I'm fine with either blake2 or md5), and should be cached after first creation. Hash for dict/set purposes should probably still take advantage of Python's built-in hashing, including the random salting (which is there to defend against a [known security vulnerability](http://ocert.org/advisories/ocert-2011-003.html)).

Although I assume the F@H usage here would have some gatekeeping before the API received a payload that could exploit that vulnerability, I also don't think it is a good idea to leave the attack vector open.

dotsdl:
I agree with @dwhswenson's points above:
1. The `__hash__` method should stick to using `hash` internally on a tuple of hashables that define the object. This is in keeping with expected behavior for this method.
2. Our "stable" hash for cross-python, cross-process use on these objects should be distinct from that produced by `__hash__`, though it could digest the same (albeit serialized) attributes.

I found this [blog post](https://death.andgravity.com/stable-hashing) as good fodder for this topic, and think we can agree that producing a stable hash for these objects is directly tied to our approach for reliable serialization of the same objects. If we can produce a consistent JSON form for each `Component` subclass, this may accomplish both our serialization scheme for cross-process transmission (and `ResultStore` representation) as well our stable hash digest.

PR #25, opened 2022-05-31
added Failure result

richardjgowers:
I think it might make sense for the result of an individual piece of work to return either a "good" or "bad" results object.  I think Exceptions are enough for some failure reasons (out of disk space, incomplete inputs, unsuitable compute environment) but by returning more than just an exception, you could make the chemistry failures a lot more useful.

It could be that we can make the `ok()` method below just an attribute of a single Result class, since the schema for a `ProtocolUnitResult` is fairly loose/nonexistant.  I'm undecided if there's value in having two schema, one for results and one for error introspection.

One annoyance of first class errors is we'd have to teach schedulers how to identify these so they correctly interrupt a dependency chain.  I think there's probably a class of errors that we don't swallow and pass onto the scheduler directly, and it can probably have a good crack at rerunning them, and there's probably a class of chemistry errors which are fatal.

### DIFF
--- a/gufe/protocols/__init__.py
+++ b/gufe/protocols/__init__.py
@@ -1,4 +1,4 @@
 from .protocol import Protocol, ProtocolResult
-from .results import ProtocolUnitResult, ProtocolDAGResult
+from .results import ProtocolUnitResult, ProtocolUnitFailure, ProtocolDAGResult, ProtocolDAGFailure
 from .protocoldag import ProtocolDAG
 from .protocolunit import ProtocolUnit
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -58,6 +58,8 @@ class ProtocolDAG:
                     for d in self._graph.successors(unit)
                 ]
             )
+            if not result.ok():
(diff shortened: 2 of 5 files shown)


richardjgowers:
@dotsdl worth thinking about how we want to capture errors

dotsdl:
Reviewing now!

dotsdl:
@richardjgowers I've given this a play and made some changes, and also added a test illustrating usage. I *think* this may give us all we need for passing back useful failure data to `ResultServer` within a `ProtocolDAGFailure`, so long as we can serialize anything that gets packed into an exception by a failed `ProtocolUnit._execute` call. 

I really like having distinct `ProtocolUnitFailure` and `ProtocolDAGFailure` classes as failure-oriented subclasses of `ProtocolUnitResult` and `ProtocolDAGResult`, respectively. I think this gives us a lot more flexibility than other approaches we might take for encoding failure.

Happy to elaborate further if needed.

dotsdl (approved):
I'm happy with this implementation, but happy to iterate more here if necessary. Great work @richardjgowers!

richardjgowers:
thanks @dotsdl !

PR #28, opened 2022-06-08
Fix coveragerc ellipsis exclusion

dwhswenson:
`exclude_lines` in `.coveragerc` is a list of regular expressions, not literal strings. Therefore including the string `...` caused all lines to be excluded from coverage. This fixes the problem.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,4 +9,4 @@ exclude_lines =
     pragma: no-cover
     -no-cov
     raise NotImplementedError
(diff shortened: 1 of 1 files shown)


IAlibay (approved):

dwhswenson:
@richardjgowers @dotsdl : I'd rather not self-merge this one, although it's pretty trivial. Could one of you take 30 seconds on this?

PR #30, opened 2022-06-21
[WIP] Storage-ready serialization

dwhswenson:
Includes #21; will be easier to review after that is merged.

This adds a new serialization cycle that (1) enables our objects to use the "path" identifiers described in #21; (2) allows our objects to ensure that a given path is only saved once, allowing (e.g.) protein files to be re-used instead of storage mutliple redundant copies.

The mechanism for this is a serialization instance method called `to_storage_ready`, which returns a dict mapping the object itself (since all our objects are hashable) to a namedtuple containing the bytes serialization, the path identifier, a cryptographic digest used as a label, the class of the object, and other possible metadata. If object A contains object B, and object B is to be stored in a separate file, then object A only stores the metadata about object B, and object B becomes another entry in the output dict. This dictionary can easily be ingested by the the storage system in #21, which just needs path and bytes.

The deserialization classmethod, `from_storage_bytes`, takes the bytes outputted by `to_storage_ready` as well as a function that can take a "path" identifier and return a the byte representation (in general, this function will be `results_client.load_bytes` for some instance `results_client`; passing this will be done behind the scenes).

Objects that have widely interoperable serialization formats (e.g., as SDF) are stored in those formats. Other objects (e.g., things specific to gufe) are stored as JSON objects.

In this PR, I'll add the core machinery, as well as support for several of our setup objects:

- [x] `SmallMoleculeComponent`
- [x] `SolventComponent`
- [x] `ChemicalSystem`
- [ ] `Protocol`
- [ ] `Transformation`

So far, I'm excluding `ProteinComponent` because I'm not yet sure how we're serializing those.


gufe/chemicalsystem.py
```diff
@@ -29,6 +31,8 @@ class ChemicalSystem(Serializable, abc.Mapping):
 
     """
 
+    _storage_path = "setup/systems/{md5}.json"
+
```
richardjgowers, line 35: Can we instead store this on the storage side of things? So storage has a dict of `{Component: _storage_path}`

dwhswenson: That would mean that external packages need to also modify our storage when they import their package. I'd rather have everything contained in their class.

richardjgowers (changes requested):
I'm not a fan of having to add a second set of serialisation methods onto objects.  If a storage backend wants extra data (e.g. `SerializationInfo`) then it can add it itself.  Can we move `_storage_path`, `to_storage_ready` and `from_storage_bytes` into `storage/`?  All components etc should work fine without knowing about storage, they just have to define a to/from dict.  Storage needs to cater for these objects, not the other way round.

dwhswenson:
Sorry that this is going to be a long comment, but I spent a lot of time trying to figure out ways to work around the problems I see, so this is kind of like a move decision tree where I have to tell you why every position I ended up in evaluated poorly in my view.

Original goal was to provide possible remediation strategies -- I'm too tired for that (beyond the approach that was already underway)

## tl;dr

Moving these functions completely into storage will do the following:

* Break discoverability.
* Break duck typing.
* Require registering all classes, which adds friction for external contributions
* Make it harder to provide alternate storage systems
* Break inheritance (unless you do a custom implementation of inheritance, and even that is going to have surprising behavior for users)

Options that I think might better meet what you want:

* Drop `to_dict` and `from_dict` in favor of the more generalized serialization methods I'm writing here (inline the dict code into these methods). This is the direction I was trying to move here.
* Maybe do a shallow `to_dict`/`from_dict`, and also require all objects carry a little extra information (specifically: label template and default values of `to_dict`). Then you can have storage be a consumer of these objects. (This more like what I've done in previous serialization systems, but I don't have the space to explore this now.)
* It might be possible to use a nested `GufeObject.Serialization` class, which contains the extra info/methods. It could be used by, e.g., `system.Serialization().get_storables(system)` or `ChemicalSystem.Serialization().from_storage(serialized_bytes, load_func)`. (Just thought of this; not entirely sure that it would work; no space to explore.)

## 1. Background: Goals and requirements for storage

### 1.1 We want to avoid duplication

Some of our objects may be common across many edges in a simulation campaign (e.g., protein representation, protocol settings, etc.) We don't want to save a new copy of these every time we save a `ProtocolDagResult`, so **we need a mechanism for storing/loading by reference** to a serialized object.

### 1.2 We might not want all objects to be stored by reference

There may be a point where we don't want separate data chunks (files) for each object. This is particularly true in settings: we might prefer to serialize the entirety of the protocol settings as a single data chunk. The reason is that, although this contains many small, reusable components, we might not particularly care that we're reusing a particular state point. In particular, one issue is that this result in the creation of many files in filesystem-based storage. **We need a way to specify which objects should be saved by reference.**

A reasonable alternative here is to say it's "turtles all the way down," and save everything by reference if we can (with the downside that it creates a lot of files).

### 1.3 Objects to store by reference might be in containers

Consider `ChemicalSystem`. We want to store its `components` by reference -- but these are the values in a dict. So **we need to be able to identify objects in containers that should be saved by reference**.

### 1.4 We use labels to store our objects

#### 1.4.1 Data chunks are labelled

Since we are storing objects be reference, we need some way to refer to them. In keeping with the path-like labels we're using for result data, we use a path-like labels for our objects. **We need a way to get the label for a given instance.** We take the `setup/` base "namespace" for this. For convenience, we further divide this namespace into other subpaces. Then each object is placed somewhere within this namespace with a file-like name. For example, a small molecule might have the label:

```python
"setup/components/{object_identifier}.sdf"
```

#### 1.4.2 A stable label

We want the object identifier to be stable across sessions, object creations, versions of our code.

Things that didn't work:

* Python's `hash`: Not stable across sessions (salted with session identifier because of security concerns)
* UUID: Not stable across creations. That is, if you create the same object twice, it will have a different UUID.
* md5 (or other hash) of the JSON string for the `to_dict()` method: Not stable across minor versions of gufe. (Later minor versions may add new fields.)

The solution we've proposed is a cryptographic hash based on md5 of the non-default values in the `to_dict` of an object. While more complicated to say, it's basically a version of the last option that can preserve its value across changes in minor version. 

This means that **we need to determine which attributes are defaults and which are not.** The approach I've take here is to ask the writer of the class to create a dict that maps the keys from `to_dict` to their default values, if such defaults exist. In many cases, this is trivial: see the `defaults_from_init` method for the convenience that works in many cases. However, there are plenty of cases where we cannot infer the defaults:

```python
class Foo:
    def __init__(self, dict_attr=None):
        if dict_attr is None:
            dict_attr = {}
        self.dict_attr = dict_attr
```

A placeholder value can also correspond to more complicated, non-truthy attributes as well, so we can't just use truthiness here -- indeed, if the default is `None` and the user gave `0`, these might have totally different meanings, so we can't make assumptions.

The information about what attributes are default is class-specific. Therefore it is an input from class writers to the storage system.

## 2. Why I don't like the suggested approach

### 2.1 This breaks discoverability

By keeping everything needed to implement a class in one place, we say to potential contributors "you can use this as a template for what you should do." The audience I'm thinking about here is a contributor who isn't a Python expert, and who just wants to toss in a quick addition. How does this contributor learn what needs to be done to add a new class?

Separating the serialization from the class immediately loses some discoverability. If I ask, "How do I serialize/deserialize this class?" I usually use something like tab completion to browse the class's methods. This approach hides that away. Even IPython's `??` won't show it.

There are two choices of where to put the serialization functions: in the same module as the class, or somewhere under `storage`. If you put it in the same module as the class, then I'm not sure what the point of separating them is, and I still don't like that the class isn't self-contained. If you put them in `storage`, then you really break discoverability. It means that a contributor who just wants to make a *narrow* contribution would need a *wide* knowledge of our code base. This seems like it would make it harder to onboard contributors, and be a likely source of confusion.

### 2.2 You can't use duck typing

The approaches I'm suggesting allow a contributor to add a new method that "just works" via duck typing. On the other hand, moving everything to storage requires a registration step, which makes it so that duck typing is not sufficient. New objects are not drop-in replacements for old ones, even if they implement the same API. I think a design that allows duck typing would be preferred to one that does not. Some of the issues below are specific examples where duck typing would solve the problem.

### 2.3 Registering classes creates an unnecessary extra step

For users to use classes from an external package, they'll need to register the class. For some external package, somewhere there needs to be code equivalent to this:

```python
from externalpackage import Thing, serialize_thing, deserialize_thing, THING_DEFAULTS
import gufe.storage
gufe.storage.register(
    serializers={Thing: serialize_thing},
    deserializers={Thing: deserialize_thing},
    template_labels={Thing: "things/{object_id}"},
    defaults={Thing: THING_DEFAULTS},
)
```

Let's hope they don't make their users do that! But we have to tell the devs "in addition to writing your code, you need to write serialization registration code." 

If the developers of `externalpackage` are at least somewhat kind, they'll put that in a `register` function. Still, it's an extra step for a user on import:

```python
import externalpackage
externalpackage.register()
```

If they want to get rid of this, those devs might stick that in as a side effect during import. Side effects during import are generally considered a bad practice in Python, and I don't think we should encourage it. In this particular case, it isn't too terrible (since the side effect should be idempotent), but it is still a messy step that adds friction for an external contributor.

### 2.4 Registering classes makes it harder to use an alternate storage system

The additional information we provide is exactly the information that would be needed by an alternative storage system (where "storage" here means a replacement for our entire stack, from `ResultsClient` down). Note that the code above was for `gufe.storage`. If the developers of `externalpackage` also wanted to support `alt_storage`, they need to duplicate the registration code above to register with `alt_storage`. So `alt_storage` is no longer a drop-in replacement for our storage.

### 2.5 You can't use inheritance

Because subclasses might have different fields that need to be serialized, the lookup needs to be by class. So making a subclass (even one that doesn't change serialization) would require a separate registration.

To an extent, you can fake inheritance by manually traversing the MRO and trying a lookup for each class, taking the first one that succeeds. (This might not work for classes whose metaclass has a custom `issubclass`, but that's an edge case I'm willing to overlook.) This fake inheritance seems potentially confusing to me: how will a user know where to look when they encounter a bug in their serialization? When the relevant method is an instance method, I can easily find where it is defined.

### 2.6 Dropping consistent labels breaks potential use cases

I think there's some confusion about `_storage_path`. First, that's a class variable because it is intended to be something that could (in unusual circumstances) by modified without getting into the core code. It could have just as easily been hard-coded in the `to_storage_ready` method. Second, I think there's a misunderstanding: this is not a filename. Maybe it would be have better to call it `_label_template`: it is the template for the label that the storage uses to label its data.

Having that be consistent between storage systems makes sense. As I've said (over and over), that labelling schema is essentially part of the gufe API. How a storage system addresses these internally is not my problem, but since we deliver bytes that include these labels, those bytes shouldn't be dependent on the storage system used.

Weird use case: with consistent labels, you mix different storage systems as fallbacks. We could make `fun_storage`, that uses our `ResultsClient` as a first place to search, and then falls back on `alt_storage` if it couldn't load something.

## 3. What are the other options

### 3.1 Drop `to_dict`/`from_dict`

Currently, we could fold `to_dict` and `from_dict` into the other serialization methods. This is what I was generally moving toward. Right now, I think they're only used within these serialization methods. I think there's room for a cleaner API, but this might solve some problems.

### 3.x Other options

Other options exist. I want to complete these comment today for further discussion, and can't write a more exhaustive summary.

gufe/solventcomponent.py
```diff
@@ -145,3 +151,13 @@ def to_dict(self):
...
                 'ion_concentration': ion_conc,
                 'neutralize': self._neutralize}
+
+    def to_storage_ready(self):
+        default_keys = get_defaults_from_init(self, self.to_dict())
+        return generic_to_storage_ready(self, self._storage_path,
+                                        default_keys)
+
+    @classmethod
+    def from_storage_bytes(cls, serialized_bytes, load_func):
```
dotsdl, line 161: Is it necessary for `load_func` to be a parameter here? From the tests it looks like it should be a detail of the `from_storage_bytes` method implemented for a given component. It's really not clear to me externally what it should be for a given component, and I think `from_storage_bytes` should only require `serialized_bytes` to do its job.

dwhswenson: Example -- Say I have this:

```python
class A:
    foo: str

class B:
    a: A
```

Let's assume that we also have the relevant serialization stuff. We serialize these objects separately (maybe instances of `A` are frequently reused), so when we serialize an instance of `B` we get 2 JSON files:

myroot/HASH_FOR_B.json

```json
{ "a": {":module:": "__main__",
        ":class:": "A",
        ":path:": "HASH_FOR_A.json"}
}
```

myroot/HASH_FOR_A.json

```json
{"foo": "user input"}
```

So here's why we need more than just the `serialized_bytes`: when you try to deserialize the instance of `B`, `serialized_bytes` is only the contents of `HASH_FOR_B`. The value of the `A`'s `foo` is not included. You need to know where to go to look up a label (`:path:`) -- and that's what the `load_func` does. It takes a label and returns that information. `load_func` will typically be an instance method of the storage object.

dwhswenson:
> I'll do this as a PR against this branch. This will use dask's tokenization approach that we are using in #36.

Quick comment on this point, @dotsdl -- Our goal was to have tokens that are persistent across (minor) versions. See discussion in 1.4.2 of my long comment. I'm not sure that dask's tools can do that.

gufe/storage/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    @property
+    def md5(self):
+        return self.metadata[':md5:']
+
+    @staticmethod
+    def create_metadata(obj, bytes_data, path_pattern):
+        md5 = hashlib.md5(bytes_data).hexdigest()
+        cls = obj.__class__
+        return {
+            ":path:": path_pattern.format(md5=md5[:10]),
```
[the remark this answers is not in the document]
dwhswenson: Agreed -- this should be the `base64.b32encode` or `base64.b32hexencode` of the full md5. (Using b32 so we don't need to worry about case-dependent labels).

I did this as is simply because I didn't want full hex versions; figured if GitHub gets by on 7 hex digits, 10 would probably be enough.

gufe/chemicalsystem.py
```diff
@@ -29,6 +32,8 @@ class ChemicalSystem(Serializable, abc.Mapping):
 
     """
 
+    _storage_path = "setup/systems/{md5}.json"
```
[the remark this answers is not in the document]
dwhswenson: The name here is misleading. It should be called `_label_template`. Regardless of what storage we're using, we have a universal label for these objects. This is the same idea as the labels we use for data, which must be known by the `Protocol` and its `Results` object.

To use the storage we've been setting up so far, we need these labels. My thoughts here were:

* it would be nice if those labels were at least somewhat informative
* it would be nice if those labels carried file extensions so people could figure out what to expect

My approach was to create an implicit structure like:

```
setup/
  components/
    ASDF.sdf
    FGHJ.json
    HJKL.sdf
  settings/
    ZXCV.json
  systems/
    QWER.json
    VBNM.json
  transformations/
    UIOP.json
```

The alternative is just have the label be the MD5 representation -- no namespace, no extension:

```
ASDF
FGHJ
HJKL
QWER
UIOP
VBNM
ZXCV
```

From the computer's standpoint, this makes no difference. From the user's standpoint, I think having the structure is more clear. Those labels are what are written into the storage as references to objects; how they are interpreted as file paths or whatnot is up to the storage backend. I think that having some sense of the structure may be particularly useful when debugging.

To have that structure we need a mapping of class to something like those strings: specifically something which provides the namespace where the objects should live and the file extension. If that mapping is outside the class itself, we hit all the issues I listed in section 2 of the [long comment above](https://github.com/OpenFreeEnergy/gufe/pull/30#issuecomment-1172045531). 

In particular, I think the concerns from @dotsdl and @richardjgowers are about the possibility of swapping out for an external storage stack. But as I discuss in sections 2.4 and 2.6 of that comment, using a consistent label for these makes it easier to use an external storage stack.

Making that mapping of class to label template be part of storage means that each storage implementation must be aware of every contributed package (or each contributed package must be aware of every storage stack). Additionally, as soon as one storage package uses a different labeling scheme, we lose some potentially interesting functionality that is available if all storage systems agree on the label. Furthermore, it might be necessary for that storage to maintain a mapping of MD5 hash to its internally-defined label, e.g.,:

```
{ "ASDF": "setup/components/ASDF.sdf",
  "FGHI": "setup/components/FGHI.json",
   ...
}
```

These issues are avoided by simple including this class variable, (which, in most cases, can probably be set in a superclass, so that contributors don't even need to know about it).

---

**tl;dr: I think we have 3 options:**

1. What I'm doing. Make a user-friendly label part of the object.
2. Drop user friendly labels. Sacrifice the user-friendliness of the labels, but this keeps the idea of label out of the objects.
3. Move the user-friendly label to storage. This isn't impossible, but it will make life harder for contributors and for us. I think this is the worst option.

dotsdl:
@dwhswenson the approach I'm pursuing (if it works out) should avoid the need for tokenization to be stable over time, so we can relax that requirement. We'd only use tokenization to avoid duplication in the storage system where possible, but it is only an optimization, not a guarantee. We won't rely on hashing/tokenization when retrieving anything; we could use UUIDs instead, but we would lose out on that storage space optimization.

I'll be pulling this thread for most of the rest of the day today. Wish me luck :sweat_smile:; will report back on where this landed, including downsides to the approach, in a PR.

dwhswenson:
We have data labeled as:

`transformations/{TRANSFORMATION_HASH}/{RUN_REPLICA}/{EXTENSION}/{UNIT}/file.data`?

If I can't get from `Transformation` to `{TRANSFORMATION_HASH}`, then how does `Results` know how to find all of the results for a transformation? Does it need to recreate every `Transformation` in the database, then do an (I think) O(N^2) operation to create equivalence classes by equality? After that it could loop through those to collect all the results.

Having a stable hash makes that an O(1) process.

richardjgowers:
A lot of the complexity here seems to come from wanting to avoid duplication in the storage medium, but this seems like a problem for the storage system to solve, not something that individual components need to solve.  All that should be required of objects that want to be stored is for them to define `to_dict` and `from_dict`.

If `to_storage_ready` was documented I might understand the problem better?

mikemhenry:
Avoiding store duplication IMHO feels like premature optimization at this point, I think starting out with that goal was a good one but it has started to reveal a lot of complexity. If people want to save money on storage they can just use a zfs deduplication setup and instead of spending money on storage they can spend money on ECC ram.

dwhswenson:
Closing this. We're relaxed the following requirements I was operating under, which made some of the additional complexity necessary:

* not adding a base class
* storing molecules as SDF
* not changing the existing `to_dict`/`from_dict` methods

PR #31, opened 2022-06-21
to_dict on unitted quantities should use to_tuple

dwhswenson:
Life is a lot easier if we play by the rule that the dict that comes out of to_dict should be JSON serializable. Quantities are not.

Choice is either create custom JSON encoders/decoders, or require to_dict to return something JSON serialization.

In this PR, I only changed SolventComponent.ion_concentration. I may come across others as I work my way through storage.


richardjgowers (approved):

mikemhenry:
I think doing what pint recommends is a safe bet

PR #32, opened 2022-06-22
Change ion defaults

dwhswenson:
If both ions are `None` and `neutralize=True` (as is the default) behavior is unclear if the system needs to be neutralized. This changes the defaults to "Na+" and "Cl-".

Second commit here disallows `None` as a valid value. I don't think there's any reason to allow it if it isn't the default, right?

While I'm at it, should I also change default on `ion_concentration` to 0.0 molar, instead of `None`?

### DIFF
--- a/gufe/solventcomponent.py
+++ b/gufe/solventcomponent.py
@@ -28,8 +28,8 @@ class SolventComponent(Component):
 
     def __init__(self, *,  # force kwarg usage
                  smiles: str = 'O',
-                 positive_ion: Optional[str] = None,
-                 negative_ion: Optional[str] = None,
+                 positive_ion: Optional[str] = 'Na+',
+                 negative_ion: Optional[str] = 'Cl-',
                  neutralize: bool = True,
                  ion_concentration: unit.Quantity = None):
         """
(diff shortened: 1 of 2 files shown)


richardjgowers (approved):

richardjgowers:
WRT concentration, there is an argument that None means not-specified whereas 0.0 means actually 0.  Should be harmless to set the default to 0.0 though.  Something I think you've brought up before that I'm not sure on is, does that concentration include the ions required to neutralise? Because with 0.0 arguably you can't neutralize if it does include those ions. @IAlibay

dwhswenson:
> WRT concentration, there is an argument that None means not-specified whereas 0.0 means actually 0. Should be harmless to set the default to 0.0 though. Something I think you've brought up before that I'm not sure on is, does that concentration include the ions required to neutralise? Because with 0.0 arguably you can't neutralize if it does include those ions.

I believe we're currently just wrapping `openmm.app.Modeller`, which is [explicit in its docs on this](http://docs.openmm.org/7.5.0/api-python/generated/simtk.openmm.app.modeller.Modeller.html#simtk.openmm.app.modeller.Modeller.addSolvent):

> **ionicStrength** (*concentration=0\*molar*) – the total concentration of ions (both positive and negative) to add. This does not include ions that are added to neutralize the system. Note that only monovalent ions are currently supported.

IAlibay:
So in practice neutralize=True with 0 mM ionic concentration just means add the relevant ions to make a neutral system. That's pretty much the standard for it. When you increase the ionic concentration then you essentially adjust the number of counter ions to aim for that ionic concentration. How that happens in practice is a bit messy, you can look at this code to get an idea of how different methods behave: https://github.com/IAlibay/saltpy/blob/main/saltpy/estimators.py

dotsdl (approved):
Looks good @dwhswenson! Merging!

PR #33, opened 2022-06-23
Add PerFileJSONMetadataStore

dwhswenson:
(This includes and extends #21, and should be merged/reviewed after that one is merged.)

This adds `PerFileJSONMetadataStore`. The original `JSONMetadataStore` wrote all metadata (hashes) to a single JSON file. That is unlikely to be optimal for many workflows. The approach used in this object creates a separate label at `metadata/{filename}.json`, corresponding to a JSON file with an object carrying the metadata. This way, we don't need to completely re-export all metadata every time we store an new object.

### DIFF
--- a/gufe/storage/metadatastore.py
+++ b/gufe/storage/metadatastore.py
@@ -7,7 +7,9 @@ import collections
 from typing import Tuple, Dict
 from .externalresource.base import Metadata
 
-from .errors import MissingExternalResourceError
+from .errors import (
+    MissingExternalResourceError, ChangedExternalResourceError
+)
 
 
 class MetadataStore(collections.abc.Mapping):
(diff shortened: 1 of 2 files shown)


dotsdl:
I think this should move out of `gufe`, probably to `openfe`?

richardjgowers:
@dwhswenson stale or good to go?

dwhswenson:
@richardjgowers : IMO, good to go, and then once we have all the currently open storage PRs in, I'll migrate most of `storage` over to `openfe`. The only thing that stays is `externalresource` stuff, because that will be used when we switch `shared` from filelike to something more abstract (#120).

richardjgowers (approved):

PR #34, opened 2022-06-24
S3Storage ExternalStorage subclass

dotsdl:
Includes #21; that PR should be merged before this one.

Added S3Storage(ExternalStorage) subclass.

This enables use of an S3 bucket and prefix as a storage backend for file/byte data.

## Todo
- [ ] unit tests for this component


dotsdl:
Closing since we want to minimize storage system(s) code in `gufe`; this will live in `fah-alchemy` for now.

PR #35, opened 2022-07-08
added ABCs for AtomMapping behaviour

richardjgowers:
These are the (proposed) abstract objects that live above our current LigandMapping thing in OpenFE.

I'm calling everything `Atom` here, but maybe it should be more generic to handle A) CG stuff and B) where components aren't even referring to particles (does this exist?).


IAlibay:
didn't read the code but atom is probably fine. I can't think of a current use case for mapping where you don't have atoms, except maybe if you want mappings to work in absolute cases (i.e. one end state is just nothing)

gufe/mapping.py
```diff
@@ -1,19 +0,0 @@
...
-
-from openff.toolkit.utils.serialization import Serializable
-
-
-class Mapping(Serializable):
-    """A mapping, usually of atoms, between two `ChemicalSystem`s.
-
-    Some, but not all, transformation protocols require a mapping to execute.
-
-    """
```
richardjgowers, line 12: So here @dotsdl was thinking about mappings as between Systems (collection of components) rather than just two Components.  Worth discussing.

IAlibay: Yeah so in the context of gufe, I agree with this. Atom mappings may span multiple components (i.e. you _should_ have the ability to perturb a side chain whilst also perturbing a ligand, or in the more reasonable case, perturb an ion to account for a ligand with a net charge changing at the same time).

Having the atom mapping be system wide could be easier to control the "link" in the mappings between the various components? Alternatively you could just bake in a reference to the component being perturbed in the mapping object and then you'd feed in a frozenset of mapping objects or something like that?

RiesBen: I have a stupid question/point regarding the first comment by @richardjgowers. A non particle related mapping could be something like a thermodynamic change  (temperature or permitivity) or a conformational description (alpha-helix  or pi-helix).  Is that something you were searching for?

IAlibay: possibly but thermodynamic control directly, but one may possibly want to have some means of encoding some collective variable of interest maybe (including possibly a titration experiment in a CpH,EMD style calculation). That being said I don't think this really should be under "mapping". I'm not sure we've really started fleshing out things beyond alchemical control but those two cases probably should go under a separate CV class.

One type of thing we could conceivably think as going under a "mapping" are restraints. One could think of applying bond/angle/dihedral restraints as a type of mapping.

RiesBen: shamless plug for an restraint finding algorithm and a "empty mapping case": https://github.com/rinikerlab/restraintmaker

gufe/mapping/systemmapping.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    stateA: ChemicalSystem
+    stateB: ChemicalSystem
+    mappings: list[ComponentMapping]
+
+    def __init__(self, stateA: ChemicalSystem, stateB: ChemicalSystem,
+                 mappings: Iterable[ComponentMapping]):
+        self._stateA = stateA
+        self._stateB = stateB
+        self._mappings = tuple(mappings)
```
richardjgowers, line 24: change to dict

richardjgowers:
To enable more flexibility, mapping is now defined as `dict[str, ComponentMapping]`.  ComponentMapping is very abstract (e.g. to allow mapping between `SolventComponent`.)  `AtomMapping` is the int to int relationships we're used to.

The gist of what this PR enables is:

```python
p1, p2 = gufe.ProteinComponent()
l1, l2 = gufe.SmallMoleculeComponent()
sol = gufe.SolventComponent()

sys1 = gufe.ChemicalSystem({'protein': p1, 'ligand': l1, 'solvent': sol})
sys2 = gufe.ChemicalSystem({'protein': p2, 'ligand': l2, 'solvent': sol})

# defining the changes
lm = gufe.LigandLigandMapping(molA=l1, molB=l2, molA_to_molB={...})
# I imagine we'll store indices differently here...
sidechain_map = gufe.ProteinMapping(molA=p1, molB=p2, molA_to_molB={...})


system_mapping = {'ligand': lm, 'protein': sidechain_map}

protocol = Protocol()

dag = protocol.create(sys1, sys2, mapping=system_mapping)
```

gufe/protocols/protocol.py
```diff
@@ -155,14 +155,14 @@ def _create(
         self,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Optional[Mapping] = None,
+        mapping: Optional[dict[str, ComponentMapping]] = None,
```
richardjgowers, line 158: @dotsdl this is the big change I'm making to Protocol API here.  Previously `Mapping` here wasn't actually well defined/respected.

gufe/mapping/atom_mapping.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @property
+    def molB(self) -> gufe.Component:
+        """A copy of the second Component in the mapping"""
+        return self._molB
+
+    @property
+    @abc.abstractmethod
+    def molA_to_molB(self) -> Mapping[int, int]:
+        """Maps the index of an item from Component A onto Component B
```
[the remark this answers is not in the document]
IAlibay: Or we switch completely to calling them "components" (I think that'd be worse for users, but maybe better for devs)

gufe/tests/test_mapping.py
```diff
@@ -0,0 +1,64 @@
...
+    def molB_to_molA(self):
+        return {v: k for k, v in self._mapping}
+
+    def molA_unique(self):
+        return (i for i in range(self._molA.to_rdkit().GetNumAtoms())
+                if i not in self._mapping)
+
+    def molB_unique(self):
+        return (i for i in range(self._molB.to_rdkit().GetNumAtoms())
+                if i not in self._mapping.values())
```
IAlibay, line 45: Given their implementation isn't defined in the abstract class I don't really care too much here, but is there worth in testing these too for the sake of ensuring we don't introduce breaking changes without knowing?

IAlibay (changes requested):
Only blockers here imho are docs related, I'd like to make sure this is easy to understand for devs and users since it'll be something both will interact with.

There's a couple extra bits, one clarification on the requirements for `_mappings_generator` and another on whether we should add a few extra tests for the expected methods for the mapping object.

gufe/mapping/atom_mapper.py
```diff
@@ -0,0 +1,14 @@
...
+import abc
+from collections.abc import Iterator
+import gufe
+
+from .atom_mapping import AtomMapping
+
+
+class AtomMapper(abc.ABC):
+    """A class for manufacturing mappings"""
+    def suggest_mappings(self, A: gufe.Component,
```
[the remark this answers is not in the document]
richardjgowers: I don't see it as fundamental. We have it for how our ligand mappers work, but it's entirely possible to come up with alternate patterns. The strict requirement is this API point.

gufe/mapping/atom_mapping.py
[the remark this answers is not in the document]
richardjgowers: I think somewhere I realised Molecule is a loaded term (refers to class names elsewhere, and/or is a single connected molecule) Component less so.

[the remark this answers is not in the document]
IAlibay: Component isn't going to float well with users (it doesn't really mean much). Subclasses of this will be interfaced with by users, so we should probably should come up with definite semantics here.

gufe/protocols/protocol.py
```diff
@@ -155,14 +155,14 @@ def _create(
...
         extend_from: Optional[ProtocolDAGResult] = None,
     ) -> List[ProtocolUnit]:
         """Method to override in custom `Protocol` subclasses.
 
         This method should take two `ChemicalSystem`s, and optionally a
-        `Mapping`, and prepare a collection of `ProtocolUnit` instances that
-        when executed in order give sufficient information to estimate the
+        dict of `Mapping`, and prepare a collection of `ProtocolUnit` instances
```
dwhswenson, line 164:
```suggestion
        dict mapping string to ``ComponentMapping``, and prepare a collection of ``ProtocolUnit`` instances
```

gufe/mapping/atom_mapping.py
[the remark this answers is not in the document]
dwhswenson: It felt to me like the `Component` idea went pretty well when I explained it in my talk in Amsterdam, so I think users will get it.

I agree that `molA` is misleading when the thing should be `componentA`.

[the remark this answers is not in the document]
IAlibay: I disagree that `Component` is reasonable for users, especially given that you always do perturbations on molecules (you won't even just perturb the entire solvent for example, that mapping is purely on the ligand). That being said I also just want this merged. I'd prefer we just stick to _something_ for now, and then I can just poll all the FE users in our lab to tell us why neither name makes sense.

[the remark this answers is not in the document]
richardjgowers: As long as (up to) a biological assembly can be referred to as a mol_A that's fine.

gufe/mapping/atom_mapping.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Not all indices will be resolvable, these items are not mapped,
+        resulting in a KeyError
```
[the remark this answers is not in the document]
richardjgowers: It could return None for missing or raise a KeyError, either is fine, I thought KeyError is more dicty, but not attached to either.

gufe/mapping/atom_mapping.py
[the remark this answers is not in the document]
IAlibay: Right because we could just throw in an entire protein component and just expect this to fall within this map? (waters and all)

I'm not sure that's super digestible from a user perspective, but I guess we've already bitten the bullet (or at least there's no easy way around this).

gufe/tests/test_mapping.py
richardjgowers: this should get tested in openfe for the actual implementations we use, these are just examples

gufe/mapping/atom_mapping.py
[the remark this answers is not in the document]
richardjgowers: Ok I've gone with Component, if everyone hates it we change it in v2

gufe/mapping/atom_mapping.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        ...
+
+    @property
+    @abc.abstractmethod
+    def molB_to_molA(self) -> Mapping[int, int]:
```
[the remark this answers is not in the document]
richardjgowers: it's abstract, so the docstring gets rewritten anyway

richardjgowers:
ok @IAlibay hopefully all happy here now

gufe/mapping/atom_mapping.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    @abc.abstractmethod
+    def componentA_unique(self) -> Iterable[int]:
+        """Indices of atoms in mol A that aren't mappable to B"""
+        ...
+
+    @property
+    @abc.abstractmethod
+    def componentB_unique(self) -> Iterable[int]:
+        """Indices of atoms in mol B that aren't mappable to A"""
```
IAlibay, line 54: here is the other one

IAlibay (approved):
Just a couple of small mol -> component changes in docs please

gufe/mapping/atom_mapping.py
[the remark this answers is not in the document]
IAlibay: We should do "user testing" as part of our v2 goal.

gufe/mapping/atom_mapping.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    @abc.abstractmethod
+    def componentB_to_componentA(self) -> Mapping[int, int]:
+        """Similar to A to B, but reversed."""
+        ...
+
+    @property
+    @abc.abstractmethod
+    def componentA_unique(self) -> Iterable[int]:
+        """Indices of atoms in mol A that aren't mappable to B"""
```
IAlibay, line 48:
```suggestion
        """Indices of atoms in component A that aren't mappable to B"""
```

gufe/mapping/atom_mapping.py
IAlibay, line 54:
```suggestion
        """Indices of atoms in component B that aren't mappable to A"""
```

IAlibay:
There, fixed it for you.

PR #36, opened 2022-07-11
DAGs now constructed using ProtocolUnit inputs; inputs can be mapped into keyword arguments of `ProtocolUnit._execute`

dotsdl:
Previously, we constructed a DAG of `ProtocolUnit`s by instantiating each one separately, e.g.:

```python

    def _create(
        self,
        stateA: ChemicalSystem,
        stateB: ChemicalSystem,
        mapping: Optional[Mapping] = None,
        extend_from: Optional[ProtocolDAGResult] = None,
    ) -> nx.DiGraph:

       # convert protocol inputs into starting points for independent simulations
       alpha = InitializeUnit(
            self.settings,
            stateA=stateA,
            stateB=stateB,
            mapping=mapping,
            start=extend_from,
        )

        # create several units that would each run an independent simulation
        simulations: List[ProtocolUnit] = [
            SimulationUnit(self.settings, window=i) for i in range(20)
        ]

        # gather results from simulations, finalize outputs
        omega = FinishUnit(self.settings, name="the end")

        # create DAG, which encodes dependency between ProtocolUnits
        dag = nx.DiGraph()
        for sim in simulations:
            dag.add_edge(sim, alpha)
            dag.add_edge(omega, sim)

    return dag
```

However, this featured a couple clear shortcomings:
1. DAG construction from `ProtocolUnit`s is a separate set of operations from unit creation, even though a given `ProtocolUnit` would likely require `ProtocolUnitResult`s from others as inputs. 
2. `ProtocolUnit._execute` would get a list of `ProtocolUnitResult`s corresponding to its dependencies in the DAG, but these would have be sorted through by that method based on each `ProtocolUnitResult`'s identifying parameters. This made accessing results from dependencies obtuse at best.

The approach implemented in this PR allows the above to become:
```python
    def _create(
        self,
        stateA: ChemicalSystem,
        stateB: ChemicalSystem,
        mapping: Optional[Mapping] = None,
        extend_from: Optional[ProtocolDAGResult] = None,
    ) -> List[ProtocolUnit]:

        # convert protocol inputs into starting points for independent simulations
        alpha = InitializeUnit(
            name="the beginning",
            settings=self.settings,
            stateA=stateA,
            stateB=stateB,
            mapping=mapping,
            start=extend_from,
        )

        # create several units that would each run an independent simulation
        simulations: List[ProtocolUnit] = [
            SimulationUnit(settings=self.settings, window=i, initialization=alpha) for i in range(20)
        ]

        # gather results from simulations, finalize outputs
        omega = FinishUnit(settings=self.settings, name="the end", simulations=simulations)

        # return all `ProtocolUnit`s we created
        return [alpha, *simulations, omega]
```

In this implementation, we can include `ProtocolUnit` instances that an instance is dependent on as keyword arguments to the instance on init, and the `ProtocolUnitResult` for those dependencies is passed through to `ProtocolUnit._execute` with the same keyword argument names as given on init. We support passing lists and dicts of `ProtocolUnit`s in as keyword parameters as well, with nesting supported.

This approach also encodes the DAG without any additional steps needed. `ProtocolUnit`, `ProtocolUnitResult`, `ProtocolDAG`, and `ProtocolDAGResult` are all `GufeTokenizable`s, and so use the same tokenization system as the rest of `gufe`.


richardjgowers (commented):
Yeah I think we discussed this way back and got stuck on the thing that it sounds like `dask` is solving for us, so this sounds good.

I agree this is nicer to create.  It does mean if you did want the DAG as a `nx.Graph` you'd have to go through the Units again, but maybe that's not awful (as it's only O(N))

dwhswenson:
For a key that encodes the protocol unit, is there a reason not to use the same key we're using for storage?

For example, if I have the location `transformations/{TRANFORMATION_ID}/{RUN_ID}/{CLONE_ID}/{EXTENSION_ID}/{UNIT_ID}/results.json`, if I drop the `results.json` from that "path", I have a unique identifier for the protocol unit.

It feel like this is the same problem that we're solving in different ways in different contexts; maybe better to unite on a shared solution?

dotsdl:
> Yeah I think we discussed this way back and got stuck on the thing that it sounds like `dask` is solving for us, so this sounds good.
> 
> I agree this is nicer to create. It does mean if you did want the DAG as a `nx.Graph` you'd have to go through the Units again, but maybe that's not awful (as it's only O(N))

Excellent, thanks! We actually do get the `nx.Graph`, but it is built outside of `Protocol._create`: `Protocol.create` feeds the list of `ProtocolUnit`s to `ProtocolDAG`, [which builds the `nx.DiGraph` representation up from them](https://github.com/OpenFreeEnergy/gufe/blob/7bf47333f55f68843b90a72d5ec1f72d4b21e518/gufe/protocols/protocoldag.py#L42).

So calling `protocol.create()` gives you a `ProtocolDAG` with an `nx.DiGraph` as its `.graph` property as before.

dotsdl:
I'm pulling my hair out trying to understand how `codecov` thinks we only have 58% coverage here. Running locally I observe 96% coverage, although I've also observed alternating 58% and 96% coverage locally after a few `pip uninstall gufe; pip install -e .`. Really not sure what's going on here.

It's particularly bizarre that codecov says other parts of the codebase we don't even touch in this PR, such as `components`, have a drop in coverage as well. Ideas? 

I tried opening a new PR in #48, with the same result. Suspecting codecov was caching results for commits, I created a squashed version in #49. Still same result.

gufe/protocols/protocoldag.py
```diff
@@ -3,14 +3,93 @@
...
+    ----------
+    name : str
+        Unique identifier for this `ProtocolDAGResult`.
+    graph : nx.DiGraph
+        The `ProtocolUnit`s, with dependencies set, as a networkx `DiGraph`.
+        Each `ProtocolUnit` features its `ProtocolUnitCompletion` as a `result` attribute.
+
+    """
+    name: Optional[str]
+    graph: nx.DiGraph
```
richardjgowers, line 32: These annotations are wrong, need the underscore

richardjgowers, line 32: These annotations are wrong, need the underscore

gufe/protocols/protocoldag.py
```diff
@@ -3,14 +3,93 @@
...
+
+    def __init__(self, *, name=None, graph):
+        self._name = name
+        self._graph = graph
+
+    def _defaults(self):
```
richardjgowers, line 38: Shouldn't this be the base class default?

gufe/protocols/protocoldag.py
```diff
@@ -3,14 +3,93 @@
...
+    def name(self):
+        return self._name
+
+    @property
+    def graph(self):
+        return self._graph
+
+    @property
+    def protocol_units(self):
+        return [pu for pu in self.graph.nodes]
```
richardjgowers, line 60: Maybe returning these in DAG order would be better? (This is built in to nx)

gufe/protocols/protocoldag.py
```diff
@@ -3,14 +3,93 @@
...
+
+    @property
+    def protocol_unit_results(self):
+        return list(nx.get_node_attributes(self.graph, "result").values())
```
richardjgowers, line 64: Is this guaranteed to be the same order as above? (It ought to be)

gufe/protocols/protocoldag.py
```diff
@@ -21,21 +100,60 @@ class ProtocolDAG:
...
+        return {'name': self.name,
+                'protocol_units': self.protocol_units}
+
+    @classmethod
+    def _from_dict(cls, dct: Dict):
+        return cls(**dct)
+
+    def _build_graph(self, protocol_units):
+        G = nx.DiGraph()
+        """Build dependency DAG of ProtocolUnits with input keys stored on edges"""
```
richardjgowers, line 138: Docstring on wrong line will trip up sphinx

gufe/protocols/protocoldag.py
```diff
@@ -45,23 +163,70 @@ def name(self):
     def graph(self):
         return self._graph
 
-    def execute(self) -> Union[ProtocolDAGResult, ProtocolDAGFailure]:
-        """Execute the full DAG in-serial, in process."""
+    def protocol_units(self):
+        return list(self._protocol_units)
+
+    def execute(self, *,
```
richardjgowers, line 169: Not sure I quite get the execute for a whole DAG, it's not really the recommended usage and misleading. I'd turn this into a convenience function rather than a method.

gufe/protocols/protocoldag.py
```diff
@@ -45,23 +163,70 @@ def name(self):
...
+        # TODO: change this part once we have clearer ideas on how to inject
+        # persistent storage use
+        if dag_scratch is None:
+            dag_scratch_tmp.cleanup()
+
         return ProtocolDAGResult(name=self._name, graph=graph)
+
+    def _pu_to_pur(
+            self,
+            inputs,
```
richardjgowers, line 220: Types?

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,312 @@
...
+
+
+class ProtocolUnitResultBase(GufeTokenizable):
+    inputs: Dict[str, Any]   # `ProtocolUnitResult`s this `ProtocolUnitResult` is dependent on
+
+    name: Optional[str]      # name of the `ProtocolUnit` that produced this `ProtocolUnitResult`
+    pure: bool               # whether `ProtocolUnit` that produced this `ProtocolUnitResult` was a function purely of its inputs
+    source_key: GufeKey             # key of the `ProtocolUnit` that produced this `ProtocolUnitResult`
+
+    outputs: Dict[str, Any]  # outputs is a dict returned by a `ProtocolUnit`'s `_execute` method
```
richardjgowers, line 38: Maybe instead of Any we can actually have a type which has the accepted types? Might catch some snafus with static typing

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,312 @@
...
+        self._traceback = traceback
+        super().__init__(name=name, source_key=source_key, pure=pure, inputs=inputs, outputs=outputs)
+
+    @property
+    def exception(self):
+        return self._exception
+
+    @property
+    def traceback(self):
+        return self._traceback
```
richardjgowers, line 150: String presumably?

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,312 @@
...
+        ):
+        """Generate a `ProtocolUnitResult`.
+
+        Parameters
+        ----------
+        name : Optional[str]
+            Name of the `ProtocolUnit` that produced this `ProtocolUnitResult`.
+        source_key : GufeKey
+            Key of the `ProtocolUnit` that produced this `ProtocolUnitResult`
+        pure : bool
```
richardjgowers, line 55: Why have pure? Isn't uuid fine at this level? One less branch to think about

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,312 @@
...
+        """
         self._name = name
+        self._pure = pure
+        self._inputs = inputs
 
-    def __hash__(self):
-        return hash(
-            (self.__class__.__name__, self._settings, frozenset(self._kwargs.items()))
-        )
+        self.dag_scratch = None
```
richardjgowers, line 181: Why set these as attributes? Just have them as variables in execute and they are persisted in the Results

dotsdl: Ah, apologies, missed removing these from an earlier iteration. These now make their way into `ProtocolUnit._execute` via `ctx`, a `Context` object.

gufe/protocols/protocoldag.py
dotsdl: It's an abstractmethod in `GufeTokenizable`, so we have to override it for subclasses. I think this is appropriate, as defining defaults generally is hard and should be made on a per-subclass basis.

gufe/protocols/protocoldag.py
dotsdl: Done!

gufe/protocols/protocoldag.py
dotsdl: Now it should be, since we traverse the result graph in the same way. Will add test.

gufe/protocols/protocoldag.py
dotsdl: Ah, good catch!

gufe/protocols/protocoldag.py
dotsdl: Fair enough, will move it to a function so it's not part of the API of `ProtocolDAG`.

dotsdl:
Realized I missed conversion of `Protocol` and `ProtocolResult` into `GufeTokenizable`s. Working on this next.

gufe/protocols/protocolunit.py
dotsdl: You're right that making `ProtocolUnit`s hash based on their inputs isn't strictly necessary, but does offer opportunities for deduplication within a Scheduler. So, it's an optimization.

The way I have it now, `pure` is default `False` (taking cues from `dask.delayed` behavior), so it's opt-in. My view is that if protocol developers don't generally use this feature, we can remove it.

gufe/protocols/protocoldag.py
dotsdl: Removed annotations, since we have properties for documentation anyway.

gufe/protocols/protocoldag.py
dotsdl: Added!

gufe/protocols/protocolunit.py
dotsdl: Can you show me how to do this? My `typing` foo is still a bit limited. Would it be something like:

```
GufeAcceptable = Union[Dict[str,GufeAcceptable],List[GufeAcceptable],int,float,bool,...]
```
?

gufe/protocols/protocolunit.py
dotsdl: Yup!

gufe/protocols/protocolunit.py
dotsdl: I went ahead and removed `pure` as a concept, and switched to just using UUIDs for `ProtocolUnit`s. I think the opportunities for deduplication across DAGs may prove rather limited, so this simplifies things here.

dotsdl:
@richardjgowers and @dwhswenson, this is finally ready for your review. I see one failing test on here that I'm not getting locally, so any insights you have on what may be happening there are appreciated. I will dig into it further myself tomorrow to get it resolved.

dotsdl:
Nevermind, looks like the odd single test now clears on Ubuntu. :shrug:

dwhswenson (changes requested):
Sorry to take so long on this; here it is (finally!) A few minor changes to be made.

I think we've discussed most of the big-picture questions in calls, but to get them written down here -- the following shouldn't block this PR, but should be addressed in the near future:

* I think we'll want to pass "clone ID" / "extension number" (or whatever we end up calling those things) so that they are included in the `ProtocolDAGResult` (or possibly within `ProtocolUnitResult`, depending on how we interpret "extension number"). Those (together with the `Transformation`) should uniquely identify a `ProtocolDAGResult`, and provide full context required when analyzing a set of `ProtocolDAGResult`s.
* We'll want to shift to a one-to-many relationship within a `ProtocolDAGResult` for `ProtocolUnit` to `ProtocolUnitResult`. This is because we may have several failed results for a given unit -- the mapping to *successful* results should be one-to-one. 

Additional minor detail; somewhat personal preference:

* I think the name `dag_scratch` may be too easily confused. I'd prefer to call this something like `staging_area`. [This is a term used for similar purposes elsewhere in computing](https://en.wikipedia.org/wiki/Staging_(data)). This will also make it more natural to change it from a path to a very simple object in the future (to abstract to non-filesystem staging areas) -- I think `unit_scratch` will necessarily always be a filesystem path, but `dag_scratch` might not. Differing naming conventions makes it clearer if these end up being different object types.

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,379 @@
...
+    def __init__(self, *, name=None, source_key, inputs, outputs, _key=None, exception, traceback):
+        self._exception = exception
+        self._traceback = traceback
+        super().__init__(name=name, source_key=source_key, inputs=inputs, outputs=outputs, _key=_key)
+
+    def _to_dict(self):
+        dct = super()._to_dict()
+        dct.update({'exception': self.exception,
+                    'traceback': self.traceback})
+        return dct
```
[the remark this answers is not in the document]
dwhswenson: RE: Serialization of exceptions/tracebacks: look into [tblib](https://github.com/ionelmc/python-tblib)?

gufe/protocols/protocoldag.py
```diff
@@ -2,66 +2,321 @@
...
+            raise KeyError("No such `protocol_unit_result` present")
+
+    def ok(self) -> bool:
+        return True
+
+
+class ProtocolDAGFailure(ProtocolDAGResult):
+    """A failed `ProtocolDAGResult`.
+
+    """
```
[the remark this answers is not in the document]
richardjgowers: I think the original idea was that this mirrored the symmetry between `ProtocolUnitResult` and `ProtocolUnitFailure`.  I think it's probably not necessary and we can just have a single results container `ProtocolDAGResult`.

If we're being pedantic, it's probably best to define `ok()` as there being a 1:1 relationship between units and results (successes)

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,379 @@
...
+        `ProtocolUnitResult` is dependent on.
+    outputs : Dict[str, Any]
+        Outputs from the `ProtocolUnit._execute` that generated this
+        `ProtocolUnitResult`.
+    dependencies : List[ProtocolUnitResult]
+        A list of the `ProtocolUnitResult`s depended upon.
+    exception : Exception
+        The exception raised during `ProtocolUnit` exectuion.
+    traceback : str
+        The traceback given by the exception.
```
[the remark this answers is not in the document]
richardjgowers: I kinda like traceback being `str` as you can stuff whatever debug info might be necessary there.

gufe/protocols/protocolunit.py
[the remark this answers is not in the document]
richardjgowers: would deserialising a traceback require the original dependencies?  i.e. to get an openmm-based traceback would you need openmm installed?  I'm imagining that compute envs might be very different to local envs, and maybe we want results to not require the full stack

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,379 @@
...
-        self._kwargs = kwargs
+        """Create an instance of a ProtocolUnit.
+
+        Parameters
+        ----------
+        name : str
+            Custom name to give this 
+        _key : GufeKey
+            Used by deserialization to set UUID-based key for this
+            `ProtocolUnit` before creation.
```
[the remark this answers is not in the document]
dotsdl: Good call! Adopted this approach!

gufe/protocols/protocolunit.py
```diff
@@ -7,79 +7,379 @@
...
+        Inputs to the `ProtocolUnit` that produced this
+        `ProtocolUnitResult`. Includes any `ProtocolUnitResult`s this
+        `ProtocolUnitResult` is dependent on.
+    outputs : Dict[str, Any]
+        Outputs from the `ProtocolUnit._execute` that generated this
+        `ProtocolUnitResult`.
+    dependencies : List[ProtocolUnitResult]
+        A list of the `ProtocolUnitResult`s depended upon.
+    exception : Exception
+        The exception raised during `ProtocolUnit` exectuion.
```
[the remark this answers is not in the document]
dotsdl: Got it, thank you!

gufe/protocols/protocolunit.py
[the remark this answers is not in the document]
dotsdl: I'm thinking we might need to go for strings for both exceptions and tracebacks here, given @richardjgowers' point above. If we can come up with something more clever later, then great, but I don't think generally we'll be able to roundtrip exceptions reliably.

gufe/protocols/protocolunit.py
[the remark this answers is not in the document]
dotsdl: Kept traceback as string for now, which includes exception information.

gufe/protocols/protocolunit.py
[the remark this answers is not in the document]
dotsdl: I've made `ProtocolUnitFailure.exception` give a tuple instead of an exception object. The first element gives the type of exception raised; the second element is a tuple giving the exception's `args` values. This should be serialization/deserialization friendly while still retaining the information we can from the exception (assuming `args` are of simple collections (lists, dicts, etc.) and primitive types).

dotsdl:
@dwhswenson I changed `dag_scratch` to `shared` and `unit_scratch` to simply `scratch`. I wasn't a fan of `staging_area` because the purpose of `dag_scratch` is to operate as a shared space `ProtocolUnit`s in a `ProtocolDAG`, not necessarily as a staging area for full export somewhere downstream. That's the concept I'm operating with, at least.

dotsdl:
The CI is likely borked here due to `openff-toolkit` 0.11.0. I think fixing it is out of scope for this PR. Tests pass in my local runs.

dotsdl:
Ready to go! @richardjgowers or @dwhswenson, please merge if satisfied.

PR #37, opened 2022-07-21
Add Settings 

mikemhenry:
Todo:
 - Get units working properly, will tackle that working with openff
 - Right now this needs to use the bio-refactor from openff, so that is why changes were made to the env file
 - That last point was more of an FYI and not a todo (and so is this one)

We need to figure out how the settings for the chemical component will fit in.


mikemhenry:
sorry for the wonky git history, I had to revert commits on main, then revert the revert on this branch, and then cherry pick them back in :upside_down_face:

richardjgowers:
@mikemhenry chemical components aren't in `Settings`. A protocol accepts the settings to define where in parameter space it operates:  https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/protocols/protocol.py#L51 Chemical components comes in when you apply the Protocol to a pair of systems: https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/transformations/transformation.py

mikemhenry:
> @mikemhenry chemical components aren't in `Settings`. A protocol accepts the settings to define where in parameter space it operates: https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/protocols/protocol.py#L51 Chemical components comes in when you apply the Protocol to a pair of systems: https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/transformations/transformation.py

So then at settings class instantiation time, we have no way to check if the forcefeild will work, which is fine, but is then the protocol responsible for doing the atom typing/parameter assignment?

richardjgowers:
Yes the Protocol applies the forcefield to the ChemicalSystems.  I think this is linked to when we debated if the nodes in a graph have a FF applied to them, and we ended up saying no to that and that the FF is instead a property of an edge.

mikemhenry:
> and we ended up saying no to that and that the FF is instead a property of an edge.
Yes! Now I remember that. Do we need to then add things like how we want our chemical components solvated in Settings? Or will users specify that somewhere else?

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+    vdW: VdWSettings
+    electrostatics: ElectrostaticSettings
+    bonds: BondSettings
+    angles: AngleSettings
+    proper_torsions: ProperTorsionSettings
+    improper_torsions: ImproperTorsionSettings
+    gbsa: GBSASettings
+
+    # Misc
+    aromaticity_model: Literal["OEAroModel_MDL"]
```
richardjgowers, line 144: Not sure we need this in the FF definition?  Isn't aromaticity already defined in the chemical models (elsewhere)

IAlibay (commented):
just leaving some initial comments - TLDR I think I've completely missed the point of what you're trying to achieve here. My understanding is that these should be simulation control settings but a lot of this is stuff that should already be hard coded into the topology files? I guess I could think of a world where someone might want to expose the bond types as a user setting but I'm not sure for what purpose.

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+class VdWSettings(SettingsBaseModel):
+    """
+    Settings for van der Waals force
+    """
+
+    combining_rules: Literal["Lorentz-Berthelot"]
+    potential: Literal["Lennard-Jones-12-6"]
+    scale: vdWScale
+    long_range_dispersion: Literal["isotropic"]
+    cutoff: float  # angstrom
```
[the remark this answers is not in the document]
mikemhenry: Sorry I was less fancy in this comment than I was in others, here we would require a unit with dimension of length. We can then convert it to whatever unit the tool needs when we pass it in.

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+    solvent_dielectric: Union[float, None]
+
+
+class BondSettings(SettingsBaseModel):
+    """
+    Settings for harmonic bonds
+    """
+
+    # U(r) = (k/2)*(r-length)^2
+    potential: Literal["harmonic"]
```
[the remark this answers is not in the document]
mikemhenry: Do we ever have to worry about which functional forms we support? If someone gives us a FF file where there isn't a 1/2, will that matter? Most FF files don't annotate their functional form, so it seems like something we should get.

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+    potential: Literal["k*(1+cos(periodicity*theta-phase))"]
+    default_idivf: Literal["auto"]
+
+
+class GBSASettings(SettingsBaseModel):
+    """ "
+    Settings for Generalized-Born surface area (GBSA) implicit solvent parameters
+    """
+
+    gb_model: Literal["HCT", "OBC1", "OBC2"]
```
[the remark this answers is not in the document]
mikemhenry: There are! Which ones do we want to support?

mikemhenry:
> just leaving some initial comments - TLDR I think I've completely missed the point of what you're trying to achieve here. My understanding is that these should be simulation control settings but a lot of this is stuff that should already be hard coded into the topology files? I guess I could think of a world where someone might want to expose the bond types as a user setting but I'm not sure for what purpose.

I believe that as it stands, these settings may be over descriptive but they are as over descriptive as they can be according to the SMIRNOFF spec. I wanted to be as strict as possible on the first pass, so then we don't miss anything.

gufe/settings/models.py
mikemhenry: It matters in how rdkit will interpret things, so if the FF was parametrized with that assumption it could matter. This is less important if we don't care about fully supporting openeye https://docs.eyesopen.com/toolkits/python/oechemtk/aromaticity.html#aromaticity-models-in-oechem-tk

gufe/settings/models.py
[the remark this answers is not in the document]
mikemhenry: We really only care about the dimension of the quantity users us give us since we can always convert. We could be more strcit but honestly I don't care if they pass us `9 * unit.angstrom` or `0.9 * unit.nm`

gufe/settings/models.py
richardjgowers: My thought was that if we're providing a fully interpreted `openff.toolkit.Molecule` is there any more interpretation of the aromaticity happening?  I'll ask over on the ff slack

gufe/settings/models.py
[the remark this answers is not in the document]
richardjgowers: Yeah the nice thing about units is you can use any unit that is in the correct dimension(s)

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+
+    class Config:
+        extra = Extra.forbid
+        # Immutability in python is never strict.
+        # If developers are determined/stupid
+        # they can always modify a so-called "immutable" object.
+        # https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability
+        allow_mutation = False
+        arbitrary_types_allowed = True
+        smart_union = True
```
richardjgowers, line 25: I'm not sure we have to worry about settings being mutable.  I think we'll store this as a "private" variable, then only return copies of it when it is accessed.

gufe/settings/models.py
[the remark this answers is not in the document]
richardjgowers: Hmm, it is interesting that we can get pretty close to matching FFs to supporting engines if we have this info on each end.

@IAlibay I think we're trying to make the supplied FF be completely unambiguous when we later use it in an engine.

gufe/settings/models.py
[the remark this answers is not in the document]
richardjgowers: I think it might be safer to accept `str` here rather than try and capture all possible (and future) models.

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+    aromaticity_model: Literal["OEAroModel_MDL"]
+
+
+class ThermoSettings(SettingsBaseModel):
+    """
+    Settings for thermodynamic parameters
+    """
+
+    temperature: PositiveFloat  # Θ
+    pressure: PositiveFloat  # M L**−1 T**−2
```
richardjgowers, line 153: Maybe this is what you're making with Matt in openff-pydantic, but ideally these'll be in temperature/pressure units (which validate that they're positive)

gufe/settings/models.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    # Tricky since this allows functions https://openforcefield.github.io/standards/standards/smirnoff/#electrostatics
+    periodic_potential: Union[Literal["Ewald3D-ConductingBoundary"], str]
+    nonperiodic_potential: Union[Literal["Coulomb"], str]
+    exception_potential: Union[Literal["Coulomb"], str]
+    scale: ElectrostaticScale
+    cutoff: Union[float, None]
+    switch_width: Union[float, None]
+    solvent_dielectric: Union[float, None]
```
richardjgowers, line 68: `Union[float, None]` is better stated as `Optional[float]`

gufe/tests/data/note.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<note>
+  <to>Richard</to>
+  <from>Irfan</from>
+  <heading>Reminder</heading>
+  <body>I'm taking holiday for the Queen!</body>
+</note>
```
richardjgowers, line 7: platty jubes ftw

gufe/settings/models.py
richardjgowers: Ok from asking around, this isn't actually used, so we can remove it

gufe/settings/models.py
richardjgowers: Also temperature and pressure are also optional (NVE/NVT)

gufe/settings/models.py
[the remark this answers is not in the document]
IAlibay: This should be encoded into the topology though, there's no user settings for any MD engine where this is actually used.

I get the intent, but I can't see where this is actually going to ever be used?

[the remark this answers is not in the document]
RiesBen: Gromacs has the bond form change between harmonic or morse potential :) 
https://manual.gromacs.org/2018.3/user-guide/mdp-options.html

gufe/settings/models.py
RiesBen: might also depend on the integrator of choice (MMC, MD, Langevin are all different here)

gufe/settings/settings.py
```diff
@@ -0,0 +1,141 @@
...
+                     ProperTorsionSettings, Settings, ThermoSettings,
+                     VdWSettings)
+
+
+def ff_settings_from_offxml(force_field) -> ForcefieldSettings:
+    """
+    Creates a ForcefieldSettings object from an open forcefield ForceField object.
+    """
+    vdW = VdWSettings(
+        cutoff=force_field.get_parameter_handler("vdW").cutoff.magnitude,
```
RiesBen, line 23: would it make sense to store force_field.get_parameter_handler("vdW") in a variable?

gufe/settings/models.py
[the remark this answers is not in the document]
IAlibay: @RiesBen the best practices here is not to use that option but define it in the ITP. Spuriously overriding bonds using that option is just a terrible idea in next to all cases.

RiesBen:
adding to my comments:
I like the implementation of the settings. (this is based on yesterday afternoon ;) I don't know what the discussions brought up. )

dwhswenson (dismissed):
My understanding is that the force field stuff is in flux right now (largely to be removed for now) so I didn't dive into reading that. Personally, I'm in favor of reducing the completeness of that in favor of something simpler (at least for now).

`ThermoSettings` is a good start, but there's a lot of validation that it should do to prevent users from requesting unreasonable thermodynamic states (namely, things that might be theoretically possible in stat mech, but aren't likely to be implemented in any MD engine).

I notice that it looks like a lot of the tests are just smoke tests -- I'm assuming this isn't the final state?

richardjgowers:
@dwhswenson wrt validating ensembles, I'd rather it was mostly that Protocols spat out `Setting`s with ensembles they can't do (e.g. if I pass in uVT to something without gcmc) rather than we try and police things in the Settings.  So disallowing NVE because we think it's not (currently) possible seems heavy handed.  But disallowing NPE because it doesn't (ever) make sense seems ok, so I think it's a little grey where we draw the line with validation.  Maybe scientific validity could be the line we draw, rather than algorithmic suitability.

gufe/settings/models.py
IAlibay: I agree with @richardjgowers that this needs to be at the protocol level.

Off the top of my head I can't think of a single case of a microcanonical calculation being used as part of a free energy calculation, but someone out there might be doing some real cool with stat mech that my pharmacist brain can't conceptualize.

I'd even venture further than what Richard is saying and propose that maybe ensemble checking shouldn't even happen here, even if someone says they'll do something silly like NPE, then the protocol should define the range of ensembles it'll take and everything else just gets told "no"?

W.r.t. integrators - yeah we'll need to accept the full range of integrators in the relevant setting for it, for non-alchemical perturbations you might want to rely on someone like gromac's `md` integrator to speed things up (e.g. in an MMPBSA style calculation where you just need a vanilla MD sim).

gufe/settings/models.py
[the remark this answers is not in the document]
mikemhenry: Removing "covalent" parts of the forcefield model settings

gufe/settings/models.py
mikemhenry: I don't think we need to go to extra effort, but I don't want to have to worry about the validation code having to re-check things, so I think I would prefer for this to act immutable.

gufe/settings/models.py
mikemhenry: If we are not including the integrator or ensemble, the best we can do is validate the inputs are the correct dimension + sign (ie no negative temperature).

gufe/settings/models.py
mikemhenry: These will all change once I add in units, but that is good to know!

gufe/settings/settings.py
mikemhenry: I'm not sure why, once I get units in this line will be:
`cutoff=force_field.get_parameter_handler("vdW").cutoff,`

gufe/tests/conftest.py
```diff
@@ -46,6 +46,23 @@ def PDBx_181L_path():
...
+@pytest.fixture
+def all_settings_path():
+    with importlib.resources.path('gufe.tests.data', 'all_settings.json') as f:
+        yield str(f)
+
+
+@pytest.fixture
+def dummy_xml_path():
+    with importlib.resources.path('gufe.tests.data', 'note.xml') as f:
+        yield str(f)
```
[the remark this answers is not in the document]
mikemhenry: Ooops, this was in here since the openfe forcefield parser throws a different warning if the xml isn't valid, verses the openffxml you passed in isn't valid, so I needed a valid xml file to test that difference in behavior. I then decided it was easier to just catch either error type and forgot to remove this. 

Do we want to handle the error "not valid xml" and "not a valid openffxml" differently?

mikemhenry:
> I notice that it looks like a lot of the tests are just smoke tests -- I'm assuming this isn't the final state?

Yes I didn't want to invest to much time into testing since I knew things would change, but the current tests IMHO are good, they test for round trip + make sure we can generate a schema. These things are easy to break once you start passing in objects that have units attached.

@richardjgowers 
Are going to do any ensemble validation? Like if a user says NVT, we make sure we have T? What ensemble are we going to support? 

Or do we want to just validate what they pass in is physical (dimension, magnitude, and sign wise)?

mikemhenry:
Hmm, while fixing the tests (the JSON file I have is missing units) I noticed the python API lets you bypass the unit check https://github.com/mattwthompson/openff-models/issues/8

gufe/tests/data/note.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<note>
+  <to>Richard</to>
+  <from>Irfan</from>
+  <heading>Reminder</heading>
+  <body>I'm taking holiday for the Queen!</body>
```
mikemhenry, line 6: RIP

mikemhenry:
This is ready for another review, do we need to support the old and new toolkit?

dwhswenson:
I'd vote for dropping old tk. @richardjgowers, any reason we need to keep supporting it? I thought that was a stopgap until the new one had dropped.

richardjgowers:
We need to run a quick back to back to check nothing broke, so this PR can
probably just be new tk.

gufe/settings/models.py
```diff
@@ -0,0 +1,118 @@
...
+    scale14: float
+    scale15: float
+
+
+class VdWSettings(SettingsBaseModel):
+    """
+    Settings for van der Waals force
+    """
+
+    combining_rules: Literal["Lorentz-Berthelot"]
```
dotsdl, line 49: Can we add `pydantic` `Field` usage to each of these attributes, with a one-liner for what each one is?

gufe/settings/models.py
```diff
@@ -0,0 +1,118 @@
...
+class ThermoSettings(SettingsBaseModel):
+    """
+    Settings for thermodynamic parameters
+    """
+
+    temperature: FloatQuantity["kelvin"]
+    pressure: FloatQuantity["standard_atmosphere"]
+    ph: Union[PositiveFloat, None]
+    redox_potential: Optional[float]
+
```
dotsdl, line 107: I don't see sampling settings, such as simulation timestep; where should this go?

dotsdl (commented):
Some things to address first; looking good @mikemhenry!

dotsdl:
I think you can go ahead and chop off the CI for `old-offtk`. I don't think it is in our interest to constrain ourselves to the pre-protein-support version of `openff-toolkit`.

gufe/settings/models.py
mikemhenry: Good point, we should make a meta issue from our setting taxonomy document to show which PR covers which bit

gufe/settings/models.py
```diff
@@ -0,0 +1,118 @@
...
+class ThermoSettings(SettingsBaseModel):
+    """
+    Settings for thermodynamic parameters
+    """
+
+    temperature: FloatQuantity["kelvin"]
+    pressure: FloatQuantity["standard_atmosphere"]
+    ph: Union[PositiveFloat, None]
+    redox_potential: Optional[float]
```
IAlibay, line 106: do we want to include chemical potential?

gufe/settings/models.py
```diff
@@ -0,0 +1,118 @@
...
+    gbsa: GBSASettings
+
+
+class ThermoSettings(SettingsBaseModel):
+    """
+    Settings for thermodynamic parameters
+    """
+
+    temperature: FloatQuantity["kelvin"]
+    pressure: FloatQuantity["standard_atmosphere"]
```
IAlibay, line 104: volume?

gufe/settings/models.py
IAlibay: Energy too

gufe/settings/models.py
```diff
@@ -0,0 +1,132 @@
...
```
IAlibay, line 105: A musing here - where do "things related to pressure that are important in defining the behaviour of a model" go? Like surface tension and isotropy.

environment.yml
```diff
@@ -4,17 +4,19 @@ channels:
   - conda-forge
   - openeye
 dependencies:
-  - numpy
+  - cachetools
```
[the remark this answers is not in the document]
mattwthompson: This was necessary when installing a pre-release version of 0.11.0, it should no longer be needed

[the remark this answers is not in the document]
mikemhenry: this was from when the new openff toolkit was installed via pip, so now it gets pulled in with the conda install

gufe/settings/models.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class VdWSettings(SettingsBaseModel):
+    """Settings for van der Waals force.
+
+    See the `SMIRNOFF vdW specification <https://openforcefield.github.io/standards/standards/smirnoff/#vdw>`_
+    for more details.
+    """
+
+    combining_rules: Literal["Lorentz-Berthelot"]
```
IAlibay, line 59: I realise populating all these might take time / effort, I'd be happy to see these being raised as issues for folks to pitch in a short amount of time on (rather than holding back the PR).

IAlibay (approved):
Assuming this is the start of more work into the settings (i.e. a discussion what protocols should provide and what can be left to settings, including the meta issue as detailed here: https://github.com/OpenFreeEnergy/gufe/pull/37/files#r995925175). I'd be happy with this PR being merged as-is.

gufe/settings/models.py
```diff
@@ -0,0 +1,143 @@
...
+
+class ThermoSettings(SettingsBaseModel):
+    """Settings for thermodynamic parameters.
+
+    No checking is done to ensure a valid thermodynamic ensemble is possible.
+    """
+
+    temperature: FloatQuantity["kelvin"] = Field(None, description="Simulation temperature, default units kelvin")
+    pressure: FloatQuantity["standard_atmosphere"] = Field(None, description="Simulation pressure, default units standard atmosphere (atm)")
+    volume: FloatQuantity["nm**3"] = Field(None, description="Simulation volume, default units are cubic nanometres")
```
richardjgowers, line 120: I'm not sure volume is right here.
1) Having volume as nm^3 is correct, but really you'd want the box vectors which are constant
2) The box vectors are more part of the chemical composition (ChemicalSystem half of things) rather than the Protocol definition.

Maybe constant V is actually denoted by Pressure not being set.

gufe/settings/models.py
```diff
@@ -0,0 +1,143 @@
...
+    ph: Union[PositiveFloat, None] = Field(None, description="Simulation pH, only used if MD engine supports constant pH")
```
richardjgowers, line 121: I think stating things are only used if supported applies to literally every field here, just drop it.

mikemhenry:
constant pH
https://pubs.acs.org/doi/10.1021/acs.jctc.2c00516

gufe/settings/models.py
mikemhenry: Cool we can drop it and let the protocol handle the logic for that

gufe/settings/models.py
mikemhenry: That sounds good, but can you give an example of what you would like to see here? Link to the openff doc? Wikipedia page? Or just a blurb: this mixing rule means (then the math expression)

gufe/settings/models.py
IAlibay: I'm not sure I agree here, this needs more discussion. 1) I don't think we can assume folks might not use these settings outside of gufe components, 2) it makes no sense to define these ensemble parameters without defining volume. You either go make the call that they aren't ensemble parameters and have incomplete ones, or you have everything.

dwhswenson: Maybe it's a little too much thermodynamics, but I like the idea that `pressure=None` implies constant volume, `temperature=None` implies constant energy, etc. Value of volume is determined by the unit cell, no? And usually that's what users have actually set -- do we require them to calculate volume and set it here as well?

I think the biggest risk is that someone might accidentally forget to set pressure or such when they meant to, leading to using the wrong ensemble. I'm not sure if we want to catch that, but the easy way to do it would be to make the default value be a custom sentinel object, so you have to explicitly set `None` if that's what you mean.

IAlibay: They may or they may not set it. But you could envision someone using these settings as part of a workflow. It depends on if you want these settings to be settings that can be independent (in which case you need to cover all thermodynamic parameters) or that has to exist within a predefined conception of how the settings will be applied (in which case you can skimp out on certain things).

mikemhenry: What if for now we just kept `temperature` and `pressure`, keep the default as `None` and see how things go after an iteration on using this in a protocol?

mikemhenry:
I should also mention that the docs work, but will need some tweaking because of how limited horizontal space is in our theme 
image (2)


This is the package I used to generate the docs, so you can see how it is supposed to look
https://sphinx-pydantic.readthedocs.io/en/latest/

mikemhenry:
I also tried to get the docs to build on the PR, but it looks like the webook needs to be re-synced and I don't have the permissions to do that
image (3)

IAlibay:
> I also tried to get the docs to build on the PR, but it looks like the webook needs to be re-synced and I don't have the permissions to do that image (3)

We need to get this done for both gufe and openfe!

mikemhenry:
As is tradition with expanding PR scope, this PR also sets up RTD docs to use our env file (with mamba)

mikemhenry:
looks like it works https://gufe--37.org.readthedocs.build/en/37/api/settings.html

IAlibay (approved):
I won't block.

PR #38, opened 2022-07-25
WIP: Add settings add chem comp

RiesBen:
Hi, 
here i try to add an initial draft of the chemComposition.
I also slightly modified the file structure, but this can be reverted if wished?

This is currently WIP ;)


gufe/settings/models/chemical_composition.py
```diff
@@ -0,0 +1,52 @@
+"""
+Pydantic models used for storing settings for chemical Composition.
+"""
+
+from .abstract import SettingsBaseModel
+from typing import List, Optional
+
+# abstract entity
+class Chemical_Entity(SettingsBaseModel):
```
dotsdl, line 9: No underscore in class name.

gufe/settings/models/chemical_composition.py
```diff
@@ -0,0 +1,52 @@
...
+    exists: bool = False
+
+
+# Solutes:
+class Solute(Chemical_Entity):
+    num_compounds: int = 0
+
+
+# Protein:
+class Protein_Cofactors(Chemical_Entity):
```
dotsdl, line 19: No underscore in class name.

dotsdl (commented):
Thanks @RiesBen! I did a quick review on mostly style things. Can you tell me what these objects are supposed to do? It seems odd to me that we are including system composition as part of settings, but I don't know if I missed out on some discussions here.

richardjgowers (changes requested):
I think this is off on a tangent a little.  `Settings` are designed to capture everything *except* this sort of information.  I think *all* of the information here will be collected in various `*Component` classes

RiesBen:
@richardjgowers @dotsdl 
Hm I see your point, I tried to add this layer as it was defined in the taxonomy doc. 
For a simulation setup I think, it might be beneficial to have a fast look up on this, to support the reduced potential settings or other layers. But I also see, that this is represented by the Component classes. 
Therefore one could also think about removing it again? Then we should also remove it from the doc, or did I miss understand the doc?

IAlibay:
> Then we should also remove it from the doc, or did I miss understand the doc?

My 2 cent is that I wouldn't remove them from the doc, the components are a "setting" in themselves, i.e. they define how the positions and chemistry of your system are "set".

PR #39, opened 2022-07-26
Serialization and tokenization via GufeTokenizableMixin

dotsdl:
This PR is an iteration on #30. 

In this approach, we implement tokenization as performed in `dask.base`, and put as much of the required machinery in `gufe.base.GufeTokenizableMixin`.

Every tokenizable `gufe` object should include this mixin.
When included, this provides the `.key` property, which is the stable identifier across platforms and Python sessions.
This is composed of `<classname>-<token>`, where `.token` is an MD5-hash of the non-default values included in the
`.to_dict(include_defaults=False, keyencode_dependencies=True)` form of the object.

We also include:
- a registry for `Component` subclasses (`gufe.component.COMPONENT_REGISTRY`); subclasses are automatically added on definition
- a registry for existing objects in memory, by key (`gufe.base.OBJECT_REGISTRY`)

### DIFF
--- a/gufe/base.py
+++ b/gufe/base.py
@@ -1,0 +1,108 @@
+import abc
+import uuid
+import sys
+import threading
+import hashlib
+import inspect
+from packaging.version import parse as parse_version
+from typing import Dict, Any, Callable
+
+
+class GufeTokenizableMixin:
+    """Mixin for all tokenizble tokenizeable gufe objects.
+
+    """
+
+    @property
+    def token(self):
+        if not hasattr(self, '_token') or self._token is None:
+            self._token = tokenize(self)
+        return self._token
+
+    @property
+    def key(self):
+        if not hasattr(self, '_key') or self._key is None:
+            prefix = type(self).__name__
+            self._key = GufeKey(f"{prefix}-{self.token}")
+        return self._key
+
+    @property
+    def defaults(self):
+        sig = inspect.signature(self.__init__)
+
+        defaults = {
+            param.name: param.default for param in sig.parameters.values()
+            if param.default is not inspect.Parameter.empty
+        }
+
+        return defaults
+
+    def _keyencode_dependencies(self, d):
+        for key, value in d.items():
+            if isinstance(value, dict):
+                self._keyencode_dependencies(value)
+            elif isinstance(value, list):
+                for i, item in enumerate(value):
+                    if hasattr(item, '_gufe_tokenize'):
+                        d[key][i] = item
+            else:
+                if hasattr(value, '_gufe_tokenize'):
+                    d[key] = value.token
+
+        return d
+
+
+class GufeKey(str):
+    def __repr__(self):
+        return f"<GufeKey('{str(self)}')>"
+
+
+STUBBED_OBJECT_REGISTRY: Dict[str, GufeTokenizableMixin] = {}
+"""Registry of token-stubbed objects.
+
+Used to avoid duplication of tokenizable `gufe` objects in memory when deserialized.
+Each key is a token, each value the token-stubbed version of the object
+corresponding to that token.
+
+"""
+
+#TODO see if you can proceed without this first
+#STUBBED_OBJECT_REGISTRY_REVERSE: Dict[GufeTokenizableMixin, str] = {}
+#"""Reverse registry of token-stubbed objects.
+#
+#Used to Avoid duplication of tokenizable `gufe` objects in memory when deserialized.
+#Each key is a token-stubbed version of a `gufe` object, and each value is a 
+#token corresponding to that object.
+#
+#"""
+
+
+# FROM dask.base
+
+# Pass `usedforsecurity=False` for Python 3.9+ to support FIPS builds of Python
+_PY_VERSION = parse_version(".".join(map(str, sys.version_info[:3])))
+_md5: Callable
+if _PY_VERSION >= parse_version("3.9"):
+
+    def _md5(x, _hashlib_md5=hashlib.md5):
+        return _hashlib_md5(x, usedforsecurity=False)
+
+else:
+    _md5 = hashlib.md5
+
+NORMALIZERS = {}
+
+def normalize_object(o):
+    method = getattr(o, "_gufe_tokenize", None)
+    if method is not None:
+        return method()
+    else:
+        raise ValueError("Cannot normalize without `_gufe_tokenize` method.")
+
+
+def tokenize(arg):
+    """Deterministic token"""
+    hasher = _md5(str(normalize_object(arg)).encode())
+    return hasher.hexdigest()
+
+
--- a/gufe/chemicalsystem.py
+++ b/gufe/chemicalsystem.py
@@ -7,10 +7,11 @@ from typing import Dict, Optional
 import numpy as np
 from openff.toolkit.utils.serialization import Serializable
 
-from .component import Component
+from .base import GufeTokenizableMixin, normalize_object
+from .component import Component, from_dict as component_from_dict
 
 
-class ChemicalSystem(Serializable, abc.Mapping):
+class ChemicalSystem(abc.Mapping, GufeTokenizableMixin):
     """A node of an alchemical network.
 
     Attributes
@@ -93,8 +94,9 @@ class ChemicalSystem(Serializable, abc.Mapping):
 
     # TODO: broken without a `Component` registry of some kind
     # should then be changed to use the registry
-    def to_dict(self):
+    def _to_dict(self):
         return {
+            "__class__": self.__class__.__name__,
             "components": {
                 key: value.to_dict() for key, value in self.components.items()
             },
@@ -105,7 +107,7 @@ class ChemicalSystem(Serializable, abc.Mapping):
     # TODO: broken without a `Component` registry of some kind
     # should then be changed to use the registry
     @classmethod
-    def from_dict(cls, d):
+    def _from_dict(cls, d):
         return cls(
             components={
                 key: Component.from_dict(value) for key, value in d["components"]
--- a/gufe/component.py
+++ b/gufe/component.py
@@ -4,37 +4,103 @@
 import abc
 from typing import Union
 
+from .base import GufeTokenizableMixin
 
-class Component(abc.ABC):
+
+COMPONENT_REGISTRY = {}
+
+def register_component(cls):
+    COMPONENT_REGISTRY[cls.__name__] = cls
+
+    
+class _ComponentMeta(type):
+    def __new__(cls, name, bases, classdict):
+        componentcls = super().__new__(cls, name, bases, classdict)
+        register_component(componentcls)
+        return componentcls
+
+
+class _ABCComponentMeta(_ComponentMeta, abc.ABCMeta):
+    # required to make use of abc.ABC in classes that use _ComponentMeta metaclass
+    # see https://stackoverflow.com/questions/57349105/python-abc-inheritance-with-specified-metaclass 
+    ...
+        
+
+class Component(abc.ABC, GufeTokenizableMixin, metaclass=_ABCComponentMeta):
     """Base class for members of a ChemicalSystem"""
+
+    def __init__(self):
+
+        # since __hash__ isn't inheritable, monkey-patching is the best we can
+        # do besides defining it in every subclass
+        self.__class__.__hash__ = Component.__hash__
 
     def __repr__(self):
         return f"{self.__class__.__name__}(name={self.name})"
 
-    @abc.abstractmethod
-    def __hash__(self):
-        pass
-
     def __lt__(self, other):
         return hash(self) < hash(other)
 
-    @abc.abstractmethod
     def __eq__(self, other):
-        pass
+        if not isinstance(other, self.__class__):
+            return NotImplemented("Component comparisons must be of the same type")
 
-    @abc.abstractmethod
-    def to_dict(self) -> dict:
-        pass
+        d = self.to_dict()
+        do = other.to_dict()
+
+        if set(d.keys()) == set(do.keys()):
+            for key in d:
+                if d[key] != do[key]:
+                    return False
+
+            return True
+
+    def __hash__(self):
+        return hash(self.key)
+
+    def _gufe_tokenize(self):
+        """Return a list of normalized inputs for `gufe.base.tokenize`.
+
+        """
+        return sorted(self.to_dict(include_defaults=False).items(), key=str)
 
     @property
     @abc.abstractmethod
     def name(self) -> str:
         pass
 
-    @classmethod
+    def to_dict(self, include_defaults=True, keyencode_dependencies=False) -> dict:
+        """Serialize to dict representation"""
+
+        d = self._to_dict()
+        d['__class__'] = self.__class__.__name__
+
+        if keyencode_dependencies:
+            d = self._keyencode_dependencies(d)
+
+        if not include_defaults:
+            for key, value in self.defaults.items():
+                if d.get(key) == value:
+                    d.pop(key)
+
+        return d
+
     @abc.abstractmethod
-    def from_dict(cls, d: dict):
-        pass
+    def _to_dict(self) -> dict:
+        ...
+
+
+    def from_dict(cls, d: dict, keydecode_dependencies=False):
+        """Deserialize from dict representation"""
+        component_type = d.pop('__class__', None)
+        if component_type is None:
+            raise KeyError("No `__class__` key found; unable to deserialize Component")
+
+        return COMPONENT_REGISTRY[component_type]._from_dict(d)
+
+    @abc.abstractmethod
+    def _from_dict(cls, d: dict):
+        ...
 
     @property
     @abc.abstractmethod
--- a/gufe/smallmoleculecomponent.py
+++ b/gufe/smallmoleculecomponent.py
@@ -5,7 +5,6 @@ import logging
 logger = logging.getLogger('openff.toolkit')
 logger.setLevel(logging.ERROR)
 from openff.toolkit.topology import Molecule as OFFMolecule
-from openff.toolkit.utils.serialization import Serializable
 import warnings
 
 from rdkit import Chem
@@ -46,7 +45,7 @@ def _ensure_ofe_version(mol: RDKitMol):
     mol.SetProp("ofe-version", __version__)
 
 
-class SmallMoleculeComponent(Component, Serializable):
+class SmallMoleculeComponent(Component):
     """A molecule wrapper suitable for small molecules
 
     .. note::
--- a/gufe/solventcomponent.py
+++ b/gufe/solventcomponent.py
@@ -114,20 +114,29 @@ class SolventComponent(Component):
         """Solvents don't have a formal charge defined so this returns None"""
         return None
 
-    def __eq__(self, other):
-        if not isinstance(other, SolventComponent):
-            return NotImplemented
-        return (self.smiles == other.smiles and
-                self.positive_ion == other.positive_ion and
-                self.negative_ion == other.negative_ion and
-                self.ion_concentration == other.ion_concentration)
+    #def __eq__(self, other):
+    #    if not isinstance(other, SolventComponent):
+    #        return NotImplemented
+    #    return (self.smiles == other.smiles and
+    #            self.positive_ion == other.positive_ion and
+    #            self.negative_ion == other.negative_ion and
+    #            self.ion_concentration == other.ion_concentration)
 
-    def __hash__(self):
-        return hash((self.smiles, self.positive_ion, self.negative_ion,
-                     self.ion_concentration))
+    def _gufe_tokenize(self):
+        """Return a list of normalized inputs for `gufe.base.tokenize`.
+
+        """
+
+        return sorted(self.to_dict(include_defaults=False).items(), key=str)
+
+        #return [self.smiles,
+        #        self.positive_ion,
+        #        self.negative_ion,
+        #        self.ion_concentration.to_tuple(),
+        #        self.neutralize]
 
     @classmethod
-    def from_dict(cls, d):
+    def _from_dict(cls, d):
         """Deserialize from dict representation"""
         ion_conc = d['ion_concentration']
         if ion_conc:
@@ -135,7 +144,7 @@ class SolventComponent(Component):
 
         return cls(**d)
 
-    def to_dict(self):
+    def _to_dict(self):
         """For serialization"""
         ion_conc = self.ion_concentration
         if ion_conc:


gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+    _md5 = hashlib.md5
+
+NORMALIZERS = {}
+
+def normalize_object(o):
+    method = getattr(o, "_gufe_tokenize", None)
+    if method is not None:
+        return method()
+    else:
+        raise ValueError("Cannot normalize without `_gufe_tokenize` method.")
```
richardjgowers, line 100: eafp

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def tokenize(arg):
+    """Deterministic token"""
```
richardjgowers, line 104: type annotation and docs, tokenization could mean 100 things, be helpful

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+            else:
+                if hasattr(value, '_gufe_tokenize'):
+                    d[key] = value.token
+
+        return d
+
+
+class GufeKey(str):
+    def __repr__(self):
+        return f"<GufeKey('{str(self)}')>"
```
richardjgowers, line 57: why isn't this just a string?

gufe/component.py
```diff
@@ -4,37 +4,103 @@
...
-        pass
+        d = self.to_dict()
+        do = other.to_dict()
+
+        if set(d.keys()) == set(do.keys()):
+            for key in d:
+                if d[key] != do[key]:
+                    return False
+
+            return True
```
richardjgowers, line 56: if keys don't match this falls through, does that default to False?  just do `return d == do`?

gufe/chemicalsystem.py
```diff
@@ -7,10 +7,11 @@
 import numpy as np
 from openff.toolkit.utils.serialization import Serializable
 
-from .component import Component
+from .base import GufeTokenizableMixin, normalize_object
+from .component import Component, from_dict as component_from_dict
```
richardjgowers, line 11: `component_from_dict` doesn't exist?

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @property
+    def token(self):
+        if not hasattr(self, '_token') or self._token is None:
+            self._token = tokenize(self)
+        return self._token
+
+    @property
+    def key(self):
+        if not hasattr(self, '_key') or self._key is None:
```
richardjgowers, line 24: I don't see any usage of `_key`?  Why not just have one way of doing this

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class GufeTokenizableMixin:
+    """Mixin for all tokenizble tokenizeable gufe objects.
+
+    """
+
+    @property
+    def token(self):
+        if not hasattr(self, '_token') or self._token is None:
```
richardjgowers, line 18: is this to cache the token?

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+            param.name: param.default for param in sig.parameters.values()
+            if param.default is not inspect.Parameter.empty
+        }
+
+        return defaults
+
+    def _keyencode_dependencies(self, d):
+        for key, value in d.items():
+            if isinstance(value, dict):
+                self._keyencode_dependencies(value)
```
richardjgowers, line 43: this branch doesn't seem to catch the return value?

gufe/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return d
+
+
+class GufeKey(str):
+    def __repr__(self):
+        return f"<GufeKey('{str(self)}')>"
+
+
+STUBBED_OBJECT_REGISTRY: Dict[str, GufeTokenizableMixin] = {}
```
richardjgowers, line 60: this isn't used?

gufe/base.py
dotsdl: Yes!

gufe/base.py
dotsdl: We might just want to drop the `token` property, since we use `key` for stubbing (keyencode/decode).

gufe/base.py
dotsdl: Ah, good catch. Yeah this and the two that follow need to be reworked to look like `dictdecode_dependencies`. They're janky currently.

gufe/base.py
dotsdl: It is with a `str(...)`, and otherwise is one. This just makes it clear it's not *just* a string.

gufe/base.py
dotsdl: Agree, can add these tomorrow.

gufe/base.py
dotsdl: Fair enough; can change this. Bit of a vestige of pulling normalizers out of `dask.base`.

gufe/base.py
dotsdl: It is now, and has been renamed. Caught me midstream. :grin:

gufe/component.py
dotsdl: Ah good point. Fixed!

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+    def defaults(self):
+        sig = inspect.signature(self.__init__)
+
+        defaults = {
+            param.name: param.default for param in sig.parameters.values()
+            if param.default is not inspect.Parameter.empty
+        }
+
+        return defaults
+
```
dwhswenson, line 80: consider making this abstract to avoid subtle errors by users

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+        """Return a list of normalized inputs for `gufe.base.tokenize`.
+
+        """
+        return sorted(self.to_dict(include_defaults=False).items(), key=str)
+
+    def to_dict(self, include_defaults=True, keyencode_dependencies=False) -> dict:
+        """Serialize to dict representation"""
+
+        d = self._to_dict()
+        d['__class__'] = self.__class__.__name__
```
dwhswenson, line 150: add `__module__`

dwhswenson:
Storage will need the following (all currently available; commenting because @dotsdl wanted to be sure that these things remain available):

* a shallow `to_dict`/`from_dict`: This is currently implemented (it would be nice if the kwarg to get it was just `shallow`, it might be convenient for storage if the default was the shallow dict)
* a way to get a unique identifier for the object: `obj.token` does it currently
* a way to identify that a given object is "one of ours" -- this can be `isinstance(obj, GufeTokenizable)`

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+                cls._keydecode_dependencies(value)
+            elif isinstance(value, list):
+                for i, item in enumerate(value):
+                    if isinstance(item, str) and item in TOKENIZABLE_REGISTRY:
+                        d[key][i] = TOKENIZABLE_REGISTRY[item]
+            else:
+                if isinstance(value, str) and value in TOKENIZABLE_REGISTRY:
+                    d[key] = TOKENIZABLE_REGISTRY[value]
+
+        return d
```
dwhswenson, line 110: This seems to expect that objects that are being loaded are already in the registry. This will not be true when deserializing.

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+            param.name: param.default for param in sig.parameters.values()
+            if param.default is not inspect.Parameter.empty
+        }
+
+        return defaults
+
+    # TODO: rewrite the below three functions to be fully recursive like 
+    # _dictdecode_dependencies
+    def _keyencode_dependencies(self, d):
+        for key, value in d.items():
```
dwhswenson, line 84: All the encoders assume that dicts never have tokenizable objects as keys. Can we trust that this is a safe assumption?

gufe/base.py
dwhswenson: also, consider using `__qualname__`

dwhswenson (commented):
Repeated structure in encoding/decoding can be abstracted out. That can also make it easier for me to do other things (like generate a list of all gufe objects within a given object in order to serialize them all). Additionally, they can be removed from the base class. I'll give more detailed suggestions in that direction soon.

(Sorry to everyone that there will be multiple reviews here -- trying to get feedback to @dotsdl as he's working on it)

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+            if isinstance(value, dict):
+                self._keyencode_dependencies(value)
+            elif isinstance(value, list):
+                for i, item in enumerate(value):
+                    if hasattr(item, '_gufe_tokenize'):
+                        d[key][i] = item.key
+            else:
+                if hasattr(value, '_gufe_tokenize'):
+                    d[key] = value.key
```
dwhswenson, line 93: I think this instead needs to be a dict that includes the key, perhaps something as simple as `{":gufe-key:": value.key}`. When loading from storage, we need some way of recognizing that this is "one of ours". It won't already be in the registry, so it needs to be something intrinsic to the serialized format.

gufe/base.py
dwhswenson: Actually, we can require storage to populate `TOKENIZABLE_REGISTRY` with everything it contains when it opens. Then this would work. The assumption is that there are relatively few objects to be stored there -- although it could get into the thousands, so I'm a little concerned about performance on opening storage in that case (especially if each object is a separate call to a cloud service.)

The alternative is that we need to provide a function to load something from storage as part of the signature here. Even with that, it's a bit tricky.

gufe/base.py
dwhswenson: ah, got confused between token and key here. Key is fine; token is not

dwhswenson:
Are we placing responsibility on the contributors to make copies of container objects like `dict`s and `list`s? I'd much rather if we didn't, and instead handled the copying internally.

Current error:

```python
# test.py

from gufe.base import GufeTokenizable

class Leaf(GufeTokenizable):
    def __init__(self, a):
        self.a = a

    def _to_dict(self):
        return {"a": self.a}

    @classmethod
    def _from_dict(cls, dct):
        return cls(**dct)


class Container(GufeTokenizable):
    def __init__(self, obj, dct):
        self.obj = obj
        self.dct = dct

    def _to_dict(self):
        return {"obj": self.obj, "dct": self.dct}

    @classmethod
    def _from_dict(cls, dct):
        return cls(**dct)


def test_serialize_dict():
    leaf = Leaf("foo")
    expected = {"dct": leaf}
    container = Container(leaf, expected.copy())
    assert container.dct == expected  # works now
    container.key
    assert container.dct == expected  # fails now
```

```pytb
_________________________________ test_serialize_dict __________________________________

    def test_serialize_dict():
        leaf = Leaf("foo")
        expected = {"dct": leaf}
        container = Container(leaf, expected.copy())
        assert container.dct == expected  # works now
        container.key
>       assert container.dct == expected  # fails now

test.py:34:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <gufe.storage.test.Leaf object at 0x14d17d970>
other = {'__class__': 'Leaf', 'a': 'foo'}

    def __eq__(self, other):
        if not isinstance(other, self.__class__):
>           return NotImplemented("Component comparisons must be of the same type")
E           TypeError: 'NotImplementedType' object is not callable

../base.py:46: TypeError
```

This is fixed if `Container.to_dict()` makes a copy of the dict that contains a `Leaf`. Something in the process of computing the key is modifying the `container.dct` here, replacing the Leaf instance with its dict form.

gufe/base.py
dotsdl: Removed `token` property to reduce the number of items in play. We just use `key`.

gufe/base.py
dotsdl: Done! This forces objects to either implement their own or `super().defaults` for this implementation.

gufe/base.py
dotsdl: I think that's a safe assumption, yes. Our messaging for `_to_dict` should be that the resulting `dict` should be JSON serializable, so using tokenizable (and therefore hashable) objects as keys would run counter to this.

Will add this explicitly to docs in `GufeTokenizable`.

gufe/base.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # TODO: rewrite the below three functions to be fully recursive like 
+    # _dictdecode_dependencies
+    def _keyencode_dependencies(self, d):
+        for key, value in d.items():
+            if isinstance(value, dict):
+                self._keyencode_dependencies(value)
+            elif isinstance(value, list):
+                for i, item in enumerate(value):
+                    if hasattr(item, '_gufe_tokenize'):
```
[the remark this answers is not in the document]
dotsdl: Done!

gufe/base.py
dotsdl: Combined @dwhswenson's formulation and made encoders/decoders work recursively.

gufe/base.py
dotsdl: Done!

gufe/base.py
dotsdl: Do the methods we have now give a storage implementation what it needs to identify (recursively) the dependents and pull them out before fully instantiating the requested object?

dotsdl: That would still be "on-demand", instead of requiring everything to be in-memory all at once that has ever been stored.

gufe/base.py
dotsdl: Added a note to `_to_dict`'s documentation.

gufe/base.py
dotsdl: Agreed, we'll proceed with {':gufe-key:': str(<GufeKey>)} as you suggest! I don't see any issue with this choice of structure.

dotsdl:
> Are we placing responsibility on the contributors to make copies of container objects like `dict`s and `list`s? I'd much rather if we didn't, and instead handled the copying internally.
> 
> Current error:
> 
> ```python
> # test.py
> 
> from gufe.base import GufeTokenizable
> 
> class Leaf(GufeTokenizable):
>     def __init__(self, a):
>         self.a = a
> 
>     def _to_dict(self):
>         return {"a": self.a}
> 
>     @classmethod
>     def _from_dict(cls, dct):
>         return cls(**dct)
> 
> 
> class Container(GufeTokenizable):
>     def __init__(self, obj, dct):
>         self.obj = obj
>         self.dct = dct
> 
>     def _to_dict(self):
>         return {"obj": self.obj, "dct": self.dct}
> 
>     @classmethod
>     def _from_dict(cls, dct):
>         return cls(**dct)
> 
> 
> def test_serialize_dict():
>     leaf = Leaf("foo")
>     expected = {"dct": leaf}
>     container = Container(leaf, expected.copy())
>     assert container.dct == expected  # works now
>     container.key
>     assert container.dct == expected  # fails now
> ```
> 
> ```
> _________________________________ test_serialize_dict __________________________________
> 
>     def test_serialize_dict():
>         leaf = Leaf("foo")
>         expected = {"dct": leaf}
>         container = Container(leaf, expected.copy())
>         assert container.dct == expected  # works now
>         container.key
> >       assert container.dct == expected  # fails now
> 
> test.py:34:
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> 
> self = <gufe.storage.test.Leaf object at 0x14d17d970>
> other = {'__class__': 'Leaf', 'a': 'foo'}
> 
>     def __eq__(self, other):
>         if not isinstance(other, self.__class__):
> >           return NotImplemented("Component comparisons must be of the same type")
> E           TypeError: 'NotImplementedType' object is not callable
> 
> ../base.py:46: TypeError
> ```
> 
> This is fixed if `Container.to_dict()` makes a copy of the dict that contains a `Leaf`. Something in the process of computing the key is modifying the `container.dct` here, replacing the Leaf instance with its dict form.

I think we're good here! All encoders/decoders do a deep copy within `modify_dependencies` from the start, and this happens before any mods to the structure. Thank you for putting together the abstracted approach! I modified it a bit to allow full recursion as well.

dotsdl:
Now working on fixing broken tests, as well as busting out the test suite for `gufe.base`. Assistance welcome here, especially on the broken tests and what we should do for them.

richardjgowers:
I've marked a bunch of the `SmallMoleculeComponent` tests as xfail.  These would `__eq__` very loosely (only requiring matching SMILES) but we're making this stricter for Tokenizables in this PR.  I'm working on better serialise/deserialise for molecules (SMC and Protein) so I'll fix these tests in that PR.

dwhswenson:
@richardjgowers : the problem is that rdkit from SDF has conformers, from SMILES does not. Tests pass if we add `AllChem.Compute2DCoords()` in `__init__` (need to update template SDFs as well).

dwhswenson:
I pushed a couple commits to make `Transformation` and `AlchemicalNetwork` into `GufeTokenizable`s. There are some issues with the underlying `_to_dict`/`_from_dict` methods implemented for `Transformation`s, but there weren't tests of the yet dicts. I'll fix those in #42.

dwhswenson:
@dotsdl : I wanted to call your attention to https://github.com/OpenFreeEnergy/gufe/pull/39/commits/9f754f4ae9f234cca32aa1b1f5e934e3f1bb782d (a very small change with a very long commit message to explain it). I'd appreciate if you can double-check my logic that it isn't breaking anything. It seems to work on the objects I've tried it on, and it fixes the problem I was having, but it seems like a potentially risky change.

gufe/base.py
```diff
@@ -0,0 +1,381 @@
...
+        return f"<GufeKey('{str(self)}')>"
+
+    def to_dict(self):
+        return {':gufe-key:': str(self)}
+
+
+# TODO: may want to make this a weakref dict to avoid holding references here
+# why are we not using WeakValueDictionary?
+TOKENIZABLE_REGISTRY: Dict[str, weakref.ref[GufeTokenizable]] = {}
+# TOKENIZABLE_REGISTRY: Dict[str, GufeTokenizable] = WeakValueDictionary()
```
dwhswenson, line 189: As I mention in the comment; any reason not to use a `weakrefs.WeakValueDictionary`? Wouldn't that abstract out some of the explicit weakrefs stuff we're doing now?

gufe/base.py
```diff
@@ -0,0 +1,381 @@
...
+    >>> s = SolventComponent()
+    >>> tokenize(s)
+    'e6eef7519854d35a5ce6c84136b3684c'
+
+    >>> tokenize(s) == tokenize(SolventComponent.from_dict(s.to_dict()))
+    True
+
+    """
+    hasher = _md5(str(normalize_object(obj)).encode())
+    return hasher.hexdigest()
```
dwhswenson, line 381: As stripped down as we've gotten it, I don't think we need all the dask-like structure here. 
* `_md5` isn't really reused in multiple places, so doesn't need to be module-level. Plus, I thought we were dropping Python 3.8, which means that we don't even need the branching to select it.
* `normalize_object` just says "I'll give you the normalized dict if you give me a gufe object, otherwise error." Do we have a use case where that error is useful? Can't this be removed and just use `_gufe_tokenize` when needed?
* `tokenize` takes the normalized dict and makes a hex token out of it. Can this just be inlined into  the `key` property?

gufe/base.py
```diff
@@ -0,0 +1,381 @@
...
+    obj = _from_dict(dct)
+    try:
+        thing = TOKENIZABLE_REGISTRY[obj.key]()
+        # weakref will return None if the object was deleted
+        if thing is None:
+            return obj
+        else:
+            return thing
+    except KeyError:
+        return obj
```
dwhswenson, line 313: With the change I made in [ebb52e3](https://github.com/OpenFreeEnergy/gufe/pull/39/commits/ebb52e3bce1c7bdd4e753e794f6dda020bf6a33f), I think that this may now be impossible to reach: when your instantiate `obj`, it registers itself if something hasn't already taken the spot for its key, so either a previous `thing` has existed or `thing is obj`.

I think it is good defensive coding to leave it here, but I think we should mark this no-cover.

gufe/base.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+def _from_dict(dct: Dict) -> GufeTokenizable:
+    module = dct.pop('__module__')
+    qualname = dct.pop('__qualname__')
+    if (qualname is None) or (module is None):
+        raise KeyError("`__qualname__` or `__module__` key not found; unable "
```
dwhswenson, line 320: In principle, this should be impossible to reach. This code should only be called if something was tested positive for `is_gufe_dict`, which requires the presence of these keys.

Again, good defensive programming: I vote keep the code but mark it no-cover.

gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -228,6 +231,9 @@ def test_deserialize_roundtrip(self, toluene, phenol):
         for x, y in zip(pos1, pos2):
             assert (x == y).all()
 
+    # TODO: determine if we want to add our own serializers for e.g. JSON
+    # based on `to_dict`
```
dwhswenson, line 235: Can we make a decision on this in this PR? The xfail here is a signifant contributor to a drop in coverage (the fail happens early).

My vote is to not support a direct JSON serialization of these objects. I'm basing that on the opinion that any user who knows that they want a JSON dump will also be smart enough to figure out how to do `json.dumps(obj.to_dict())`. If we support it, it should be in the base class.

dwhswenson (commented):
I went through and added nearly all the tests we need. There are a few corners I didn't get; as mentioned in review comments, I think some of these may be things we exclude from coverage anyway.

Most significant change I made was to register the instance on instantiation. This does require that keys are truly immutable, but I think we were already expecting that. Moving that to the metaclass was the simplest way to fix the problem that we weren't getting the same-in-memory object back from `to_dict`.

gufe/base.py
dwhswenson: Second thought here: I think it makes sense to have this logic, but to move it lower in the stack. I think this logic should be in `import_qualname` and tested as part of that.

gufe/base.py
richardjgowers: I didn't know about that, so can switch

dotsdl: Switched to using `WeakValueDictionary`!

gufe/base.py
dotsdl: Agree, simplifying now. Can re-add complexity later if needed.

dotsdl: I do like keeping the `tokenize` standalone function, even if its use is internal, so have not inlined it into `GufeTokenizable.key`.

gufe/base.py
dotsdl: Agreed, thanks for doing this!

gufe/base.py
dotsdl: I'm fine with removing it; agree with your assessment that it shouldn't be present. If we observe that there are edge cases in which we get a `KeyError`, then we'll need to understand and deal with them.

dotsdl:
> @dotsdl : I wanted to call your attention to [9f754f4](https://github.com/OpenFreeEnergy/gufe/commit/9f754f4ae9f234cca32aa1b1f5e934e3f1bb782d) (a very small change with a very long commit message to explain it). I'd appreciate if you can double-check my logic that it isn't breaking anything. It seems to work on the objects I've tried it on, and it fixes the problem I was having, but it seems like a potentially risky change.

Thanks for this @dwhswenson! I *think* this makes sense, but will poke at it some more tomorrow before proceeding. Agree it's a delicate change.

gufe/tests/test_smallmoleculecomponent.py
dotsdl: I'm fine with leaving it here. I've added exclusion of the test modules to `.coveragerc`; this has reduced coverage further in this instance, but makes sense in my mind to do.

dotsdl:
I believe this is now ready to merge. @richardjgowers if you agree please do!

dotsdl:
Merging this will allow us to move on to #42 and #45.

richardjgowers (approved):

PR #40, opened 2022-07-27
Solvent: make concentration always a unit, even if 0.0M

richardjgowers:
I think @dwhswenson asked about this and we forgot to implement, it seems neater to always have this exist rather than `None` sometimes

### DIFF
--- a/gufe/components/solventcomponent.py
+++ b/gufe/components/solventcomponent.py
@@ -31,7 +31,7 @@ class SolventComponent(Component):
                  positive_ion: str = 'Na+',
                  negative_ion: str = 'Cl-',
                  neutralize: bool = True,
-                 ion_concentration: unit.Quantity = None):
+                 ion_concentration: unit.Quantity = 0.0 * unit.molar):
         """
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


gufe/solventcomponent.py
```diff
@@ -69,15 +69,15 @@ def __init__(self, *,  # force kwarg usage
...
-               not ion_concentration.is_compatible_with(unit.molar)):
-                raise ValueError(f"ion_concentration must be given in units of"
-                                 f" concentration, got {ion_concentration}")
-            # concentration requires both ions be given
-            if ion_concentration > 0:
-                if self._negative_ion is None or self._positive_ion is None:
-                    raise ValueError("Ions must be given for concentration")
+
+        if (not isinstance(ion_concentration, unit.Quantity)
+           or not ion_concentration.is_compatible_with(unit.molar)):
```
dwhswenson, line 74: just readability, but I think the `or` should be moved over one space to align *after* the paren.

Actually, for something like this, I'm starting to prefer the style black uses for long if statements (which makes a clearer separation of the block following the if):

```python
if (
    thing
    or thing2
):
    do_something()
```

dwhswenson (approved):
LGTM. One suggestion on style/readability. But the substance of the PR is good.

PR #41, opened 2022-07-27
Reorganized component modules within a submodule of their own

dotsdl:
This is just an internal reorganization. Component classes such as `SmallMoleculeComponent` are still exposed at the `gufe` top level, allowing for e.g.:

```
from gufe import SmallMoleculeComponent
```


richardjgowers (approved):

dotsdl:
Fixing tests, then will merge.

PR #42, opened 2022-07-28
Load/Store of transformations/networks

dwhswenson:
This builds on #39 and should only be merged after that one. The additional changes are (as of writing) only in `resultclient.py` (and soon, the associated tests).

This adds the following methods to API for `ResultClient`:

* `store_transformation(transformation)`
* `load_transformation(key) -> Transformation`
* `store_network(network)`
* `load_network(key) -> Network`

Currently, we're using keys for gufe tokenizables that will put those objects at `setup/{QUALNAME}/{TOKEN}.json`, where qualname and token come from the gufe key.


dwhswenson:
This is now blocked by the same problem identified in https://github.com/OpenFreeEnergy/gufe/pull/39#issuecomment-1199651948 and https://github.com/OpenFreeEnergy/gufe/pull/39#issuecomment-1199662648: The serialization via OpenFF's `to_dict` is problematic. In this PR, the issue becomes that conformers are stored as bytes, and Python's JSON serializer can't handle bytes.

Possible solutions:

1. Use custom JSON encoders/decoders to handle bytes. This is relatively straightforward, and we're likely to do it with OpenFF units anyway.

2. Modify the `SmallMoleculeComponent.to_dict()`, either by modifying the `conformers` entry in the OpenFF output, or by completely rewriting the serialization.

dwhswenson:
Update on status here: We've implemented the `SmallMoleculeComponent.to_dict` method, but now we need something for `ProteinComponent`. This is waiting on #45. Once that is in, the existing code might just immediately pass, and will only require unit tests for serialization of `AlchemicalNetwork`.

dotsdl:
I think much of the content here should make its way into `openfe` instead, yes?

dwhswenson:
> I think much of the content here should make its way into `openfe` instead, yes?

I'd like to finish this PR in gufe for historical connection with existing work, but I think that (at minimum) the entirety of `ResultClient` moves to `openfe`. What else moves is a point to discuss further.

There are a few methods here (notably `get_all_gufe_objs`; perhaps also a modified version of `recursive_build_object_cache`) that are reusable enough for other storage approaches that they might be worth including in `gufe`.

dwhswenson:
Should be ready for review. I'd like the merge this in here, so that the diff is easier to see (and because there are a few things that are being added to gufe generally) and then migrate most of the storage stuff to `openfe`.

gufe/storage/resultclient.py
```diff
@@ -92,11 +103,127 @@ def __init__(self, external_store):
     def delete(self, location):
         self._result_server.delete(location)
 
-    def store_protocol_dag_result(self, result):
+    def store_result(self, result):
```
richardjgowers, line 106: UnitResult or DagResult or both?

gufe/storage/resultclient.py
```diff
@@ -92,11 +103,127 @@ def __init__(self, external_store):
...
+            key = self._gufe_key_to_storage_key(prefix, o.key)
+
+            # we trust that if we get the same key, it's the same object, so
+            # we only store on keys that we don't already know
+            if key not in self.result_server:
+                data = json.dumps(o.to_keyed_dict(),
+                                  sort_keys=True).encode('utf-8')
+                self.result_server.store_bytes(key, data)
+
+    def store_transformation(self, transformation):
```
richardjgowers, line 133: is store returning anything?  and/or should it return anything, what might be useful?  A copy of the gufekey it got stored as? (although this is already an attribute I guess)

gufe/storage/resultclient.py
dwhswenson: not part of this PR

gufe/storage/resultclient.py
dwhswenson: There's no additional information to be given by returning anything, although I'm not opposed to it returning something if there's value in it.

gufe/storage/resultclient.py
```diff
@@ -92,11 +103,127 @@ def __init__(self, external_store):
...
         self._result_server.delete(location)
 
-    def store_protocol_dag_result(self, result):
+    def store_result(self, result):
         # I don't know how we get the path information for the protocol dag
         # results
         self.result_server.store(...)
 
+    @staticmethod
+    def _gufe_key_to_storage_key(prefix, key):
```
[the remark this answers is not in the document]
dwhswenson: I think we're going to change the types here (as the comments say several times), but sure, done.

richardjgowers (approved):

PR #43, opened 2022-08-01
to_dict serialization of SmallMoleculeComponent 

richardjgowers:
@dwhswenson this is how I'm thinking of redoing serialisation. As we discussed, the conformers are todo.

I'm anticipating that proteins are a fancier version of this (holding hierarchical information too)

Couple of things:
- I think this might make our sdf code redundant, and we can remove that (in favour of just leaning on rdkit's version of this).
- We still currently use rdkit as our internal molecule, but maybe we should use openff (or maybe we should serialise/deserialise an rdkit molecule inside to_dict/from_dict, it's constructable atom by atom like we're doing here)


gufe/components/smallmoleculecomponent.py
```diff
@@ -79,6 +81,10 @@ class SmallMoleculeComponent(Component, Serializable):
     def __init__(self, rdkit: RDKitMol, name: str = ""):
         name = _ensure_ofe_name(rdkit, name)
         _ensure_ofe_version(rdkit)
+        if not list(rdkit.GetConformers()):
+            # do we require at least 1 or exactly 1?
```
richardjgowers, line 85: I think we can allow more than one, but raise a warning that only the 0th element will be used as the "real" conformation.

dwhswenson:
There's a bunch of stuff in here that should switch from OpenMM units to OpenFF units when openff-toolkit 0.11 comes out. I've marked those with `off-tk 0.11`, and put in comments what the line should change to, so it'll be really easy to make that transition. (I'm assuming we will be requiring openff-toolkit >=0.11 for other reasons, anyway.)

> I think this might make our sdf code redundant, and we can remove that (in favour of just leaning on rdkit's version of this).

RDKit's SDF won't write out properties. We build on the RDKit SDF writer to include properties. We also expect explicit hydrogens, so people using RDKit's SDF as an onramp need to know to set `removeHs=False`. All in all, unless we're talking about fully removing RDKit as a requirement, I think that having a 1-step onramp/offramp to SDF is probably useful.

> We still currently use rdkit as our internal molecule, but maybe we should use openff (or maybe we should serialise/deserialise an rdkit molecule inside to_dict/from_dict, it's constructable atom by atom like we're doing here)

I assume that both RDKit and openff-toolkit will be hard requirements for the foreseeable future. In that case, I don't know that it matters very much. We would need to convert to openff for serialization or to rdkit for generating atom mappings, or visualizations, or whatever else we use based on that.

gufe/components/smallmoleculecomponent.py
dwhswenson: warning added

dwhswenson:
@richardjgowers : I guess since you technically opened the PR, you can't review it (at least, I can't request that you review it.) But this is ready for review. I'll mark it as I've approved it to get the approver thing cleared; feel free to merge if you're happy with it.

We'll have to change the name of the `to_dict`/`from_dict` to the `_to_dict` for #39, but that should be straightforward.

dwhswenson (approved):

PR #44, opened 2022-08-02
remove python 3.8 CI

richardjgowers:


IAlibay:
Do we want to bump to 3.10 or do we have a dependency block here?

richardjgowers:
let's find out I guess

richardjgowers:
Encountered problems while solving:
    - package rdkit-Release_2017_09-py36_0 requires python 3.6*, but none of the providers can be installed

mikemhenry:
I'd vote we add python 3.10 but not require it for merging and we can use that as a canary. 

`rdkit-Release_2017_09` Are we pinning that somewhere?

mikemhenry:
@richardjgowers I think this is ready to go!

PR #45, opened 2022-08-02
Protein serialization

dwhswenson:
This will add protein serialization, which will pass through openff-toolkit but require openff-toolkit 0.11.

Question: Do we combine the `ProteinComponent` and `SmallMoleculeComponent` into a single `MoleculeComponent`? We originally split these because we expected different serialization -- specifically, we planned to write small molecules as SDF and protein as some protein-focused file format (where we never found one that we liked, so we abandoned that approach).

Follow-up: Currently, the `ProteinComponent` has methods to create from PDB or (unimplemented) from PDBx. The `SmallMoleculeComponent` has the ability to get to/from SDF. If combining into a single class, the question is what to do with these methods: allow all (meaning you could read small molecule as PDB -- isn't that problematic?); or remove all. If we remove them, we might still keep conveniences `protein_from_pdb()` and `small_molecule_from_sdf()` as bare methods that return a `MoleculeComponent`.

Note that removing SDF serialization will break network graphml serialization as well, but it is probably better to move the internal representation there to what comes out of our `to_dict`. There's also a case to be made that we no longer have a real need for graphml serialization.


dwhswenson:
(First commits here are just to get the openff-toolkit 0.11.rc into the environment)

dwhswenson:
Updating status here: Problems are as follows:

1. We don't want to duplicate code for serialization via OpenFF molecule, which we're maintaining in gufe (instead of using OpenFF's internal mechanisms) in order to ensure stability. We can use the same mechanism for proteins and small molecules.
2. Protein serialization will require`openff-toolkit` 0.11. That's currently in RC (although the most recent RC has been out for a month, so hopefully a release comes soon?)

Problem 1 is a design issue on our side. In the original post on this PR, I proposed making a single class for both `ProteinComponent` and `SmallMoleculeComponent`. On reflection, I think the better approach is a single superclass/mixin `ExplicitMoleculeComponent` that handles shared serialization, as well as any other aspects that are common here. This allows the individual classes to add in reasonable on-/off-ramps via PDB, SDF, or whatever other file formats we want to support for that specific molecule "type."

Problem 2 is more a policies issue -- can we allow gufe to depend on the RC, until the code we need is released in final form? Commits in this PR so far are just to get the RC into our CI environment.

richardjgowers:
Re 1 isn’t such a problem and tbh maybe we can merge Small and Protein
components.

Re 2, It’s also that the RC doesn’t have icodes yet anyway. Pinning to the
rc isn’t a drama.

RiesBen:
Things to do:
- [ ] read in pdb_file sucht that a nice rdkit is created
- [ ] fix bond-order
- [ ] fix formal charge 
- [ ] make sure to_pdbfile writes correct file
- [ ] write tests
- [ ] final testing

gufe/components/explicitmoleculecomponent.py
```diff
@@ -0,0 +1,200 @@
...
+            for conf in m.conformers
+        ]
+
+        d = {
+            'atoms': atoms,
+            'bonds': bonds,
+            'name': self.name,
+            'conformers': conformers,
+        }
+
```
richardjgowers, line 142: so this looks like it needs the protein "layer" adding in.  One way for this is for `ProteinComponent` to have a `_to_dict` which calls `super()` and then adds in a few extra entries to the dict created here

off the top of my head, extra things include:
- resname, resid, icode per residue
- chain ids per chain
- grouping of atoms into residues (`[[0, 1, 2], [3, 4, 5], ...]`)
- grouping of residues into chains (`[[0, 1, 2, ....], [400, 401, ...]]`)

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,224 @@
...
+            a.SetProp("name", atom.name)
+            a.SetIntProp("id", atomID_orig)
+
+            a.SetProp("resName", atom.residue.name)
+            a.SetIntProp("resId", int(atom.residue.index))
+            
+            # add residue information
+            mi  =  Chem.AtomPDBResidueInfo()
+            mi.SetResidueName(atom.residue.name)
+            mi.SetResidueNumber( int(atom.residue.index))
```
richardjgowers, line 110: insertion codes?

richardjgowers:
also we should try and push the bond order updates to the xml upstream to openmm, then we can hopefully unvendor this...

gufe/components/proteincomponent.py
RiesBen: ? this is for pdb resdiue naming in the write out :)

RiesBen: I moved it to the writeout - I think that makes it safer.

RiesBen: ah wait I think I might not know what an insertion code is.

RiesBen: added it, still don't know what it exactly does ;)

gufe/components/explicitmoleculecomponent.py
RiesBen: should be there :)

gufe/components/sub_files/data/pdbNames.xml
```diff
@@ -0,0 +1,278 @@
+<Residues>
```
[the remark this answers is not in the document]
RiesBen: going to happen soon ;)
but its residues.xml at the moment, that changed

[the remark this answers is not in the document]
RiesBen: The current vendored files are only temporary as we plan to push these changes upstream to openMM and then remove these temporary files from our repo again! : )

RiesBen:
Hey everyone,
the merge request should be functional! 
right now waiting for the CI (locally testing was fine).

please ignore formating of any file in sub_files, as they are going to vanish pretty soon again. I will after this merge push these changes to openmm and in a second iteration remove the sub_files folder again.

Thanks for your review!
Cheers

RiesBen:
vendored code  changes were pushed upstream: :)
https://github.com/openmm/openmm/pull/3770

RiesBen:
TODO: One day this should use rdkit's MonomerInfo so that we go into openff correctly. - DONE :)

RiesBen:
Missing pieces:
-[ ] excluding sub_files folder from CI check (help on this pls.)

rest I think is done. (for now)

dwhswenson:
* Can we move `gufe/component/sub_files/` to `gufe/vendor/pdbfile/` or something like that? This way we can block the `vendor` directory from things like coverage, and that will be constant as we move things in and out of vendored status. It's also a more standard way of communicating to code readers that this is vendored code (which means that you shouldn't change it unless necessary).
* I don't think the core repo is the right place for the ipynb file in the dev directory. We need to establish a right place for it (it is a good thing to have), but I don't think we want the churn of notebooks in here if we can avoid.

This is only a partial review, but wanted to get a few thoughts in place ASAP. I'll take a deeper look soon.

RiesBen:
@dwhswenson  good comments, done :)

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
+        openmm_pos = self.to_openmm_positions()
+
+        # write file
+        if isinstance(out_path, str):
+            out_file = open(out_path, "w")
+        elif isinstance(out_path, io.TextIOWrapper):
+            out_file = out_path
+            out_path = str(out_file.name)
+        else:
+            raise ValueError("Out path type was not as expected!")
```
dwhswenson, line 514: This breaks the ability to use `io.StringIO`.

I think you really just want something here like:

```python
if not isinstance(out_path, io.TextIOBase):
    # allows pathlike/str; we close on completion
    out_path = open(out_path, mode='w')
    must_close = True

PDBFile.writeFile(open_top, openmm_pos, file=out_path)

if must_close:
    # we only close the file if we had to open it
    out_path.close()
```

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
+        # get top:
+        top = self.to_openmm_topology()
+
+        # get pos:
+        np_pos = deserialize_numpy(self.to_dict()["conformers"][0])
+        openmm_pos = (
+            list(map(lambda x: np.array(x), np_pos)) * omm_unit.angstrom
+        )
+
+        # write file
```
dwhswenson, line 547: see comment on pdbfile

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
+            "conformers": conformers,
+            "molecules": molecule_props,
+        }
+
+        return d
+
+    # NOT implemented:
+    def to_openeye(self) -> OEMol:
+        """OEChem representation of this molecule"""
+        raise NotImplementedError
```
dwhswenson, line 639:
```suggestion
        raise NotImplementedError()
```

& same throughout

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
-        raise NotImplementedError()
+    def from_gmx(cls, gmx_top: str, gmx_gro: str):
+        raise NotImplementedError
 
-    @property
-    def total_charge(self):
-        return None
+    @classmethod
+    def from_amber(cls, name=""):
+        raise NotImplementedError
```
dwhswenson, line 664: is it worth even putting the skeletons of these in now? are we even planning to support on-ramp from a .gro of a protein?

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
 
-    def _defaults(self):
-        return super()._defaults()
+    @classmethod
+    def from_pdbxfile(cls, pdbxfile: str, name=""):
+        raise NotImplemented
```
dwhswenson, line 668: did this not work? It looked like all the code for it should have been implemented.

dwhswenson (commented):
Core concepts are fine; lots of little things to change that will help make the code more maintainable in the long run.

My little code samples aren't tested, so I could be wrong in whether they'd work, but the goal is to simplify things where possible.

One other question: Do we want to include all the PDB files or are these files that we can download on the fly? This seems like a lot to add to the repo, and for users to have working tests, it will be part of the footprint of the package as well. @richardjgowers : thoughts on that?

gufe/tests/test_proteincomponent.py
```diff
@@ -0,0 +1,301 @@
...
+        if("CREATED WITH OPENMM" in in_line):
+            continue
+        out_line = out_lines[i]
+        if(in_line == out_line):
+            continue
+        else:
+            not_equal.append((in_line, out_line))
+
+    not_equal_found = len(not_equal) > 0
+    if(not_equal_found):
```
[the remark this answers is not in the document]
RiesBen: oooooh! ok cool! :)

gufe/tests/test_proteincomponent.py
```diff
@@ -0,0 +1,301 @@
...
+    in_pdb = open(in_file_path, "r")
+    out_pdb = open(out_file_path, "r")
+
+    in_lines = in_pdb.readlines()
+    out_lines = out_pdb.readlines()
+
+    not_equal = []
+    assert len(in_lines) == len(out_lines)
+
+    for i, in_line in enumerate(in_lines):
```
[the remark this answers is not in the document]
RiesBen: does this then make sense? 

```
    in_lines = in_lines[1:]
    out_lines = out_lines[1:]
    
    assert len(in_lines) == len(out_lines)
    assert in_lines == out_lines
```

or would you remove the len(x) == len(y) as well? :)

gufe/vendor/pdb_file/singelton.py
```diff
@@ -0,0 +1,15 @@
+"""
```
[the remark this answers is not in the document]
RiesBen: in my current branch this is correct?  hm... did you change it?

gufe/components/proteincomponent.py
RiesBen: na, I will remove them.

gufe/components/proteincomponent.py
RiesBen: In theory it should be there, but somehow, there is a problem (maybe with our testfile?)

RiesBen: image

something is weird with the connectivity.... as it is 0 for this atom (which it definitly should not be and is correctly assigned in Topology, somehow it is overwritten somewhere?/unlinked?)

gufe/components/explicitmoleculecomponent.py
```diff
@@ -0,0 +1,124 @@
...
+    def from_json(cls, json_str):
+        dct = json.loads(json_str)
+        return cls.from_dict(dct)
+
+    @classmethod
+    def from_rdkit(cls, rdkit: RDKitMol, name: str = ""):
+        """Create a SmallMoleculeComponent, copying from an RDKit Mol"""
+        return cls(rdkit=Chem.Mol(rdkit), name=name)
+
+    # Not Implemented - interface Functions:
```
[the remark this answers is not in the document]
RiesBen: hmmm true, but this would look like this:
image

And if I understand the code correctly the behaviour would change right :)

gufe/components/proteincomponent.py
```diff
@@ -1,86 +1,671 @@
...
+            Quantity containing protein atom positions
+        """
+        np_pos = deserialize_numpy(self.to_dict()["conformers"][0])
+        openmm_pos = (
+            list(map(lambda x: np.array(x), np_pos)) * omm_unit.angstrom
+        )
+
+        return openmm_pos
+
+    def to_pdbFile(self, out_path: Union[str, io.TextIOWrapper] = None) -> str:
```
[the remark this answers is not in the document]
RiesBen: Hm I like this behaviour, so the out_path can be created in the function call:
'''python 
out_path = to_pdbFile(out_path=out_dir+"/"+out_file_name)
'''

but I guess depends.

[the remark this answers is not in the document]
RiesBen: might be called lazy

gufe/components/proteincomponent.py
RiesBen: hm.... ok I got it, the atom names were not recognized by openmm.... fixed.

gufe/components/proteincomponent.py
RiesBen: hm on TextIOBase I get a mypy error:
image

However I'm not sure what the TextIO type should be, as it seems not to be accesible???

gufe/tests/conftest.py
```diff
@@ -41,12 +41,48 @@ def PDB_181L_path():
...
+        "thrombin_protein_openmmClean.pdb",
+        "tnsk2_protein_openmmClean.pdb",
+        "tyk2_protein_openmmClean.pdb",
+        ]
+
+    pdb_paths = []
+    for pdb_path in pdb_filenames:
+        with importlib.resources.path('gufe.tests.data', pdb_path) as f:
+            pdb_paths.append(str(f))
+    yield pdb_paths
```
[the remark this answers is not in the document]
RiesBen: I like the idea, but currently I don't manage to implement it.

gufe/components/explicitmoleculecomponent.py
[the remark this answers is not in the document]
dwhswenson: The decorator [`@abc.abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) means that you need to provide an implementation in a subclass -- it will raise an error if you try to create an instance of a subclass that doesn't instantiate this.

That said, it is technically possible to call this as `GufeTokenizable._to_dict(some_obj)`, which would currently return None instead of raising a `NotImplementedError`. I'd say that is incorrect behavior in `GufeTokenizable`, and should be changed there. That one is an edge case where I can't imagine it mattering in practice. However, with `GufeTokenizable._from_dict`, you can actually call `GufeTokenizable._from_dict({})` without error, and you would normally call a classmethod without needing an instance, so this is worth fixing. (Not as part of this PR: I'll fix it separately.) 

However, back to the original point for this PR: you don't need to add these functions here, and it's actually problematic since you don't have them decorated with `abc.abstractmethod`.)

RiesBen (approved):

dwhswenson:
TODO: (for my own reference when I come back to this on Monday):

- [ ] mark some should-be-impossible sections of code as `-no-cov-`
- [ ] add tests to ensure that all supported input types (string pathname, `pathlib.Path`, `io.StringIO`, `io.TextIOWrapper`) have at least smoke tests to ensure that usage works without error.

This should bring the patch coverage above target (although I might not try to get the patch coverage to 100%).

PR #46, opened 2022-08-02
add latest openff-toolkit to matrix

mikemhenry:
This might do the opposite of what I want, but haven't found a way locally to test GHA other than https://github.com/rhysd/actionlint but that is only static


mikemhenry:
@richardjgowers I ran out of time to finish this off but if you want to look at it, I am either doing something really silly OR there is an issue with serializing these units from the new toolkit.

dwhswenson:
@mikemhenry : see fixes in #45.

dwhswenson:
This is why I said leave it red until openff-toolkit 0.11 is released. Among other things, that switches from `openmm.unit` to openff.units`. There's no reason for us to support both -- we'll pin to the newer version.

mikemhenry:
~Your right we don't need to support both, but is there any reason to support a version other than 0.11? I'm not sure if we should wait for it to be released when we could just install from the RC channel. I see in #45 that is what you are doing.~

Ah I see what you are saying now! Yes lets keep it red but now it is in the matrix so we can make sure that PRs are set to support the new toolkit.

mikemhenry:
@richardjgowers @dwhswenson 
I think we are good to go, now it will just stay red, and with the name changes it is easy to tell when things are working on the new toolkit vs the old one.

dotsdl:
Is this PR still relevant now that `openff-toolkit` 0.11.0 is out?

richardjgowers:
we're pure offtk 0.11 now, so this is redundant

PR #47, opened 2022-08-12
rename base to tokenize_base

richardjgowers:
@dotsdl `base.py` is a bad name, let's be more descriptive


dotsdl:
Let's call it `tokenize.py`.

dotsdl (approved):
Thanks @richardjgowers! LGTM

PR #51, opened 2022-08-27
openff tk made optional dependency

richardjgowers:
openff tk doesn't support M1s for now, so make it optional for dev reasons


dwhswenson:
I think you may find a lot of issues if you try to use an Apple Silicon env right now -- lots of things don't seem to have it yet. Try installing as an [x86 environment](https://towardsdatascience.com/how-to-manage-conda-environments-on-an-apple-silicon-m1-mac-1e29cb3bad12) (may also require [installing Rosetta 2](https://osxdaily.com/2020/12/04/how-install-rosetta-2-apple-silicon-mac/)). This is what is working for me.

richardjgowers:
Yeah I think I’m gonna also have to do that eventually, this gets a lot of
things working at least

PR #56, opened 2022-09-13
an attempt to simplify ProtocolUnits:

richardjgowers:
- dependencies are now explicity given in a separate arg
- all other arguments are "simple" json-serializables

### DIFF
--- a/gufe/protocols/protocolunit.py
+++ b/gufe/protocols/protocolunit.py
@@ -5,6 +5,8 @@
 part of a `ProtocolDAG`.
 
 """
+from __future__ import annotations
+
 
 import abc
 from dataclasses import dataclass
@@ -30,29 +32,11 @@ class Context:
     shared: PathLike
 
 
-class ProtocolUnitMixin:
-
-    def _list_dependencies(self, cls):
-        deps = [] 
-        for key, value in self.inputs.items():
-            if isinstance(value, dict):
-                for k, v in value.items():
-                    if isinstance(v, cls):
-                        deps.append(v)
-            elif isinstance(value, list):
-                for i in value:
-                    if isinstance(i, cls):
-                        deps.append(i)
-            elif isinstance(value, cls):
-                deps.append(value)
-        return deps
-
-
-class ProtocolUnitResultBase(GufeTokenizable, ProtocolUnitMixin):
+class ProtocolUnitResultBase(GufeTokenizable):
     def __init__(self, *, 
             name: Optional[str] = None, 
             source_key: GufeKey, 
-            inputs: Dict[str, Any], 
+            dependencies: Dict[str, GufeKey],
             outputs: Dict[str, Any],
             _key: GufeKey = None
         ):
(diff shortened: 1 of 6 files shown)


dotsdl:
Sorry I missed this one @richardjgowers; I believe I was just about to travel when you created this one. Would you still like a review?

richardjgowers:
@dotsdl yeah I think I still like this change enough to discuss.  The tl;dr is that Units get their dependencies separately (rather than as any argument).  This then simplifies things like `execute` as you don't have to scan every object, and identifying dependencies becomes simpler (they're all in one place).  I don't see it as a huge annoyance to have to separate out dependencies in setting up Units.

gufe/protocols/protocolunit.py
```diff
@@ -186,10 +155,10 @@ class ProtocolUnitFailure(ProtocolUnitResultBase):
 
     """
 
-    def __init__(self, *, name=None, source_key, inputs, outputs, _key=None, exception, traceback):
+    def __init__(self, *, name=None, source_key, dependencies, outputs, _key=None, exception, traceback):
         self._exception = exception
         self._traceback = traceback
-        super().__init__(name=name, source_key=source_key, inputs=inputs, outputs=outputs, _key=_key)
+        super().__init__(name=name, source_key=source_key, dependencies=dependencies, outputs=outputs, _key=_key)
```
dotsdl, line 162: I think we'll still need a separate `inputs` kwarg here?

gufe/tests/test_protocol.py
```diff
@@ -20,43 +20,42 @@
...
 
 
 class FinishUnit(ProtocolUnit):
-    @staticmethod
-    def _execute(ctx, *, simulations, **inputs):
+    def _execute(self, ctx, **dependencies):
+        # sims are given a string of numeric key, extract them in ascending order
+        _, sims = zip(*sorted(dependencies.items(), key=lambda x: int(x[0])))
 
-        output = [s.outputs['log'] for s in simulations]
```
dotsdl, line 56: Though this works, I think it introduces a bit too much dependency on naming conventions used when initializing a `ProtocolUnit`, such as `FinishUnit` here. Instead of a list of `SimulationUnit`s in `simulations`, now we have a dict with a very particular convention for names. This gets even more complicated if other `ProtocolUnit`s that aren't `SimulationUnit`s need to be included as dependencies.

Although we have no real examples so far of this, I believe the existing system makes it easier to write reusable `ProtocolUnit`s that are usable across many `Protocol`s. One could create a package of many `Protocol`s that use a set of `ProtocolUnit`s defined in the package as primitives, but this would get difficult to do if a given `ProtocolUnit` must take in a dict with a particular convention used for input instead of more Pythonic arguments.

Thoughts?

gufe/protocols/protocolunit.py
```diff
@@ -30,29 +32,11 @@ class Context:
...
-        return deps
-
-
-class ProtocolUnitResultBase(GufeTokenizable, ProtocolUnitMixin):
+class ProtocolUnitResultBase(GufeTokenizable):
     def __init__(self, *, 
             name: Optional[str] = None, 
             source_key: GufeKey, 
-            inputs: Dict[str, Any], 
+            dependencies: Dict[str, GufeKey],
```
dotsdl, line 39: I think we'll still need a separate `inputs` kwarg here? Otherwise we're dropping some of the `ProtocolUnit` state on the floor when we create its `ProtocolUnitResult`.

gufe/protocols/protocolunit.py
```diff
@@ -323,16 +295,19 @@ def execute(self, *,
...
 
         try:
-            outputs = self._execute(context, **inputs)
+            outputs = self._execute(context, **dependencies)
             result = ProtocolUnitResult(
-                name=self.name, source_key=self.key, inputs=inputs, outputs=outputs
+                name=self.name,
+                source_key=self.key,
+                dependencies=dependencies,
+                outputs=outputs
```
dotsdl, line 303: I think we'll still need a separate `inputs` kwarg here? Otherwise we're dropping some of the `ProtocolUnit` state on the floor when we create its `ProtocolUnitResult`.

dotsdl (commented):
A few comments. I still prefer the existing system, which does feature a bit more magic under the hood, but I think avoids some complexity for a protocol developer. We'll want some feedback from @ijpulidos on this as well.

PR #59, opened 2022-09-20
Old offtk

richardjgowers:
Something screwy is happening with offtk 0.11 in openfe-land, so we need (temporary) compatibility with offtk 0.10 back again

### DIFF
--- a/gufe/components/smallmoleculecomponent.py
+++ b/gufe/components/smallmoleculecomponent.py
@@ -5,8 +5,21 @@ import logging
 logger = logging.getLogger('openff.toolkit')
 logger.setLevel(logging.ERROR)
 from openff.toolkit.topology import Molecule as OFFMolecule
+
+
+def _old_OFF() -> bool:
+    from openff.toolkit import __version__ as OFF_VERSION
+    a, b, c = map(int, OFF_VERSION.split('.'))
+    return (a, b) < (0, 11)
+
+
+USING_OLD_OFF = _old_OFF()
+
 import warnings
-from openff.units import unit
+if USING_OLD_OFF:
+    from openmm import unit as omm_unit
+else:
+    from openff.units import unit as off_unit
 
 from rdkit import Chem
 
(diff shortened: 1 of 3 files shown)


gufe/components/smallmoleculecomponent.py
```diff
@@ -5,8 +5,29 @@
 logger = logging.getLogger('openff.toolkit')
 logger.setLevel(logging.ERROR)
 from openff.toolkit.topology import Molecule as OFFMolecule
+
+
+def _old_OFF() -> bool:
+    from openff.toolkit import __version__ as OFF_VERSION
+    a, b, c = map(int, OFF_VERSION.split('.'))
+    return (a, b) < (0, 11)
```
dwhswenson, line 13: This will error on import if a user is using a release candidate. If we'l need this with any regularity, the thing to use is `packaging.version.Version`.

dwhswenson (commented):
I pushed some changes in [862ea7d](https://github.com/OpenFreeEnergy/gufe/pull/59/commits/862ea7d5cb5bd480a8eb886f8443d14d936ca007) that I think make it a little cleaner, and certainly easier to maintain if we need to keep this around for any length of time. If you hate them, you can revert back.

* If we're going to maintain support against multiple versions, that should be added to the CI matrix. This will fix the codecov failures.
* As mentioned in comment below, this will break in environments with an RC installed (which is actually what I had locally).

mikemhenry:
Bit of an example on how to support both toolkits for the units: https://github.com/openmm/openmmforcefields/pull/237/files (if needed)

dwhswenson:
I like @mikemhenry's solution for a couple points:

* Use API to decide which API we use (does it have `atom.element`?) instead of parsing version. We'll still need to define a different atomic number extractor for each, but this solves the problem with running under an RC.
* Takes similar approach to mine where the unit-package-selection logic is abstracted away from the specific details. I didn't know about the `ensure_quantity` from `openff.units.openmm`, which is why I hacked my own.

mikemhenry:
> I didn't know about the ensure_quantity from openff.units.openmm, which is why I hacked my own.

It was just added yesterday? to help make this easier, so that isn't your fault 🤣

PR #62, opened 2022-09-22
Fix `ProtocolUnit` deserialization

dwhswenson:
Objects that set their `_key` in their `_from_dict()` also need to register that key with the TOKENIZABLE_REGISTRY. This fixes it for `ProtocolUnit`.

You only notice the error if you deserialize in a different process from the one you serialized in (or at least if you wipe the registry between serialization and deserialization.)

The first commit is a rough fix; I think I'd rather add a private `_set_key()` method on `GufeTokenizable`. This might also remove the old key (generated on initialization) from the registry, and makes the process re-usable for other classes if needed.

Tasks:

- [x] Write `_set_key` method
- [x] Write test for `_set_key` on dummy object
- [x] Use `_set_key` for `ProtocolUnit`
- [x] Test `ProtocolUnit` serialization


richardjgowers:
the `_set_key` method seems less fragile

gufe/protocols/protocolunit.py
```diff
@@ -269,6 +270,7 @@ def _from_dict(cls, dct: Dict):
         obj = cls(name=dct['name'],
                   **dct['inputs'])
         obj._key = _key
+        TOKENIZABLE_REGISTRY.setdefault(_key, obj)
```
richardjgowers, line 273: this should be using `_set_key` right?

dwhswenson:
This is ready for review. I'd like to merge it before I go too much further into working on #90.

gufe/tokenization.py
```diff
@@ -382,7 +399,7 @@ def from_dict(dct) -> GufeTokenizable:
     # was to a deleted object.
     thing = TOKENIZABLE_REGISTRY.get(obj.key)
 
-    if thing is None:
+    if thing is None:  # -no-cov-
```
richardjgowers, line 402: can we test this by saving a dumped object in the testsuite and loading it in a test?

dwhswenson: > can we test this by saving a dumped object in the testsuite and loading it in a test?

No, this line cannot be tested in the current implementation. It should never be reached; but that's likely to change in #90.

We already do the test you're asking for, by clearing the registry and loading a dumped object (in fact, we do this for *every* object that uses the `GufeTokenizableTestsMixin`).

To see why this doesn't hit that line, note that we can't get `obj.key` in line 400 unless `obj` already exists. Current approach to get the key is to create an object, then get the key from that. On creation, it gets added to the `TOKENIZABLE_REGISTRY`.  So you should have always created an object by here; I actually think this code should be unreachable. This PR affects objects with nondeterministic keys (essentially, random numbers on creation). I think the reason this was hit before was that reloading an object went like this:

1. Create object that should have `key=key1`: on creation it gets `key` that is randomly assigned, call that `key2`. `key1 != key2`. The `TOKENIZABLE_REGISTRY` knows there is an object called `key2`.
2. Set the `key=key1`.
3. Get to line 400 above: `obj.key` gives `key1`, but the registry only has `key2`.

In the new code, step 2 both removes `key2` from the registry and adds `key1`. So we always find an object there.

I'm hoping to change the requirement to recreate the entire object in #90, so that we only generate the key, not a copy of the object. In that case, this line will be reached every time we reload an unknown object.

richardjgowers (approved):

richardjgowers:
@dwhswenson do you want this merged now? I'm working on docs in gufe so go ahead if so

dwhswenson:
I was going to give @dotsdl the rest of the day to take a look, if he wants to. It might affect stuff he's doing downstream, too. Plan to merge by end of my work day (leave it open for at least 6 hours, merging no earlier than Fri 04 Nov 22:00 GMT (17:00 my local).)

dotsdl (approved):
Looks great @dwhswenson! Thanks for this! Merge it!

PR #65, opened 2022-09-26
Use clear names openff-toolkit versions in CI

dwhswenson:
Currently, the labels that get shown on GitHub for the individual actions are "true" and "false", where "true" means "using old OpenFF toolkit (0.10.6)" and false means "using current release of OpenFF toolkit."

Ever since this was added to the matrix, branch protection is complaining because the old action names (without "true"/"false") no longer exist. I figured I'd change this to something more informative, and update branch protection rules based on this.

I know that @richardjgowers and @dotsdl are both away now, so asking @mikemhenry to review; but let's not merge until tomorrow morning (US time) in case they want to review.


dwhswenson:
... how is mypy failing on a change that only affected the CI file? (I won't have time to figure this out today.)

mikemhenry (approved):
LGTM! Thanks for making that more clear

dwhswenson:
mypy problem identified: https://mypy-lang.blogspot.com/2022/09/mypy-0981-released.html (under "Generate Error on Unbound TypeVar Return Type")

The way we define OEMol as a typevar is no longer considered legitimate. That is:

```python
# in custom_typing.py
OEMol = TypeVar('OEMol')

# in smallmoleculecomponent.py
def to_openeye(self) -> OEMol:
    ...
```

will raise an error. Annoyingly, I can't seem to get the new mypy from conda-forge to reproduce locally.

I'm new enough to typing that I don't know what the "right" way to fix this is -- maybe just use a string literal; anyone have more knowledge?

richardjgowers:
Honestly we could just remove the typing or the entire method…

dwhswenson:
> Honestly we could just remove the typing or the entire method…

yeah, I'm marking it "ignore" in the protein serialization branch.

PR #66, opened 2022-09-28
[WIP] Fix protein serialization

dwhswenson:
When I tried to actually use protein serialization, it didn't work. I'll try to add a regression test to figure the problem, but these changes fixed it for #42.

I'm a little unhappy here with the idea that we're serializing ints from RDKit enums; I assume those are unlikely to change, but it does mean that our serialization is only as stable as those mappings. If there's a better approach, I'm open to suggestion!

Tasks:

- [ ] Use vendored versions of the rdkit enums
- [ ] Add unit test for serialization problem

### DIFF
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -12,7 +12,9 @@ from openmm import app
 from openmm import unit as omm_unit
 
 from rdkit import Chem
-from rdkit.Chem.rdchem import Mol, Atom, Conformer, EditableMol, BondType
+from rdkit.Chem.rdchem import (
+    Mol, Atom, Conformer, EditableMol, BondType, AtomMonomerType
+)
 
 from .explicitmoleculecomponent import ExplicitMoleculeComponent
 from ..vendor.pdb_file.pdbfile import PDBFile
@@ -330,6 +332,11 @@ class ProteinComponent(ExplicitMoleculeComponent):
             # put mi_dict back to class
             atom_monomerInfo = Chem.AtomPDBResidueInfo()
             for key, val in atom_mi_dict.items():
+                if key == "MonomerType":
+                    # DANGER: this risks tying serialization to specific
+                    # RDKit versions
+                    val = AtomMonomerType.values[val]
+
                 f = getattr(atom_monomerInfo, "Set" + str(key))
                 f(val)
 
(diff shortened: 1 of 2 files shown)


gufe/components/proteincomponent.py
```diff
@@ -330,6 +332,11 @@ def _from_dict(cls, ser_dict: dict, name: str = ""):
             # put mi_dict back to class
             atom_monomerInfo = Chem.AtomPDBResidueInfo()
             for key, val in atom_mi_dict.items():
+                if key == "MonomerType":
+                    # DANGER: this risks tying serialization to specific
+                    # RDKit versions
+                    val = AtomMonomerType.values[val]
```
richardjgowers, line 338: I'd be curious which values from the enum we're using.  Aren't we only using PDBResidue here?

gufe/components/proteincomponent.py
```diff
@@ -346,7 +353,7 @@ def _from_dict(cls, ser_dict: dict, name: str = ""):
         for bond in ser_dict["bonds"]:
             atomBeginIdx = bond[0]
             atomEndIdx = bond[1]
-            bondType = bond[2]
+            bondType = BondType.values[bond[2]]
```
richardjgowers, line 356: If we're worried about rdkit breaking their enums we could do have our own equivalent enum for serialisation, but I'd be surprised if they broke this.

dwhswenson: This was my thought, too. I think the risk is very small, but it is nonzero. It's more likely that rdkit might expand their enums with new items, and this way we'd automatically have those new items be available. But maybe that doesn't matter for our use cases.

Anyway, I'm fine with it either way. Let me know which you prefer.

gufe/components/proteincomponent.py
dwhswenson: Yeah, I think that's the only one.

richardjgowers: So PDBResidue is actually the only possible MonomerType (currently), so we're OK to just hardcode to this tbh

richardjgowers (commented):
I think after the rewrite I did, this might have also been fixed.

dwhswenson:
> I think after the rewrite I did, this might have also been fixed.

I suspect it probably did. I'll merge those changes into #42; if the problems I had there are fixed, then I'll close this PR.

PR #68, opened 2022-09-29
Add custom_json from OpenPathSampling / SimStore

### LINKED ISSUE #64: natively key mangle common non-json-friendly objects
for example, Path, Quantity / unit

dwhswenson:
This adds in the custom JSON setup I originally developed for SimStore. Quick overview:

#### What's wrong with Python's JSON

`json.dumps` and `json.loads` both provide a couple ways to extend to arbitrary types: you can either provide a custom function (`default` for dump, `object_hook` for load) or you can provide a custom class (`cls` for both). That class will subclass `json.JSONEncoder` or `json.JSONDecoder` and implement either `default` or `object_hook`.

The problem is that these approaches are not composable. So if you have an encoder for OpenFF units and another encoder for pydantic models, you need to write yet a third encoder to combine the two.

#### What SimStore does

In SimStore, I created a `JSONCodec` class that combines the functionality of the `JSONEncoder` and `JSONDecoder`. Then I created a `JSONSerializationDeserialization` class that takes multiple `JSONCodec`s on initialization and dynamically generates a `json.JSONEncoder` and `json.JSONDecoder` from them. You then have a global instance of `JSONSerializationDeserialization` that you can use. Furthermore, this allows addition of codecs at runtime.

This could be used as `json.dumps(obj, cls=GLOBAL_SERIALIZER.encoder)`, or you can directory use the `serializer` / `deserializer` methods, which return the callable (that has to do with SimStore's model; we'll probably just use the `encoder`/`decoder` attributes.

The API for `JSONCodec` is pretty simple. In general, you need to define a `to_dict` function for the object and a `from_dict` function. The codec is an instance of `JSONCodec` that takes the `to_dict` and `from_dict` as parameters. You also need to provide a way to recognize 

Advantages:

* allows arbitrary composition of individual codecs
* even allows addition of codecs at runtime (e.g., from external packages)
* code is already in place and tested

#### Practical usage

To add a codec for another type, in principle you need need 4 things:

* a way of identifying whether the in-memory object should be serialized by this codec (`is_my_obj`); this defaults to an `isinstance` check on the provided class `cls`
* a way of converting that into a JSON-ready format; this is the user-provided `to_dict` method
* a way of identifying whether a given dict should be deserialized by this codec (`is_my_dict`), which should be the default method unless a user
* a way of converting the dictionary to the in-memory object; this is the user-provided `from_dict` method

In practice, you typically only need to provide `to_dict` and `from_dict` (and the class so we can do an `isinstance` check). However, if you want to get into the weeds, you can customize `is_my_obj` and `is_my_dict`.

You can see examples of this in the tests, where I create a custom codec for `numpy.ndarray`. Since the `to_dict` of that actually returns `bytes`, I also create a custom codec for `bytes`.

Resolves #64.

### DIFF
--- a/gufe/custom_json.py
+++ b/gufe/custom_json.py
@@ -1,0 +1,165 @@
+# This file was originally part of OpenPathSampling/SimStore.
+
+import json
+import functools
+import inspect
+from collections import namedtuple, defaultdict
+
+
+class JSONSerializerDeserializer(object):
+    """
+    Tools to serialize and deserialize objects as JSON.
+
+    This wrapper object is necessary so that we can register new codecs
+    after the original initialization.
+
+    Attributes
+    ----------
+    encoder:
+        subclass of ``JSONEncoder``; use as ``json.dumps(obj, cls=encoder)``
+    decoder:
+        subclass of ``JSONDecoder``; use as ``json.loads(string,
+        cls=decoder)``
+
+    Parameters
+    ----------
+    codecs : list of :class:`.JSONCodec`s
+        codecs supported
+    """
+    def __init__(self, codecs, named_codecs=None):
+        self.named_codecs = {} if named_codecs is None else named_codecs
+        self.codecs = []
+        for codec in codecs:
+            self.add_codec(codec)
+
+        self.encoder, self.decoder = self._set_serialization()
+
+    def _set_serialization(self):
+        encoder, decoder = custom_json_factory(self.codecs)
+        self._serializer = functools.partial(json.dumps, cls=encoder)
+        self._deserializer = functools.partial(json.loads, cls=decoder)
+        return encoder, decoder
+
+    def add_codec(self, codec):
+        """Add a new codec to the supported codecs
+
+        Parameters
+        ----------
+        codec : :class:`.JSONCodec`
+            codec to add
+        """
+        if codec in self.codecs:
+            return
+
+        if codec is not None:
+            self.codecs.append(codec)
+
+        self.encoder, self.decoder = self._set_serialization()
+
+    def replace_named_codec(self, codec_name, codec):
+        self.named_codecs[codec_name] = codec
+        self.add_codec(None)
+
+    def serializer(self, obj):
+        """Callable that dumps to JSON"""
+        return self._serializer(obj)
+
+    def deserializer(self, string):
+        """Callable to loads JSON"""
+        return self._deserializer(string)
+
+
+def custom_json_factory(coding_methods):
+    """Create JSONEncoder/JSONDecoder for special types
+    """
+    class CustomJSONEncoder(json.JSONEncoder):
+        def default(self, obj):
+            for coding_method in coding_methods:
+                result = coding_method.default(obj)
+                if result is not obj:
+                    return result
+            return json.JSONEncoder.default(self, obj)
+
+    class CustomJSONDecoder(json.JSONDecoder):
+        def __init__(self, *args, **kwargs):
+            super(CustomJSONDecoder, self).__init__(
+                object_hook=self.object_hook, *args, **kwargs
+            )
+
+        def object_hook(self, dct):
+            for coding_method in coding_methods:
+                result = coding_method.object_hook(dct)
+                if result is not dct:
+                    return result
+            return dct
+
+    return (CustomJSONEncoder, CustomJSONDecoder)
+
+
+class JSONCodec(object):
+    """Custom JSON encoding and decoding for non-default types.
+
+    Parameters
+    ----------
+    cls : class
+        Class for this codec. Assumes that all subclasses should be treated
+        the same way. Can be ``None`` if ``is_my_obj`` and ``is_my_dict``
+        are given.
+    to_dict : Callable
+        method that converts the object to a dictionary
+    from_dict : Callable
+        method that restores the object based on the dictionary made by
+        to_dict
+    is_my_obj : Optional[Callable]
+        Method to determine whether the input object should be treated by
+        this encoder. Default behavior is to use ``isinstance(cls)``, and to
+        create a dict that also includes the class name and the module of
+        the object.
+    is_my_dict : Optional[Callable]
+        Method to determine whether the input dictionary should be treated
+        by this decoder. Default behavior assumes usage of the default
+        ``is_my_obj``.
+    """
+    def __init__(self, cls, to_dict, from_dict, is_my_obj=None,
+                 is_my_dict=None):
+        if is_my_obj is None:
+            is_my_obj = self._is_my_obj
+
+        if is_my_dict is None:
+            is_my_dict = self._is_my_dict
+
+        self.cls = cls
+        self.to_dict = to_dict
+        self.from_dict = from_dict
+        self.is_my_obj = is_my_obj
+        self.is_my_dict = is_my_dict
+
+    def _is_my_dict(self, dct):
+        expected = ['__class__', '__module__', ':is_custom:']
+        is_custom = all(exp in dct for exp in expected)
+        if is_custom:
+            return (dct['__class__'] == self.cls.__name__
+                    and dct['__module__'] == self.cls.__module__)
+
+    def _is_my_obj(self, obj):
+        return isinstance(obj, self.cls)
+
+    def default(self, obj):
+        if self.is_my_obj(obj):
+            dct = {}
+            if self.cls:
+                dct.update({
+                    '__class__': self.cls.__name__,
+                    '__module__': self.cls.__module__,
+                    ':is_custom:': True,
+                })
+            # we let the object override __class__ and __module__ if needed
+            dct.update(self.to_dict(obj))
+            return dct
+        return obj
+
+    def object_hook(self, dct):
+        if self.is_my_dict(dct):
+            obj = self.from_dict(dct)
+            return obj
+        return dct
--- a/gufe/tests/test_custom_json.py
+++ b/gufe/tests/test_custom_json.py
@@ -1,0 +1,117 @@
+from gufe.custom_json import (
+    JSONSerializerDeserializer, custom_json_factory, JSONCodec
+)
+import json
+import pytest
+
+import numpy as np
+from numpy import testing as npt
+
+bytes_codec = JSONCodec(
+    cls=bytes,
+    to_dict=lambda obj: {'latin-1': obj.decode('latin-1')},
+    from_dict=lambda dct: dct['latin-1'].encode('latin-1'),
+)
+
+numpy_codec = JSONCodec(
+    cls=np.ndarray,
+    to_dict=lambda obj: {
+        'dtype': str(obj.dtype),
+        'shape': list(obj.shape),
+        'bytes': obj.tobytes()
+    },
+    from_dict=lambda dct: np.frombuffer(
+        dct['bytes'], dtype=np.dtype(dct['dtype'])
+    ).reshape(dct['shape'])
+)
+
+
+class TestJSONSerializerDeserializer(object):
+    def test_add_codec(self):
+        # without bytes codec, can't serialize numpy
+        serialization = JSONSerializerDeserializer([numpy_codec])
+        obj = np.array([[1.0, 0.0], [2.0, 3.2]])
+        with pytest.raises(TypeError):
+            serialization.serializer(obj)
+        # add the codec and it will work
+        serialization.add_codec(bytes_codec)
+        serialized = serialization.serializer(obj)
+        assert len(serialization.codecs) == 2
+        reconstructed = serialization.deserializer(serialized)
+        npt.assert_equal(obj, reconstructed)
+
+
+class CustomJSONCodingTest(object):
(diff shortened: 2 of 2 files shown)


dwhswenson:
Last 2 commits show usage:

* To see how you create a codec, see what I did for `pathlib.Path` in https://github.com/OpenFreeEnergy/gufe/pull/68/commits/4970f171b67122a4f74e2f60c693ba8beca30e31
* To see how our global `JSON_HANDLER` works (both in terms of registering default codecs and using this to serialize things) see https://github.com/OpenFreeEnergy/gufe/pull/68/commits/18efa987694084c3b44cd38019231ee493bcbe0e

An external tool that wants to register a handler can do so with:

```python
JSON_HANDLER.register(my_codec)
```

This should be ready for review now. I haven't added OpenFF unit support yet, but that should be pretty easy.

IAlibay:
So given it's used by multiple projects, would it not make more sense to spin this off as its own small package? Like it's not very big so I could understand the need to vendor it, but it would make maintenance easier?

dwhswenson:
> So given it's used by multiple projects, would it not make more sense to spin this off as its own small package? Like it's not very big so I could understand the need to vendor it, but it would make maintenance easier?

There's some extra functionality in the SimStore version, because it has some needs that we don't (explicit distinction of simulation and data objects; with different serialization paths for each). But eventually splitting it off is definitely a possibility.

gufe/tests/test_custom_json.py
```diff
@@ -0,0 +1,145 @@
...
+    cls=np.ndarray,
+    to_dict=lambda obj: {
+        'dtype': str(obj.dtype),
+        'shape': list(obj.shape),
+        'bytes': obj.tobytes()
+    },
+    from_dict=lambda dct: np.frombuffer(
+        dct['bytes'], dtype=np.dtype(dct['dtype'])
+    ).reshape(dct['shape'])
+)
```
richardjgowers, line 32: This is a good example, but we also need these built in as "batteries included" components, along with openff.units.  Can we merge these into the actual library and keep the existing tests (that require them built in), then separately have a dumb class we define in this test file to show the extension of Serializer.

gufe/tests/test_custom_json.py
```diff
@@ -0,0 +1,145 @@
...
+        # test that the default behavior is obeyed
+        obj = {'test': 5}
+        json_str = '{"test": 5}'
+        encoder, decoder = custom_json_factory([self.codec])
+        assert json.dumps(obj, cls=encoder) == json_str
+        assert json.loads(json_str, cls=decoder) == obj
+
+
+
+class TestNumpyCoding(CustomJSONCodingTest):
```
richardjgowers, line 96: I'd also like to see a test where we store (hardcode) a reference numpy serialization (rather than dynamically generate it) so we test that we're not inadvertently changing the format as we go.  This could be as simple as a bytes variable as a fixture, which we deserialize in a test.

gufe/tests/test_custom_json.py
richardjgowers: & Pathlib.path

richardjgowers (changes requested):
A question I still have regarding usage, if I have a protocol which has a custom object, do I define/register the codec *before* attempting to deserialize the object?  Or is it enough that the codec is registered in the `Protocol.from_dict` method?

dwhswenson:
> A question I still have regarding usage, if I have a protocol which has a custom object, do I define/register the codec before attempting to deserialize the object? Or is it enough that the codec is registered in the Protocol.from_dict method?

It needs to be registered before deserialization, because we need to know how to recognize that object. So one option (though I don't love it) is to register on import of an external package.

Actually, I suppose I could add a reference to the codec in the dict that gets saved. In that case, we could autoregister (as long as the package was installed). This would involve some fancy tricks; I'd rather not try to implement it in this PR, but it is a thought for later (basically, the dict it returns can return a reference back the codec itself -- this is a little messy because the codec is not a class, but an instance, so knowing its name is tricky and probably requires something like looping over the contents of the module, and finding the name(s) that point to the codec itself -- documenting this here in case I decide to come back to this idea).

gufe/tests/test_custom_json.py
dwhswenson: Sure, I'll move them over. I didn't know if we wanted to handle numpy objects; so I only put them in tests.

Do we really want a bytes codec? Or should I make it so that the numpy codec just returns string directly?

It's kind of a question of how much non-JSON standard do we want to enable storage of: tuples? sets? dicts with non-string keys? All of this is possible; just a questions of what is worth the time to support (OPS supports tuples and sets, so implementation exists if we want them).

dwhswenson:
> @dwhswenson so re: loading codecs, would it be enough to write this:

No. You're making this harder on the module contributor than it needs to be. All they need to do is define and register the codec. The rest is handled by storage.

Here are a few things wrong with that code snippet:

1. This approach breaks composability. Instead of `s = JSONSerializerDeserializer([custom_codec_X])` you need `s = JSONSerializerDeserializer([custom_codec_X] + all_other_known_codecs)`, at least if any other codecs are used internally. Finding `all_other_known_codecs` is going to be problematic. That's why we use a centralized registry. 

    The better way here might be `JSON_HANDLER.register(custom_codec_X)`, where JSON_HANDLER is a an application-wide global instance of `JSONSerializerDeserializer`. I think it makes more sense to do that at import. If we really wanted to, we could automate that with a metaclass, although that would break the ability for `customjson` to be universally used between different applications. (Codecs for OPS may be handle types differently from codecs for OpenFE.)

    If you want each class to fully know how to (de)serialize itself (not rely on composition of codecs) then don't do this at all and just use the standard JSON machinery. That can work, but might mean that different people serialize the same objects differently, and requires a lot more boilerplate code from the contributors. My approach was intended to reduce the burden on contributors.

2. (Minor). Not `def custom_codec_X()`. It isn't a function. It is an instance of `JSONCodec`: `custom_codec_X = JSONCodec(...)`. The initialization parameters of `JSONCodec` are the class to serialize (usually) and `to_dict`/`from_dict` functions.3. 

3. (Also minor). The argument in `_from_dict` for a `GufeTokenizable` is a `dict`, not a `string`.

I think the code we want looks like this:

```python
# contributor module; first 3 lines are different if you have custom codecs
custom_codec_X = JSONCodec(...)
JSON_HANDLER.register(custom_codec_X)


# everything in here is independent of whether you use custom json
class Something(GufeTokenizable):
    @classmethod
    def _from_dict(cls, dct):
        return cls(**dct)  # the same as a gufe tokenizable that doesn't use custom JSON

    def _to_dict(self):
        return self.__dict__   # or whatever; the same as a gufe tokenizable that doesn't use custom JSON
```

It actually gets used with storage as a consumer or dicts. This is on us; the contributor never needs to know this.

```python
# in whatever file actually writes things to JSON (something in storage; possibly only in OpenFE?)
# this is on us, users/contributors never need to think about this
# this currently exists in resultclient.py in #42 

class ResultClient(_ResultContainer):
    ... # lots of stuff
    def _store_gufe_tokenizable(self, prefix, obj):
        for o in get_all_gufe_objs(obj):
        ... # bunch of stuff
        # thing to see: here we use JSON_HANDLER.encoder (not implemented in #42 yet)
        data = json.dumps(o.to_keyed_dict(), sort_keys=True, cls=JSON_HANDLER.encoder).encode('uff-8')
        ... # etc

    def _load_gufe_tokenizable(self, prefix, gufe_key):
        ... # similarly, a bunch of stuff until we get to the JSON
        dct = json.loads(keyencoded_json, cls=JSON_HANDLER.decoder)
        ... # etc
```

What the contributor is really doing here is giving guidance on how to serialize to JSON (which we expect to be a common format). However, there's nothing we do in gufe that guarantees that JSON serialization is actually used. I don't know if fah-alchemy is passes its objects through JSON at all, or whether it is just directly serializing them (handling custom types in some other way). For OpenFE, we'll want to serialize to JSON; having this in gufe encourages standardization across any tools that *do* use JSON. But custom type handling can be done differently for different packages.

dwhswenson:
One point I'm stuck on here: What do we want in the first batch of default codecs? `pathlib.Path`, certainly; do we want `numpy.ndarray`, and if so, do we want a separate `bytes` codec as well, or to just directly dump ndarray to a string form?

richardjgowers:
@dwhswenson bytes seems very useful, also openff.units.Quantity

richardjgowers:
and the pattern of Codecs as classes and registering Codecs looks good to me, make it so etc

dwhswenson:
Assuming we follow the 4-step proposal I gave above, this is ready for review and merge; with a follow-up PR planned for openff units. Take a look, @richardjgowers?

gufe/custom_json.py
```diff
@@ -0,0 +1,162 @@
...
+    def serializer(self, obj):
+        """Callable that dumps to JSON"""
+        return self._serializer(obj)
+
+    def deserializer(self, string):
+        """Callable to loads JSON"""
+        return self._deserializer(string)
+
+
+def custom_json_factory(coding_methods):
```
richardjgowers, line 69: I'd be tempted to make this private and keep the API surface smaller (or document and test it properly)

gufe/custom_json.py
```diff
@@ -0,0 +1,162 @@
+# This file was originally part of OpenPathSampling/SimStore, which is
+# distributed under the MIT license.
+# Portions Copyright (c) 2014-2022 the contributors to OpenPathSampling
+# Permissions are the same as those listed in the gufe LICENSE
+
+import json
+import functools
+
+
+class JSONSerializerDeserializer(object):
```
richardjgowers, line 10: type annotations on this would help a lot

richardjgowers (changes requested):
I'm not seeing that the default codecs work without having to explicitly include them?  I would consider defaults to work without thinking about it.

richardjgowers:
@dwhswenson this doesn't work, imho it should
```python
class Thing(gufe.tokenization.GufeTokenizable):
    def __init__(self, v):
        self.vals = v

    def _defaults(self):
        return {}

    def _to_dict(self):
        return {'vals': self.vals}

    @classmethod
    def _from_dict(cls, dct):
        return cls(**dct)


def test_thing():
    t = Thing(np.arange(5))

    dd = t.to_dict()

    t2 = Thing.from_dict(dd)

    assert t == t2
```

dwhswenson:
> this doesn't work, imho it should

This has nothing to do with the code here. This will never touch the custom JSON stuff. `to_dict` never sees JSON, just a dictionary.

This is on gufe.GufeTokenizable and specifically on the types supported by `normalize`.

richardjgowers:
Ok there's a difference between json and dicts.  But I can't even currently create a gufe tokenizable that holds a numpy array, so it'll never be able to use this machinery later.  So it sounds like normalize needs to get fixed else anything downstream can't get into play.

dwhswenson:
I'll redo `normalize` in a different PR. It's a distinct problem. But I'd like to get feedback from @dotsdl on it, since that's the approach he wanted to use.

richardjgowers (approved):
ok this LGTM, but I'll raise an issue about the blocker to actually using this

richardjgowers:
> Only question RE: last point -- when do we want to ensure serialization stability? Namely, do we need to plan for backward compatibility here?

Ideally since numpy arrays etc have been stable for a while, this should be very backwards compatible.  Then only "unstable" thing I can see we're using is openff-units, which at least we're well integrated with.

dwhswenson:
> Then only "unstable" thing I can see we're using is openff-units, which at least we're well integrated with.

This is what I was referring to, in the context that the serialization may change to be standardized to be more like whatever that openff-models does internally.

gufe/custom_json.py
dwhswenson: Added in [c7abb1c](https://github.com/OpenFreeEnergy/gufe/pull/68/commits/c7abb1c7c17abb5ae5e8baa67ec3124300377e66), although I'm not sure how much they actually help....

gufe/custom_json.py
dwhswenson: It's tested in anything that inherits from `CustomJSONTest`, in the `test_round_trip` method.

If, by documentation, you mean on the dynamically-defined classes, I disagree. I don't want to override the superclass docs, which you get now if you do

```python
help(JSON_HANDLER.encoder)
```

The only possible reason to override the super's docs is that in decoder, technically you'll get an error if you specify `object_hook` (since we're already specifying it).

richardjgowers: Re docs: This function isn't well documented, it took me a while to get what was going on.

dwhswenson: Do you mean that it isn't clear that it is a factory function returning dynamically created classes? Or do you mean that mechanism of the `default` / `object_hook` methods are unclear?

dwhswenson: Added more docs on this method in [e320f93](https://github.com/OpenFreeEnergy/gufe/pull/68/commits/e320f933332f0732af998cbe71f0aac567affaab). Honestly not sure what else to do if this doesn't explain it well enough.

PR #69, opened 2022-09-29
Add test for roundtrip with cleared registry

dwhswenson:
Main goal here is to add a test to `GufeTokenizableTestsMixin` that will automatically test that the full `to_dict`/`from_dict` cycle works if the `TOKENIZABLE_REGISTRY` has been cleared, to mock deserializing in a different process from the serialization.

This led to some weirdness:

* I got the same problem with `WeakValueDictionary` that I mentioned this morning; where I get a `KeyError`, but if you drop into the pdb on error, the key is in the dict. I fixed this by putting a `try`/`except` around that.
* There is a genuine error in serialization of `SmallMoleculeComponent`. It's because we serialize with explicit hydrogens, but don't force explicit hydrogens when we take in an rdkit molecule. So the original ethane has smiles `CC` and the deserialized one has `[H]C([H])([H])C([H])([H])[H]` (@richardjgowers is going to love this one!). I think the options are either:
  * `addHydrogens` on taking in the molecule
  * `removeHydrogens` before calculating the SMILES string for hashing

Tasks:

- [x] Draft test
- [x] Fix problem with WeakValueDictionary
- [x] Fix problem with `SmallMoleculeComponent`

### DIFF
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -68,16 +68,7 @@ class ExplicitMoleculeComponent(Component):
             )
 
         self._rdkit = rdkit
-        self._hash = self._hashmol(name=name)
-
-    def __hash__(self):
-        return hash(self._hash)
-
-    def __eq__(self, other):
-        return hash(self) == hash(other)
-
-    def _hashmol(self, name):
-        return hashmol(self._rdkit, name=name)
+        self._hash = hashmol(self._rdkit, name=name)
 
     def _defaults(self):
         return super()._defaults()
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -4,7 +4,6 @@ from rdkit import Chem
 
 from .component import Component
 from .. import __version__
-from ..molhashing import hashmol
 
 # typing
 from ..custom_typing import RDKitMol
@@ -48,9 +47,6 @@ class ExplicitMoleculeComponent(Component):
     This provides basic serialization and conversion to different
     representations. Specific file formats, such as SDF for small molecules
     or PDB for proteins, should be implemented in subclasses.
-
-    We default to use SMILES as the basis of the hash, but this can be
-    overridden using the `_hashmol` method.
     """
 
     def __init__(self, rdkit: RDKitMol, name: str = ""):
@@ -68,7 +64,8 @@ class ExplicitMoleculeComponent(Component):
             )
 
         self._rdkit = rdkit
-        self._hash = hashmol(self._rdkit, name=name)
+        self._smiles = None
+        self._name = name
 
     def _defaults(self):
         return super()._defaults()
@@ -76,11 +73,14 @@ class ExplicitMoleculeComponent(Component):
     # Props
     @property
     def name(self) -> str:
-        return self._hash.name
+        return self._name
 
     @property
     def smiles(self) -> str:
-        return self._hash.smiles
+        if self._smiles is None:
+            self._smiles = Chem.MolToSmiles(self._rdkit)
+
+        return self._smiles
 
     @property
     def total_charge(self):
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -1,6 +1,7 @@
 import json
 import warnings
 from rdkit import Chem
+from typing import Optional
 
 from .component import Component
 from .. import __version__
@@ -64,7 +65,7 @@ class ExplicitMoleculeComponent(Component):
             )
 
         self._rdkit = rdkit
-        self._smiles = None
+        self._smiles: Optional[str] = None
         self._name = name
 
     def _defaults(self):
@@ -78,7 +79,7 @@ class ExplicitMoleculeComponent(Component):
     @property
     def smiles(self) -> str:
         if self._smiles is None:
-            self._smiles = Chem.MolToSmiles(self._rdkit)
+            self._smiles = Chem.MolToSmiles(Chem.RemoveHs(self._rdkit))
 
         return self._smiles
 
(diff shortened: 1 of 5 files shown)


mikemhenry:
> `removeHydrogens` before calculating the SMILES string for hashing

Won't this bite us when we are working with different protonation states of the same ligand?

richardjgowers:
Aren't we falling back to using gufe hash now? So never smiles anyway?

dwhswenson:
`SmallMoleculeComponent` still had `__eq__` and `__hash__` based on `molhashing.hashmol`. That's probably leftover code that should have been removed -- that should solve this.

dwhswenson:
> Aren't we falling back to using gufe hash now? So never smiles anyway?

That fixes the equality check, but we still have this weirdness:


```python
In [1]: from rdkit import Chem
   ...: from rdkit.Chem import AllChem
   ...: from gufe import SmallMoleculeComponent
   ...: from gufe.tokenization import TOKENIZABLE_REGISTRY

In [2]: m = Chem.MolFromSmiles("CC")
   ...: AllChem.Compute2DCoords(m)
   ...: smc = SmallMoleculeComponent.from_rdkit(m, name="Billy")

In [3]: dct = smc.to_dict()

In [4]: TOKENIZABLE_REGISTRY.clear()

In [5]: copy = SmallMoleculeComponent.from_dict(dct)

In [6]: smc == copy  # fixed; as expected
Out[6]: True

In [7]: smc.smiles == copy.smiles  # not fixed
Out[7]: False
```

richardjgowers:
To answer the original question, we definitely want explicit Hs around else
we might drop positions?

dwhswenson:
> To answer the original question, we definitely want explicit Hs around else we might drop positions?

I'm not sure what you mean here. I'm only asking whether the SMILES string from `component.smiles` is the implicit or explicit form.

Right now (after [ce79601](https://github.com/OpenFreeEnergy/gufe/pull/69/commits/ce79601041cf85af5d14465b1de2bd0840f8dea3)), I think the only purpose of the SMILES string is that the user can request `component.smiles`. We don't use it internally for anything. Equality and hashing are based on gufe key. So do we want it to be the SMILES string with explicit or implicit hydrogens?

The options I'm suggesting here are:

1. Add `AddHs` in `ExplicitMoleculeComponent.__init__`
2. Add `RemoveHs` in `molhashing.hashmol`

The only other place we use SMILES is in the `SolventComponent`, where it defines the solvent molecule type (and we use the implicit form for the default).

dwhswenson:
Version 1: Add Hs (but only if there are no Hs in the molecule, because it was adding hydrogens to proteins as well -- maybe to balance negatively charged residues? In any case, the number of atoms didn't stay constant.) Problem: What if you have a mix of implicit and explicit H? We claim we can take an RDKit molecule, so we should be able to handle any RDKit molecule.

Version 2: Remove H when calculating the SMILES string.


```diff
diff --git a/gufe/components/explicitmoleculecomponent.py b/gufe/components/explicitmoleculecomponent.py
index 5021509..9a0b4db 100644
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -54,6 +54,9 @@ class ExplicitMoleculeComponent(Component):
     """
 
     def __init__(self, rdkit: RDKitMol, name: str = ""):
+        if not any(a.GetAtomicNum() == 1 for a in rdkit.GetAtoms()):
+            rdkit = Chem.AddHs(rdkit)
+
         name = _ensure_ofe_name(rdkit, name)
         _ensure_ofe_version(rdkit)
         conformers = list(rdkit.GetConformers())
```

```diff
diff --git a/gufe/molhashing.py b/gufe/molhashing.py
index 9b56353..8cf14e5 100644
--- a/gufe/molhashing.py
+++ b/gufe/molhashing.py
@@ -13,7 +13,7 @@ class MoleculeHash(NamedTuple):
 
 def hashmol(mol, name=""):
     # MolToSMILES are canonical by default
-    return MoleculeHash(Chem.MolToSmiles(mol), name)
+    return MoleculeHash(Chem.MolToSmiles(Chem.RemoveHs(mol)), name)
 
 
 def serialize_numpy(arr) -> str:
```

gufe/components/explicitmoleculecomponent.py
```diff
@@ -68,16 +68,7 @@ def __init__(self, rdkit: RDKitMol, name: str = ""):
...
-
-    def __hash__(self):
-        return hash(self._hash)
-
-    def __eq__(self, other):
-        return hash(self) == hash(other)
-
-    def _hashmol(self, name):
-        return hashmol(self._rdkit, name=name)
+        self._hash = hashmol(self._rdkit, name=name)
```
richardjgowers, line 71: For this to be used by the gufe machinery, shouldn't it be called _key? Either way, we can probably remove the hashmol concept completely and replace with the hash of the to_dict string.

dwhswenson: We're not using this for hashing. At all. I switched it to gufe `to_dict`-based things in [ce79601](https://github.com/OpenFreeEnergy/gufe/pull/69/commits/ce79601041cf85af5d14465b1de2bd0840f8dea3).

I didn't remove/rename `hashmol` just because I was trying to keep the PR small. I'll do that now.

richardjgowers:
Ok sorry I get it now, so the smiles aren't used in the hash. I think no H is the more intuitive/expected representation here.

dwhswenson:
> Ok sorry I get it now, so the smiles aren't used in the hash. I think no H is the more intuitive/expected representation here.

Yeah, I was a little confused that you seemed to be arguing the other direction. I'll make that switch; I think it'll break a test in the CLI (we tested for a case with explicit hydrogens) but I think it'll make more sense.

dwhswenson:
Okay, now I'm happy with this (assuming it passes tests). @richardjgowers, take another look over it?

richardjgowers (approved):

PR #75, opened 2022-10-03
Add __init__.py to vendor/

dwhswenson:
Getting errors on OpenFE PRs with `ModuleNotFoundError: No module named 'gufe.vendor'` -- hoping this is all that is needed to fix.


richardjgowers (approved):
Shouldn't be necessary but sure

PR #76, opened 2022-10-04
Performance improvements to tokenization, deterministic sorting

dotsdl:
We now do the following:
- tokenization via `_gufe_tokenize` is done of keyed-dict form, avoiding
  a long chain of serialization for e.g. `AlchemicalNetwork`s featuring
  `ProteinComponent`s
- `GufeTokenizable.__lt__` is done on `self.key` value, not on the results
  from `hash(self)`, which does not deliver the same result across
  Python processes
- explicit sorting in key places for `ChemicalSystem`,
  `AlchemicalNetwork`, to ensure `.key` stability

### DIFF
--- a/gufe/chemicalsystem.py
+++ b/gufe/chemicalsystem.py
@@ -68,9 +68,6 @@ class ChemicalSystem(abc.Mapping, GufeTokenizable):
             f"{self.__class__.__name__}(name={self.name}, components={self.components})"
         )
 
-    def __lt__(self, other):
-        return hash(self) < hash(other)
-
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
             return False
@@ -96,7 +93,7 @@ class ChemicalSystem(abc.Mapping, GufeTokenizable):
     def _to_dict(self):
         return {
             "components": {
-                key: value for key, value in self.components.items()
+                key: value for key, value in sorted(self.components.items())
             },
             "box_vectors": self.box_vectors.tolist(),
             "name": self.name,
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from .component import Component
 from .. import __version__
+from ..tokenization import module_qualname
 
 # typing
 from ..custom_typing import RDKitMol
@@ -114,3 +115,14 @@ class ExplicitMoleculeComponent(Component):
 
     def _to_dict(self) -> dict:
         raise NotImplementedError()
+
+    def to_dict(self, include_defaults=True) -> dict:
+        dct = self._to_dict()
+        dct.update(module_qualname(self))
+
+        if not include_defaults:
+            for key, value in self.defaults.items():
+                if dct.get(key) == value:
+                    dct.pop(key)
+
+        return dct
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -117,6 +117,21 @@ class ExplicitMoleculeComponent(Component):
         raise NotImplementedError()
 
     def to_dict(self, include_defaults=True) -> dict:
+        """Generate full dict representation, with all referenced
+        `GufeTokenizable` objects also given in full dict representations.
+
+        Parameters
+        ----------
+        include_defaults : bool
+            If `False`, strip keys from dict representation with values equal
+            to those in `defaults`.
+
+        See also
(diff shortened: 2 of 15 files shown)


dwhswenson:
Obvious performance gains to be made: `Component`s do not contain gufe objects, but their dict form may contain many dicts/iterables. The default `to_dict` uses `dict_encode_dependencies`, so recursively looks for gufe tokenizables in all containers. For objects that can not contain gufe tokenizables, we should override:

```python
def to_dict(self, include_defaults=True):
    dct = self._to_dict()
    if not include_defaults:
         ...  # as exists currently
```

dwhswenson:
> Will add tests for this; a bit tricky since it requires spinning up separate Python processes to validate that keys don't change across processes.

Can't you just hard-code a key? It shouldn't change across versions either. (Or if it does, we should be alerted with a failing test.)

The implementation so far looks good to me. Should I add the overrides for `to_dict` for components here or in a separate PR?

dotsdl:
@dwhswenson feel free to add any `to_dict` optimizations here!

dotsdl:
Added stability checks in d2a7f02 for all `GufeTokenizable`s that we expect should have stable keys.

@dwhswenson after you add the optimizations you have in mind for `Component` subclasses, we should be good to go on this PR.

richardjgowers (approved):
Lgtm

dwhswenson:
Anything that contains a `ProteinComponent` is failing with these keys (and locally I'm getting the same keys as tests). @dotsdl : I can replace these with the keys that I find, but you should be aware that something is different in your environment that is breaking the transferability of these keys.

I've added the faster `to_dict` implementations. I added them explicitly in `SolventComponent` and `ExplicitMoleculeComponent` rather than the generic component superclass because I figure it is possible that someone in the future will come up with a component that *does* contain other gufe tokenizables.

dotsdl:
@dwhswenson that's concerning! I ran my tests on `mskcc2`, with `openff-toolkit` 0.11.0, `rdkit` 2022.03.05. Will try to isolate how this is happening.

dotsdl:
I'm pulling my hair out trying to pin down unstable `.key`s for `AlchemicalNetwork`, `Transformation`, `NonTransformation` and `ProteinComponent` examples in our tests. My suspicion is that something about `ProteinComponent` tokenization is nondeterministic and that's causing the other failures on up the tree. 

What's *really* weird is I can get the tests to all pass locally, but after committing or pushing the files to my test host the tests fail again. It's spooky really, and I don't understand it.

dotsdl:
@dwhswenson are there any aspects of the `ProteinComponent._to_dict` that may be nondeterministic? Or perhaps how the `ProteinComponent.from_pdb_file` pulls in its state in the first place?

dwhswenson:
> @dwhswenson are there any aspects of the `ProteinComponent._to_dict` that may be nondeterministic? Or perhaps how the `ProteinComponent.from_pdb_file` pulls in its state in the first place?

I haven't investigated in too much depth, but relevant info is that I match CI's Ubuntu keys on macOS. It can't be completely random, and can't be Linux-specific.

dotsdl:
So even with sorting the call to `self._rdkit.GetPropsAsDict` in `ProteinComponent._to_dict`, we're still getting the shell game behavior I demonstrated earlier today. Fixing the keys based on what the tests give works initially, but after committing the changes the tests are immediately broken again. It's the strangest thing to observe.

Our working hypothesis is that some things in `ProteinComponent._to_dict` are not deterministic. Since we are pulling things out of RDKit, are we guaranteed to get the same ordering of atoms, bonds, and conformers? Should we be sorting any of these instead?

dwhswenson:
Another suggestion: instead of tokenizing with the string rep, use json.dumps with sorted keys. It occurred to me on the flight that we wouldn't even have these dictionary order questions if we did that (and that's what I've always done before in this situation). 

You can also compare those two string reps to see exactly what differs. (Do an assert test against that string, which you can dump to a file. pytest will give you the diff.) Might be easier than diving through details of rdkit code.

dotsdl:
@richardjgowers can you have a look at the `_to_dict` implementation to see if there are any non-deterministic list generations in there?

dotsdl:
FYI @dwhswenson: in efd7172 I removed optimizations for `ExplicitMoleculeComponent` and `SolventComponent` `to_dict`; they don't appear to really make a difference in my tokenization timings, and I'd rather avoid the additional complexity of having the overrides given this. We can re-add later if they prove vital for larger systems, though.

PR #78, opened 2022-10-16
Remove `old-offtk` from CI env

dotsdl:
I've also set old toolkit CI checks as no longer required on the repo itself in `main` branch protections. Is there a reason for maintaining any pre `openff-toolkit` 0.11.0 compatibility at this point?


richardjgowers (changes requested):
We're still checking that off tk 11 didn't break anything (scientifically) between versions. Plus once we do remove support we can likely remove some code paths, rather than just the CI matrix entry

IAlibay:
> We're still checking that off tk 11 didn't break anything (scientifically) between versions. Plus once we do remove support we can likely remove some code paths, rather than just the CI matrix entry

Are you alluding to [OpenFE #142](https://github.com/OpenFreeEnergy/openfe/pull/142) here? If so, I wonder if it would make sense to just assume this is a breaking change, validate #142 as OpenFF 0.11+ and then if there's a big problem we can revert the two?

richardjgowers:
I'm gonna bump this to the next release, so this release is both 10 and 11 compatible

richardjgowers (approved):

PR #89, opened 2022-11-02
Add class-level loggers to GufeTokenizables

dwhswenson:
This was discussed at the OpenFE meeting today. We decided we need more logging. To have those be easily filtered, and to make them consistent in naming, I recommended class-level loggers that we define in `GufeTokenizable`.

This now gives you a default logger at `self.logger` for any tokenizable. It's logger name is the module + qualname.


dwhswenson:
This is ready for review.

gufe/tokenization.py
```diff
@@ -77,6 +78,16 @@ def _gufe_tokenize(self):
...
         return normalize(self.to_keyed_dict(include_defaults=False))
 
+    @property
+    def logger(self):
+        if (logger := getattr(self, '_logger', None)) is None:
+            cls = self.__class__
+            logname = cls.__module__ + "." + cls.__qualname__
+            logger = logging.getLogger(logname)
+            self._logger = logger
+        return logger
```
richardjgowers, line 88: stuff like this needs documenting

richardjgowers (changes requested):
docs

gufe/tokenization.py
dwhswenson: done; take another look

gufe/tokenization.py
```diff
@@ -77,6 +78,17 @@ def _gufe_tokenize(self):
         """
         return normalize(self.to_keyed_dict(include_defaults=False))
 
+    @property
+    def logger(self):
+        """Return logger named by this module and class qualname"""
+        if (logger := getattr(self, '_logger', None)) is None:
+            cls = self.__class__
+            logname = cls.__module__ + "." + cls.__qualname__
```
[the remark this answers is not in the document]
dwhswenson: Per-instance logging instead of per-class? It's not impossible, but it would mean a lot more loggers. I'm not sure how much overhead that would add.

Note that ["Multiple calls to getLogger() with the same name will always return a reference to the same Logger object."](https://docs.python.org/3/library/logging.html#logger-objects) So in the current implementation we do not create one logger instance per instance, but one logger instance per class and all members of the class have  pointer to it.

(`@classmethod` and `@property` don't play well together, otherwise I would have implemented this as a class-property. I decided keeping it a property is more important than making it clear that this is a class property -- it's already documented in the docs for getLogger.)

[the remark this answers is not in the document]
richardjgowers: Its more looking at the test cases, once we've got more than one Unit running we'll want to be able to easily go from logs to work unit. Maybe this can be solved through storing one set of logs per Unit though, so the file name is the Key.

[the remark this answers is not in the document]
dwhswenson: FWIW, having this here in the superclass makes it easy to change that in the future if we want. Maybe we go with this middle road now, and can check the performance impact of adding more fine details later? (We can also drop down to module-only if we find that per-class is an issue.)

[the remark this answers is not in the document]
dwhswenson: > Its more looking at the test cases, once we've got more than one Unit running we'll want to be able to easily go from logs to work unit. Maybe this can be solved through storing one set of logs per Unit though, so the file name is the Key.

Or by defining `ProtocolUnit.logger` and overriding this.

gufe/tests/test_tokenization.py
```diff
@@ -211,6 +215,30 @@ def test_notequal_different_type(self):
 
         assert l1 != l2
 
+    @pytest.mark.parametrize('level', ["DEBUG", "INFO", "CRITICAL"])
+    def test_logging(self, level):
+        stream = io.StringIO()
+        handler = logging.StreamHandler(stream)
+        fmt = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
```
dotsdl, line 222: Can we go with pipes as the delimiter here? Less commonly found in messages, so makes raw log parsing simpler downstream.

Also, can we put the `levelname` first? Might be able to set padding as well so the result looks something like:

```
DEBUG    | gufe.transformations.Transformation | eb839ae80d905adee778c32f516944ad | I'm a silly goose.
INFO     | gufe.chemicalsystem.ChemicalSystem | e1cb9ce41e88ee474cf5b962c9388159 | No, I am in fact, the silly goose.
CRITICAL | gufe.network.AlchemicalSystem       | 54aa4f41f09cb07ebe6febdc65f1c278 | I AM THE SILLIEST GOOSE.

```

Here we have `levelname | name | gufe-key token | message`.

We may also want a timestamp out front, but this may only be desirable for non-STDOUT handlers.

dotsdl (approved):
Looking great @dwhswenson! Just one proposed change, but I'm going to give my approval now.

gufe/tests/test_tokenization.py
dwhswenson: This is a format that's only used in this specific test. End user can set any desired format in logging configuration.

dotsdl: Ah good point! Agree that this is best handled at the application layer.

richardjgowers: @dwhswenson what would be the format string to pull the gufe key out of an object?  (i.e. what formatter do I set to get Dotson's example above?)

richardjgowers: Hmm, I think that's why I thought of having the gufekey be in the name of the logger (one per instance), but then equally it becomes impossible(?) to address the logger via `getLogger()` as it has a random name.

dwhswenson: No, you could have the logger name include the gufe key. That's not a problem. But then you have a lot more loggers in the logger registry. This could make logging slower; I'm not sure how the backend for that works, but I know the typical usage is 1 logger per module or, at most, 1 per class, so it wouldn't surprise me if 1 per instance is a Really Bad Idea™.

dwhswenson:
Interesting edge case I came across here: If you want gufe keys built into logging, this is impossible:

```python
class Leaf(GufeTokenizable):
    def __init__(self, a, b=2):
        self.logger.info(f"{a=}")
        self.logger.debug(f"{b=}")
        # logging before setting self.a / self.b means that can't calculate the key,
        # so the above lines give an error while trying to set the key
        self.a = a
        self.b = b
```

Solution is that you can only use `self.logger` after everything needed for the key has been set.

gufe/tests/test_tokenization.py
dwhswenson: We can now inject the gufe key. Turns out I didn't need to write my own class to handle the injection, because there's `logging.LoggerAdapter` for exactly this purpose.

dwhswenson:
... and fixed the problem mentioned above. Now if you log before the gufe key can be calculated, it will report key as `UNKNOWN`.

gufe/tokenization.py
```diff
@@ -50,6 +51,31 @@ class _ABCGufeClassMeta(_GufeTokenizableMeta, abc.ABCMeta):
...
+            extra_dict = {
+                'gufekey': gufekey
+            }
+
+            if save_extra_dict:
+                self._extra_dict = extra_dict
+
+        extra.update(extra_dict)
+        kwargs['extra'] = extra
+        return msg, kwargs
```
richardjgowers, line 76: document this

richardjgowers (changes requested):
docs

richardjgowers (approved):

PR #90, opened 2022-11-04
[WIP] Some cleanup of GufeTokenizable

dwhswenson:
This is intended to clean up a few edges of `GufeTokenizable`. Here are some of the discrete changes:

- [x] Use JSON to tokenize — this will allow us to use our custom JSON objects
- [x] Add `copy_with_replacement` method
- [ ] Tests for `copy_with_replacement`
- [ ] Move `defaults` to a classmethod
- [ ] Switch to recreating the token, not the object, when reloading

I'd like to have #62 merged before this, as I think it might be help with some of the problems I'm having early on.

### DIFF
--- a/gufe/tokenization.py
+++ b/gufe/tokenization.py
@@ -10,6 +10,7 @@ import importlib
 from pathlib import Path
 import inspect
 import copy
+import json
 from typing import Dict, Any, Callable, Union, List, Tuple
 import weakref
 from gufe.custom_json import JSONSerializerDeserializer
@@ -75,7 +76,8 @@ class GufeTokenizable(abc.ABC, metaclass=_ABCGufeClassMeta):
         """Return a list of normalized inputs for `gufe.base.tokenize`.
 
         """
-        return normalize(self.to_keyed_dict(include_defaults=False))
+        return tokenize(self)
+        # return normalize(self.to_keyed_dict(include_defaults=False))
 
     @property
     def key(self):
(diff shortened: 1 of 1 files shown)


dwhswenson:
This is ready for review. CodeCov patch is complaining about a single like where I changed `default` -> `default()`, since it is no longer a property -- I don't think it lost coverage because of me; it's just that I touched a line that I suspect already wasn't covered.

gufe/tokenization.py
```diff
@@ -309,6 +312,15 @@ def from_shallow_dict(cls, dct: Dict):
...
         return from_dict(dct)
 
+    def copy_with_replacements(self, **replacements):
+        dct = self._to_dict()
+        if invalid := set(replacements) - set(dct):
+            raise TypeError(f"Invalid replacement keys: {invalid}. "
+                            f"Allowed keys are: {set(dct)}")
+
+        dct.update(replacements)
+        return self._from_dict(dct)
```
richardjgowers, line 322: document this

richardjgowers (changes requested):
docs

gufe/tokenization.py
dwhswenson: Done in https://github.com/OpenFreeEnergy/gufe/pull/90/commits/cfa6122bacab27325d8e503554c1031b3d5e4da8. Take another look -- this PR is going to be blocking proper use of settings, so it needs to get in. (This particular method was just something else I had already tossed in when I started working on the more important stuff, and probably should have been a separate PR.)

dwhswenson:
@dotsdl: Since this involves API-breaking changes, I'd like to at least be sure you've taken a look at this before it is merged. I don't think it requires a heavy code review from you (@richardjgowers has already reviewed the code), but please read the summary in the top post so you know what is changing and can warn if any of that will cause major issues for you.

dotsdl:
Thanks @dwhswenson; reviewing now!

dotsdl (commented):
Looks great @dwhswenson! Thank you for greatly simplifying our tokenization scheme with these changes!

dotsdl (approved):

richardjgowers (approved):

PR #91, opened 2022-11-04
[DOCS] Now build without warnings, and fail on warning on RTD

richardjgowers:


gufe/protocols/protocolunit.py
```diff
@@ -30,26 +31,24 @@ class Context:
...
-                    if isinstance(i, cls):
-                        deps.append(i)
-            elif isinstance(value, cls):
-                deps.append(value)
-        return deps
-
-
-class ProtocolUnitResultBase(GufeTokenizable, ProtocolUnitMixin):
-    def __init__(self, *, 
+def _list_dependencies(inputs, cls):
```
richardjgowers, line 34: @dwhswenson it might cause you grief in a merge conflict, but I got rid of this mixin since it didn't make sense to unite ProtocolUnit and Result with a common ancestor, so we can just have a function instead

dwhswenson: Talk to @dotsl on that -- he had a reason to do this, but I don't remember the argument.

dwhswenson: (And I accidentally notified the wrong person ... that was supposed to be @dotsdl (fixed now); sorry dotSL)

Issue #93, opened 2022-11-10
newer mypy giving new errors


richardjgowers:
updated to v0.990 and got some new errors that must have snuck in

Looks like it's this:

https://github.com/hauntsaninja/no_implicit_optional

PR #94, opened 2022-11-10
mypy: Fixing implicit None errors

richardjgowers:
fixes issue #93

we fortran now

### DIFF
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -374,14 +374,14 @@ class ProteinComponent(ExplicitMoleculeComponent):
 
         return openmm_pos
 
-    def to_pdb_file(self, out_path: Optional[Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]] = None) -> str:
+    def to_pdb_file(self, out_path: Union[str, bytes, PathLike[str], PathLike[bytes], io.TextIOBase]) -> str:
         """
         serialize protein to pdb file.
 
         Parameters
         ----------
-        out_path :  Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]
-            provide path or any string based stream (e.g. FileIO ) to the resulting file, by default None
+        out_path :  Union[str, bytes, PathLike[str], PathLike[bytes], io.TextIOBase]
+            provide path or any string based stream (e.g. FileIO ) to the resulting file
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


gufe/components/proteincomponent.py
```diff
@@ -374,7 +374,7 @@ def to_openmm_positions(self) -> omm_unit.Quantity:
 
         return openmm_pos
 
-    def to_pdb_file(self, out_path: Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase] = None) -> str:
+    def to_pdb_file(self, out_path: Optional[Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]] = None) -> str:
```
richardjgowers, line 377: I'm not sure these are correct as optional, from reading the function body I think we actually need something here as there isn't a default filename

gufe/network.py
```diff
@@ -98,7 +98,9 @@ def _defaults(self):
 
     def to_graphml(self) -> str:
         """ """
+        raise NotImplementedError
```
richardjgowers, line 101: @dwhswenson I thought we had this already, or was this a different Network object?

gufe/components/proteincomponent.py
dwhswenson, line 377:
```suggestion
    def to_pdb_file(self, out_path: Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]) -> str:
```

Not optional, then make it required?

gufe/components/proteincomponent.py
```diff
@@ -421,7 +421,7 @@ def to_pdb_file(self, out_path: Union[Union[str, bytes, PathLike[str], PathLike[
         return out_path
 
     def to_pdbx_file(
-        self, out_path: Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase] = None
+        self, out_path: Optional[Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]] = None
```
dwhswenson, line 424:
```suggestion
        self, out_path: Union[Union[str, bytes, PathLike[str], PathLike[bytes]], io.TextIOBase]
```

gufe/network.py
dwhswenson: `LigandNetwork`, not `AlchemicalNetwork`. We actually don't use `AlchemicalNetwork` yet.

dwhswenson (changes requested):
For things that aren't actually optional, just make them required?

Otherwise looks good.

PR #95, opened 2022-11-11
Added raise_error kwarg for protocoldag.execute, ProtocolUnit.execute

dotsdl:
This change raises the error from a given ProtocolUnit.execute call if `raise_error=True`. In `protocoldag.execute`, giving this same kwarg will cause the method to stop on hitting an exception in the ProtocolUnit it is executing. This makese debugging easier when necessary.

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -206,7 +206,9 @@ class ProtocolDAG(GufeTokenizable, DAGMixin):
 
 
 def execute(protocoldag: ProtocolDAG, *, 
-            shared: Optional[PathLike] = None) -> ProtocolDAGResult:
+            shared: Optional[PathLike] = None,
+            raise_error: bool = False,
+            ) -> ProtocolDAGResult:
     """Execute the full DAG in-serial, in process.
 
     This is intended for debug use for Protocol developers.
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -205,10 +205,10 @@ class ProtocolDAG(GufeTokenizable, DAGMixin):
         return cls(**dct)
 
 
-def execute(protocoldag: ProtocolDAG, *, 
-            shared: Optional[PathLike] = None,
-            raise_error: bool = False,
-            ) -> ProtocolDAGResult:
+def execute_DAG(protocoldag: ProtocolDAG, *,
+                shared: Optional[PathLike] = None,
+                raise_error: bool = True,
+                ) -> ProtocolDAGResult:
     """Execute the full DAG in-serial, in process.
 
     This is intended for debug use for Protocol developers.
@@ -223,7 +223,9 @@ def execute(protocoldag: ProtocolDAG, *,
        is removed after. Used by some `ProtocolUnit`s to pass file contents
        to dependent `ProtocolUnit`s.
     raise_error : bool
-        If True, raise an exception if a ProtocolUnit fails.
+        If True, raise an exception if a ProtocolUnit fails, default True
+        if False, any exceptions will be stored as `ProtocolUnitFailure`
+        objects inside the returned `ProtocolDAGResult`
 
     Returns
     -------
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -222,6 +222,8 @@ def execute(protocoldag: ProtocolDAG, *,
        Path to scratch space that persists across whole DAG execution, but
        is removed after. Used by some `ProtocolUnit`s to pass file contents
        to dependent `ProtocolUnit`s.
+    raise_error : bool
+        If True, raise an exception if a ProtocolUnit fails.
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


dotsdl:
Needs tests yet.

dotsdl:
Ready to go! I don't understand where all the `mypy` failures are coming from, since these appear irrelevant to the changes given here.

dotsdl:
Looks like the new mypy complaints are coming from `mypy 0.990`, which came out four days ago.

gufe/tests/test_tokenization.py
```diff
@@ -80,7 +80,7 @@ def instance(self):
         """
         ...
 
-    def teardown(self):
+    def teardown_method(self):
```
dotsdl, line 83: Small fix to get rid of massive deprecation warnings from `pytest`.

richardjgowers:
Mypy drama is fixed in #94

dwhswenson:
> Looks like the new mypy complaints are coming from `mypy 0.990`, which came out four days ago. I'd rather we fix those in a separate PR.

@dotsdl : see #94.

dotsdl:
Haha thanks y'all, only just noticed #94 myself, then saw your comments here :sweat_smile:

dwhswenson (approved):
For discussion: Should we break API here and use `raise_error=True` for default?

My argument is that any normal user who uses this will probably want to raise an error in this case. Any developer who wants to suppress the errors can set the flag. Breaking API is annoying to anyone currently using this method in their application code -- which would probably be just me and you, so if we will want to change it, now is the time.

Otherwise, LGTM.

gufe/protocols/protocoldag.py
```diff
@@ -210,7 +210,9 @@ def _from_dict(cls, dct: Dict):
 
 
 def execute(protocoldag: ProtocolDAG, *,
```
richardjgowers, line 212: unrelated, but can we change this to `execute_DAG()`, it always confuses me WRT `Unit.execute()`

richardjgowers (changes requested):
Is there something in error propagation that makes this necessary?  If so, we should probably change how we're catching and storing errors.

If errors are being caught fine, what's wrong with just...
```python
r = unit.execute(...)
if not r.ok():
    raise Error
```
on the scheduler side?  This PR seems to add in something which is fewer lines of code externally.

dwhswenson:
> Is there something in error propagation that makes this necessary?

The problem here is that we're exposing `execute` as the recommended way for people to run this when they're working without one of our (as-yet incomplete) schedulers. So they're running `execute` from a Jupyter notebook or from their custom Python scripts.

Then they get errors, but don't see them in any way they expect, because we're swallowing the error so that we can treat it as a first-class result. That's the right thing when schedulers are your clients, but not when users are.

IMO, this is a pretty straightforward way to reduce that confusion. I also think it would be further reduced by changing the default to `raise_error=True`: that's what the naive user expects, and anyone building a scheduler system should be smart enough to know to set `raise_error=False` when they call `execute`.

richardjgowers:
@dwhswenson ok that's a good argument for `raise_error` in `execute` (the function) but not the method on `Unit`.  I don't like that we made a decision to have a `Failure` result, and then we're immediately not using it... I think better would be to write the error catching / raising in `execute` (the function) so it serves as an example executor.

dwhswenson:
The main advantage of the unit having an option to raising the error is that it makes it so that errors within the unit act like all other errors the user might encounter, without us having to actually serialize exceptions. The issue is whether we want to actually raise the original error, or raise a separate error (at the executor level) that says "An error occurred." 

#### When do we want the actual error?

* **Serial executor**: In the serial executor, my intuition would be to raise the actual error, which is what this code does. For one thing, a serial executor might be able to handle some types of errors. It would be more natural for a user to handle these with the standard Python `except`, rather than parsing the string representation of the error that we return.
I think the "double stack trace" might be somewhat confusing. Finally, these errors may not be consistent with other errors in the user's environment: for example, `ipython` and `rich` provide mechanisms to reformat tracebacks. We lose those if we just `print` the current traceback.

* **Worker-based executor**: It's not clear to me that any worker-based executor would want to the actual error. Erroring in the unit would mean crashing the worker, which seems like a bad idea. Erroring in the executor makes more sense, although I don't think any of our planned worker-based executors want the executor to crash on a worker error. Any error handling of the actual error would be based on an `isinstance` check, not a try/except (since the executor would need to deserialize the exception). I think that the error message inconsistency I didn't like in the serial version would actually be a feature, not a bug, for the worker-based executor: It would be easier to distinguish the worker's traceback from the executor's.

#### Mechanisms to raise the actual error

1. **Raise it from within the unit**: This is what this code does. 

2. **Have full serialization/deserialization of exceptions**: This is an option we rejected earlier. I'd still say that it seems like a lot of effort to get it right, for very little benefit.

gufe/protocols/protocoldag.py
dotsdl: I'm fine with either name. Agree it's easily confused with `Unit.execute()`; was easier to keep straight when this was the `ProtocolDAG.execute()` method, but now not so much.

richardjgowers (changes requested):
Ok we can keep the two extra kwargs.  Can we rename `execute()` to `execute_DAG` or something, and also change the default of `raise_error` to `True` in this function (but not the method).

gufe/protocols/protocoldag.py
richardjgowers: renamed to execute_DAG

richardjgowers (approved):

PR #97, opened 2022-11-21
succesful results -- Adding Property to ProtocolDag 

RiesBen:
adding property, this makes it easier and solves a problem in openfe nicely.
It basically just adds another attribute, similar to `protocol_unit_failures` , wich will help to remove code duplication all over our projects.

This also contributes to solve a problem in equil_rbfe_methods.py, where only successful runs are required.


richardjgowers (changes requested):
Needs a test but looks good.

Might be worth quickly deciding with @dwhswenson and @dotsdl if `protocol_unit_results` should implicitly be only successes (so getting all results becomes `protocol_unit_results + protocol_unit_failures`).

RiesBen:
Good point :) 

However I think renaming `ProtocolUnitResult` to `ProtocolUnitSuccess` would be much cleaner.
Then `ProtocolUnitResultBase` gets `ProtocolUnitResult` and following logic would apply:

- protocol_unit_results = Iterable[ProtocolUnitSuccess] +Iterable[ProtocolUnitFailure]
- protocol_unit_success = Iterable[ProtocolUnitSuccess]
- protocol_unit_failures = Iterable[ProtocolUnitFailure]

dwhswenson (approved):
LGTM. I think we're going to get enough requests for this, and I can see ways this might be convenient in #102, as well.

richardjgowers (approved):

RiesBen:
Thanks a lot for the merge :)

PR #100, opened 2022-11-29
Speed up protein test loading

dwhswenson:
This essentially makes our predefined PDB loaders session-scoped. This cuts out a lot of network activity and should give us as notable speedup when running the full test suite.

In principle, it would probably be better to structure everything to use the `request` fixture, but this is a smaller code change for essentially the same result.

### DIFF
--- a/gufe/tests/conftest.py
+++ b/gufe/tests/conftest.py
@@ -22,14 +22,21 @@ else:
 
 ## helper functions
 
+class URLFileLike:
+    def __init__(self, url, encoding='utf-8'):
+        self.url = url
+        self.encoding = encoding
+        self.data = None
 
-def load_url_data(url, encoding='utf-8'):
-    if not HAS_INTERNET:
-        pytest.skip("Skipping because internet seems faulty")
+    def __call__(self):
+        if not HAS_INTERNET:  # pragma: no-cover
+            pytest.skip("Skipping because internet seems faulty")
 
-    req = urllib.request.urlopen(url)
-    # we convert to StringIO because req.read() returns bytes, not string
-    return io.StringIO(req.read().decode(encoding))
+        if self.data is None:
+            req = urllib.request.urlopen(self.url)
+            self.data = req.read().decode(self.encoding)
+
+        return io.StringIO(self.data)
 
 
 def get_test_filename(filename):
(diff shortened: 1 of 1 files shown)


dwhswenson:
This didn't speed stuff up as much as I had expected in CI -- I wonder if there's some local caching of network activity or something? (That would be fancy, but probably worth the savings for GitHub.) Or maybe the performance of the test suite just varies by more than the speed boost here, especially on macOS, where I think we're heavily throttled.

It does speed things up when I run tests locally, so hoping this will get merged anyway. Ready for review.

RiesBen (approved):
hi, looks good to me :) 
I guess one could also play around with the timeout option for `urllib.request.urlopen`. But this makes it a bit less robust.

gufe/tests/conftest.py
```diff
@@ -22,14 +22,21 @@
...
+            pytest.skip("Skipping because internet seems faulty")
 
-    req = urllib.request.urlopen(url)
-    # we convert to StringIO because req.read() returns bytes, not string
-    return io.StringIO(req.read().decode(encoding))
+        if self.data is None:
+            req = urllib.request.urlopen(self.url)
+            self.data = req.read().decode(self.encoding)
+
+        return io.StringIO(self.data)
```
richardjgowers, line 39: This is fun, but why can't we use a pytest session scoped fixture here?  Otherwise it's an odd pattern that slows down anyone trying to work on these tests

richardjgowers (changes requested):

gufe/tests/conftest.py
dwhswenson: Goals:

1. Have reasonable test labels so that if one PDB test fails, we immediately see the name of the file that fails.
2. Don't have the boilerplate of writing a new fixture for every PDB file we get from a URL.
3. Don't require downloading *all* PDB files if you're only testing one of them.
4. Only download URLs once per session.

Is there a way to accomplish all 4 goals with pytest fixtures? I don't immediately see it. A single fixture to create a dict with all URLs breaks 3. A fixture per URL breaks 2.

Existing code accomplishes 1-3, this change adds 4.

richardjgowers (approved):

PR #101, opened 2022-11-30
Improvements for settings and serialization/deserialization

ijpulidos:
In order for the NEQ protocol on perses to work at https://github.com/choderalab/perses/pull/1066 it was noted that the gufe settings objects needed some improvements, namely, defining defaults for the fields and changes in the serialization/deserialization.

As far as we discussed with @mikemhenry , there is still some discussion about either having defaults that are hidden from the users or make the users to be explicit about them.

This is just a first incomplete attempt at a solution.

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -124,10 +124,11 @@ class Protocol(GufeTokenizable):
         return {}
 
     def _to_dict(self):
-        return {'settings': self.settings}
+        return {'settings': self.settings.json()}
 
     @classmethod
     def _from_dict(cls, dct: Dict):
+        # TODO: We need to deserialize from a json, not a dict
         return cls(**dct)
 
     @classmethod
--- a/gufe/settings/models.py
+++ b/gufe/settings/models.py
@@ -56,13 +56,13 @@ class VdWSettings(SettingsBaseModel):
     for more details.
     """
 
-    combining_rules: Literal["Lorentz-Berthelot"]
-    potential: Literal["Lennard-Jones-12-6"]
-    scale: vdWScale
-    long_range_dispersion: Literal["isotropic"]
+    combining_rules: Literal["Lorentz-Berthelot"] = "Lorentz-Berthelot"
+    potential: Literal["Lennard-Jones-12-6"] = "Lennard-Jones-12-6"
+    scale: vdWScale = vdWScale()
+    long_range_dispersion: Literal["isotropic"] = "isotropic"
     cutoff: FloatQuantity["angstrom"] = Field(9.0, description="Default units are angstroms")
     switch_width: FloatQuantity["angstrom"] = Field(1.0, description="Default units are angstroms")
-    method: Literal["cutoff"]
+    method: Literal["cutoff"] = "cutoff"
 
 
 class ElectrostaticSettings(SettingsBaseModel):
@@ -74,10 +74,10 @@ class ElectrostaticSettings(SettingsBaseModel):
     """
 
     # Tricky since this allows functions https://openforcefield.github.io/standards/standards/smirnoff/#electrostatics
-    periodic_potential: Union[Literal["Ewald3D-ConductingBoundary"], str]
-    nonperiodic_potential: Union[Literal["Coulomb"], str]
-    exception_potential: Union[Literal["Coulomb"], str]
-    scale: ElectrostaticScale
+    periodic_potential: Union[Literal["Ewald3D-ConductingBoundary"], str] = "Ewald3D-ConductingBoundary"
+    nonperiodic_potential: Union[Literal["Coulomb"], str] = "Coulomb"
+    exception_potential: Union[Literal["Coulomb"], str] = "Coulomb"
+    scale: ElectrostaticScale = ElectrostaticScale()
     cutoff: FloatQuantity["angstrom"] = Field(None, description="Default units are in angstroms")
     switch_width: FloatQuantity["angstrom"] = Field(None, description="Default units are in angstroms")
     solvent_dielectric: Optional[float]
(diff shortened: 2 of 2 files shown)


gufe/protocols/protocol.py
```diff
@@ -124,10 +124,11 @@ def _defaults(self):
         return {}
 
     def _to_dict(self):
-        return {'settings': self.settings}
+        return {'settings': self.settings.json()}
 
     @classmethod
     def _from_dict(cls, dct: Dict):
+        # TODO: We need to deserialize from a json, not a dict
```
dwhswenson, line 131: You should still be receiving a dict, but it is a dict with a single key `settings`, which maps to a JSON string. The stuff in openff-models should have a way to reload from that JSON string; @mikemhenry can probably tell you details.

Should look something like:

```python
@classmethod
def _from_dict(cls, dct: Dict):
    settings = some_openff_models_magic(dct['settings'])
    return cls(settings=settings)
```

dwhswenson (commented):
Looking only at the serialization side:

From a "get it working" perspective, this is a great start; thanks! For the long term, I think we'll probably switch to a different approach, though.

The reason for a special JSON serialization in openff-models is that there are types in the models (i.e., OpenFF units) that are not natively serializable by JSON. However, any type that you might need as a setting is also something you might need as a result, or as an attribute of an object from our data models (e.g., `SmallMoleculeComponent`) -- units being a prime example that are needed at all levels.

We have a JSON serialization framework that allows extension to custom types. It already includes support for OpenFF units, numpy arrays, and `pathlib.Path` objects (see #68). In general, I'd rather type handling for serialization be handled in one place, so that a contributor only needs to implement it once. This approach means that we can automatically consume any known type without contributors needing to worry about how to serialize that type, and it is easy for contributors to add to our known types.

If you come across any other types that don't serialize, there's an open PR (#81) with documentation on how to add support for custom types. It also contains some generally useful info for writing (and testing) `GufeTokenizable`s, so I'd recommend taking a glance at it -- for example, we have a serialization testing mix-in, so you don't need to write your own serialization tests.

This means that in the long term, I'd like the `_to_dict` to just return `self.settings.dict()` (which should a be deep (recursive) dictionary), and `_from_dict` should have a similar ability to recreate the objects from that dictionary. It looks to me like that isn't currently possible with `openff-models`, although @mikemhenry will know better than I do.

Note to self (and team): We could probably make a JSON codec that consumes `Settings` -- am I correct that all settings inherit from a single base class?

gufe/protocols/protocol.py
mikemhenry: It should be as "simple" as using something like `parse_raw` https://pydantic-docs.helpmanual.io/usage/models/#helper-functions to turn the json string back into a settings object. I know this works since that is how I setup the test.

dwhswenson:
> This is correct, I'll take a stab at doing this like you outlined in #68 but it might require some changes to openff-models (which is fine). 

This should be as easy as defining a method to create a (shallow) dictionary of the model. Is `model.dict()` deep (recursive) or shallow? In the worst case, it should be easy to introspect the field names and use `getattr` to make the dict. The `from_dict` will just be `cls(**input_dict)` -- all the recursion and handling of other types should be handled by the JSON system.

Side note: I'm recommending making a `JSONCodec` over making `Settings` into `GufeTokenizable`s. The main difference is in deduplication -- as a `GufeTokenizable`, each `Settings` instance would only show up once; so if you reused a given thermodynamic state in different protocols, it would only be stored one time. That sounds nice, but the downside is that you don't have a single "file" (literal file in the planned OpenFE storage) that contains the entire settings. Using `JSONCodec` keeps all settings for a protocol in one location, at the cost of duplication.

> I think my point of confusion here was that in pydantic land, there are methods to customize the serialization and deserialization of objects, but it looks like with this framework, the "knowledge" on how to do that exists outside the class.

Yeah, part of the distinction here is that we need it in multiple places. If all you needed to serialize was the pydantic model, and you never needed to worry about users externally adding support for more types, then the approach is openff-models is great. But we may need to worry about users coming with other external types, and we definitely need to serialize in contexts other than the pydantic model.

richardjgowers:
@dwhswenson Yeah all Settings ought to (not enforced, duck typing is ok) inherit from some `gufe.Settings` base class (which hopefully inherits from both `GufeTokenizable` and `openff-models.BaseModel`), so ideally we can make these objects "just work" if they only contain the data types we've already got codecs for

dwhswenson:
(sorry; somehow the first time I wrote this I edited @richardjgowers's comment instead of making a quote-reply)

> @dwhswenson Yeah all Settings ought to (not enforced, duck typing is ok) inherit from some `gufe.Settings` base class (which hopefully inherits from both `GufeTokenizable` and `openff-models.BaseModel`), so ideally we can make these objects "just work" if they only contain the data types we've already got codecs for

If it is a `GufeTokenizable`, that will take precedence. No need for `JSONCodec`.

However, if it is a `GufeTokenizable`, then every instance of a specific `Settings` subclass will get deduplicated (which means instead of having one JSON file in our storage with all the settings, we'll have lots of JSON files with parts of the settings). I though we said we wanted the deduplication to stop at the level of settings.

You might be able to get around that by having each settings object's `_to_dict` call each subsetting's `to_dict()` (no underscore; deep dictionary). That might create some recursion mess, would have to try it to see.

Other points:

* `GufeTokenizable` can't do pure duck-typing because you at least need the metaclass. Current implementation of `is_gufe_obj` is also based on an `isinstance` check.
* Both `GufeTokenizable` and `openff_models.BaseModel` will have a lot of serialization code. If inheriting from both, we should be very careful that one doesn't shadow the other. (I think that pydantic models have `dict()`, not `to_dict()`, so maybe we're okay -- but this risks confusion.)

mikemhenry:
I'd rather not make settings from `GufeTokenizable` and instead take the encoder approach.

I think as it stands, `openff-models` dict isnt recursive (as in it will keep units as an object) but json will serialize it to text, so I should be able to do this with just the encoder approach (which I will try once I finish up a pymbar 3.x release and dockerize fah-alchemy)

mikemhenry:
Going to close this PR in favor of #105 for the json stuff, and a new one I will open to make it easier to get the default settings, change the stuff you want, then make a settings object.

PR #102, opened 2022-11-30
[WIP] Various result handling improvements from working on `fah-alchemy`

dotsdl:
This PR will be a grab-bag of changes to ProtocolDAGResult and friends from results handling development in `fah-alchemy`.

Trying to keep number of changes to minimum necessary.

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -206,6 +206,18 @@ class ProtocolDAGResult(GufeTokenizable, DAGMixin):
         return all(any(pur.ok() for pur in self._unit_result_mapping[pu])
                    for pu in self._protocol_units)
 
+    def terminal_protocol_unit_results(self) -> list[ProtocolUnitResult]:
+        """Get ProtocolUnits that terminate the DAG
+
+        Returns
+        -------
+        terminal_units : list[ProtocolUnit]
+          any ProtocolUnits which do not have a ProtocolUnit which follows on
+          (depends) on them
+        """
+        return [u for u in self._protocol_unit_results
+                if not nx.ancestors(self._result_graph, u)]
+
 
 class ProtocolDAG(GufeTokenizable, DAGMixin):
     """An executable directed, acyclic graph (DAG) composed of `ProtocolUnit`
--- a/gufe/protocols/protocolunit.py
+++ b/gufe/protocols/protocolunit.py
@@ -85,7 +85,7 @@ class ProtocolUnitResult(GufeTokenizable):
 
     def _gufe_tokenize(self):
         # tokenize with uuid
-        return uuid.uuid4()
+        return uuid.uuid4().hex
 
     @classmethod
     def _defaults(cls):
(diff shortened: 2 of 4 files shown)


gufe/protocols/protocoldag.py
```diff
@@ -141,9 +141,15 @@ def result_to_unit(self, protocol_unit_result: ProtocolUnitResult):
             raise KeyError("No such `protocol_unit_result` present")
 
     def ok(self) -> bool:
-        # ensure that for every protocol unit, there is definitely a
-        # corresponding result
-        return set(self._protocol_units) == set(self._unit_result_mapping.keys())
+        # ensure there are no failed untis and that for every protocol unit,
+        # there is definitely a corresponding result
+        return (len(self.protocol_unit_failures) == 0 and
+                set(self._protocol_units) == set(self._unit_result_mapping.keys()))
```
richardjgowers, line 147: if we're having Fails as first class, then not having failures isn't a requirement.  e.g I have a DAG with A, B, C units, I run with a scheduler that allows 3 retries, my results might be `[A_ok, B_fail, B_ok, C_ok]`, and that should be `ok() -> True`.  This is what the current `ok()` method does, checks that for every Unit there is at least one success

If we want to have it work this way, we need somewhere to store the Failures

gufe/protocols/protocoldag.py
```diff
@@ -141,9 +141,15 @@ def result_to_unit(self, protocol_unit_result: ProtocolUnitResult):
...
+
+
+    # TODO: add a convenience method to get terminal protocol units
+    def terminal_protocol_unit_results(self):
+        ...
```
richardjgowers, line 152: I can't remember which way round our DAG is, but this is something like:
 `[pur for pur in self.protocol_unit_results if not nx.ancestors(self._result_graph, pur)]`

gufe/protocols/protocoldag.py
richardjgowers: Looking at the source we can't currently hold multiple results for a given Unit, but we should imho (and it's easy enough)

richardjgowers (changes requested):

gufe/protocols/protocoldag.py
dotsdl: Oh, the current implementation doesn't actually check for at least one success, but at least one result/failure for each unit. So for a `ProtocolDAG` that fails on only the last unit, it was yielding `ok() -> True`. This is a legit bug that I'll fix here.

dwhswenson: Do we want to merge #97 and maybe build on the convenience property there?

gufe/protocols/protocoldag.py
dotsdl: Something like that! I think this will be useful for protocol authors in writing their `Protocol._gather` methods, since they'll likely pull things out of the `ProtocolUnitResult`s with no dependencies. I'm considering using terminal nodes as the basis for what paths get pushed to the objectstore in `fah-alchemy`, so probably would use this method in my compute services as well.

richardjgowers:
@dotsdl I think I finished your PR, check I got your intent right and merge at will

richardjgowers (approved):

dotsdl:
Many thanks @richardjgowers! Merging!

PR #103, opened 2022-12-01
Protocoldagresult multiple results

richardjgowers:
Somewhat of a rebuttal to #102

Allows many `ProtocolUnitResult`s per `ProtocolUnit` in a `ProtocolDAGResult`

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -2,8 +2,9 @@
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
 import abc
+from collections import defaultdict
 import os
-from typing import Iterable, List, Dict, Set, Optional, Union, Any
+from typing import Iterable, List, Optional, Union, Any
 from os import PathLike
 from pathlib import Path
 import tempfile
@@ -68,7 +69,7 @@ class ProtocolDAGResult(GufeTokenizable, DAGMixin):
     protocol_units : List[ProtocolUnit]
         `ProtocolUnit`s (given in DAG-dependency order) used to compute this
         `ProtocolDAGResult`.
-    protocol_unit_results : List[ProtocolUnit]
+    protocol_unit_results : List[ProtocolUnitResult]
         `ProtocolUnitResult`s (given in DAG-dependency order) corresponding to
         each `ProtocolUnit` used to compute this `ProtocolDAGResult`.
     graph : nx.DiGraph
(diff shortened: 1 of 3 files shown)


gufe/protocols/protocolunit.py
```diff
@@ -49,7 +49,7 @@ def _list_dependencies(inputs, cls):
     return deps
 
 
-class ProtocolUnitResultBase(GufeTokenizable):
```
richardjgowers, line 52: I got rid of the base class as it was tiresome to type `Union[ProtocolUnitResult, ProtocolUnitFailure]` everywhere.  Now Failure is just a special case of Result

gufe/tests/test_protocol.py
```diff
@@ -226,7 +226,7 @@ def test_dag_execute(self, protocol_dag):
         assert finishresult.name == "the end"
 
         # gather SimulationUnits
-        simulationresults = [dagresult.unit_to_result(pu)
+        simulationresults = [dagresult.unit_to_results(pu)[0]
```
richardjgowers, line 229: `unit.to_result` is renamed `unit_to_results` as there are now many results per Unit.  This is slightly ugly, as more often than not, people will be interested in the success Result unless they explicitly ask for *all* results.

Perhaps I could keep `unit_to_result` (singular) and make this always return the corresponding `.ok()` Result, then `unit_to_all_results` would give the list of multiple possible Results (some fails possible)

gufe/protocols/protocoldag.py
```diff
@@ -113,38 +118,61 @@ def _to_dict(self):
...
         except KeyError:
             raise KeyError("No such `protocol_unit_result` present")
 
     def ok(self) -> bool:
-        # ensure that for every protocol unit, there is definitely a
-        # corresponding result
-        return set(self._protocol_units) == set(self._unit_result_mapping.keys())
+        # ensure that for every protocol unit, there is an OK result object
+        return all(any(pur.ok() for pur in self.unit_to_results(pu))
+                   for pu in self._protocol_units)
```
richardjgowers, line 175: this should fix the bug that @dotsdl noted in the #102 PR

dotsdl:
Hitting this today; thanks @richardjgowers! Will definitely help resolve or obviate #102.

gufe/tests/test_protocol.py
dotsdl: > Perhaps I could keep `unit_to_result` (singular) and make this always return the corresponding `.ok()` Result, then `unit_to_all_results` would give the list of multiple possible Results (some fails possible)

I like this proposal. I think it has the characteristics we want here.

dotsdl (approved):
Just one comment, agreeing with your proposal that would improve usability with a success comingled among 0 or more failures for each `ProtocolUnit`. Giving approval now assuming you go forward with that.

dotsdl:
Thanks for putting this together @richardjgowers! This will reduce the scope of #102, and these are really nice additions.

PR #105, opened 2022-12-09
Fix issues with custom JSON codecs

dwhswenson:
This fixes issues with inheritance in classes given to JSON codecs. It also adds codecs for `Settings` (technically `SettingsBaseModel`) and for OpenFF units/quantities (I though we already had these? Apparently not.)

The following now works for me, when this branch is combined with the cleanup in #101 making it easier to deal with forcefield settings:

```python
import json
from gufe.settings import settings
from openff.units import unit

s = settings.Settings(
    forcefield_file="foo.bar",
    forcefield_settings=settings.ForcefieldSettings(),
    thermo_settings=settings.ThermoSettings(temperature=300 * unit.kelvin)
)
dumped = json.dumps(s, cls=JSON_HANDLER.encoder)
loaded = json.loads(dumped, cls=JSON_HANDLER.decoder)
```

There's enough in here that I thought I'd make is a separate PR from #101; but this will be very useful to some of the goals over there.

### DIFF
--- a/gufe/custom_codecs.py
+++ b/gufe/custom_codecs.py
@@ -6,7 +6,28 @@
 import pathlib
 import numpy as np
 
+import gufe
 from gufe.custom_json import JSONCodec
+from gufe.settings.models import SettingsBaseModel
+import openff.units
+import functools
+
+def default_from_dict(dct):
+    dct = dict(dct)  # make a copy
+    module = dct.pop('__module__')
+    qualname = dct.pop('__class__')
+    del dct[':is_custom:']
+
+    cls = gufe.tokenization.get_class(module, qualname)
+    return cls(**dct)
+
+
+def inherited_is_my_dict(dct, cls):
+    dct = dict(dct)
+    module = dct.pop('__module__')
+    classname = dct.pop('__class__')
+    stored = gufe.tokenization.get_class(module, classname)
+    return cls in stored.mro()
 
 
 PATH_CODEC = JSONCodec(
@@ -33,4 +54,23 @@ NUMPY_CODEC = JSONCodec(
     ).reshape(dct['shape'])
 )
 
+SETTINGS_CODEC = JSONCodec(
+    cls=SettingsBaseModel,
+    to_dict=lambda obj: {field: getattr(obj, field)
+                         for field in obj.__fields__},
+    from_dict=default_from_dict,
+    is_my_dict=functools.partial(inherited_is_my_dict,
+                                 cls=SettingsBaseModel)
+)
 
+OPENFF_QUANTITY_CODEC = JSONCodec(
+    cls=openff.units.units.Quantity,
+    to_dict=lambda obj: {'magnitude': obj.m, 'unit': obj.u},
+    from_dict=lambda dct: dct['magnitude'] * dct['unit']
+)
+
+OPENFF_UNIT_CODEC = JSONCodec(
+    cls=openff.units.units.Unit,
+    to_dict=lambda obj: {'unit': str(obj)},
+    from_dict=lambda dct: getattr(openff.units.unit, dct['unit']),
+)
--- a/gufe/custom_json.py
+++ b/gufe/custom_json.py
@@ -66,8 +66,8 @@ class JSONCodec(object):
             dct = {}
             if self.cls:
                 dct.update({
-                    '__class__': self.cls.__name__,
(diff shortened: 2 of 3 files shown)


dwhswenson:
@mikemhenry : I need to get some fah-alchemy stuff done now; if you have time to pick up the tests here (basically fill in what is needed to use the `CustomJSONCodingTest` ABC in `test_custom_json.py`) that would be helpful. Otherwise, I'll come back to those early next week.

mikemhenry:
Sounds, this is so helpful! Thanks!!!

dwhswenson:
Looks like, now that we're properly reporting the subclass, the tests for `pathlib.Path` should change a bit, too -- the new results are probably what we should have had from the beginning (specific subclass, not generic superclass)

dwhswenson:
One other point I noticed: the object doesn't exactly round-trip, because in the original settings object `s` (based on defaults in #101) several values don't have units. When they go through a serialization/deserialization cycle, they acquire the units. (`cutoff=9` and `switch_width=1` both acquire units of angstroms after serialization).

gufe/custom_codecs.py
```diff
@@ -33,4 +54,23 @@
...
+    cls=openff.units.units.Quantity,
+    to_dict=lambda obj: {'magnitude': obj.m, 'unit': obj.u},
+    from_dict=lambda dct: dct['magnitude'] * dct['unit']
+)
 
+OPENFF_UNIT_CODEC = JSONCodec(
+    cls=openff.units.units.Unit,
+    to_dict=lambda obj: {'unit': str(obj)},
+    from_dict=lambda dct: getattr(openff.units.unit, dct['unit']),
+)
```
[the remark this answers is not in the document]
ijpulidos: I also have the feeling you can do that, maybe this helps https://pint.readthedocs.io/en/stable/getting/tutorial.html?highlight=automatically#string-parsing

gufe/custom_json.py
```diff
@@ -66,8 +66,8 @@ def default(self, obj: Any) -> Any:
             dct = {}
             if self.cls:
                 dct.update({
-                    '__class__': self.cls.__name__,
-                    '__module__': self.cls.__module__,
+                    '__class__': obj.__class__.__qualname__,
+                    '__module__': obj.__class__.__module__,
```
mikemhenry, line 70: What do these changes do? I think it broke the test:
```
FAILED gufe/tests/test_custom_json.py::TestPathCodec::test_default - AssertionError: assert {':is_custom:...h': 'foo/bar'} == {':is_custom:...h': 'foo/bar'}
  Omitting 3 identical items, use -vv to show
  Differing items:
  {'__class__': 'PosixPath'} != {'__class__': 'Path'}
  Full diff:
    {
     ':is_custom:': True,
  -  '__class__': 'Path',
  +  '__class__': 'PosixPath',
  ?                +++++
     '__module__': 'pathlib',
     'path': 'foo/bar',
    }
FAILED gufe/tests/test_custom_json.py::TestPathCodec::test_round_trip - AssertionError: assert {':is_custom:': True, '__class__': 'PosixPath', '__module__': 'pathlib', 'path': 'foo/bar'} == PosixPath('foo/bar')
FAILED gufe/tests/test_tokenization.py::test_path_to_json - AssertionError: assert PosixPath('foo/bar') == {':is_custom:': True, '__class__': 'PosixPath', '__module__': 'pathlib', 'path': 'foo/bar'}
FAILED gufe/tests/test_transformation.py::TestTransformation::test_dump_load_roundtrip - KeyError: '__class__'
```

dwhswenson: Commented [here](https://github.com/OpenFreeEnergy/gufe/pull/105#issuecomment-1344848538):

> Looks like, now that we're properly reporting the subclass, the tests for `pathlib.Path` should change a bit, too -- the new results are probably what we should have had from the beginning (specific subclass, not generic superclass)

Old version used the saved superclass, and would always recreate an instance of that, instead of the specific subclass. Obviously, we want to recreate a subclass, not the generic superclass.

mikemhenry: Got it! Now I understand what you mean by:

> the new results are probably what we should have had from the beginning

dotsdl:
Is there anything blocking this PR?

dotsdl:
Closing in favor of #110.

PR #109, opened 2022-12-16
Default Settings

mikemhenry:
Remove all the cruft we don't need right now. Makes it easy for a user to grab the default settings and modify them as needed:

``` python
from gufe.settings.models import Settings
from openff.units import unit

my_settings = Settings.get_defaults()
my_settings.thermo_settings.temperature = 298 * unit.kelvin
```

I thought about programmatically walking through the `Settings` object to create the defaults (so if things change, we wouldn't necessarily have to re-write the method) but this would give us the flexibility of a default for a pydantic field to be `None` but let us choose something different for `get_defaults`, e.g. `temperature=300 * unit.kelvin`


dotsdl:
Closing in favor of #110.

PR #110, opened 2022-12-16
Custom json updates and default settings

mikemhenry:
This PR combines #109 and #105 

Not saying we merge this one instead of the others separately, but they are so related for testing it was easier to merge them into one.


gufe/tests/test_custom_json.py
```diff
@@ -122,7 +128,31 @@ def setup(self):
...
+        self.codec = OPENFF_UNIT_CODEC
+        self.objs = [
+            unit.amu,
+        ]
+        self.dcts = [
+            {
+                ":is_custom:": True,
+                "__class__": "Unit",
+                "__module__": "pint.util",
+                "unit": "unified_atomic_mass_unit",
```
mikemhenry, line 156: @dwhswenson so something that I think is an issue that might need to be fixed in openff-units is how this class comes in, since it is showing up as pint and not an openff-units class

dotsdl:
Should we just retire #109 and #105 and review this one? Will that make review simpler?

mikemhenry:
@dwhswenson Should I fix this warnings 
```
gufe/tests/test_custom_json.py::TestNumpyCoding::test_not_mine
  /home/mmh/miniconda3/envs/gufe-serial/lib/python3.10/site-packages/_pytest/fixtures.py:900: PytestRemovedIn8Warning: Support for nose tests is deprecated and will be removed in a future release.
  gufe/tests/test_custom_json.py::TestNumpyCoding::test_not_mine is using nose-specific method: `setup(self)`
  To remove this warning, rename it to `setup_method(self)`
  See docs: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose
    fixture_result = next(generator)
```

Or fix them in a different PR?

dwhswenson:
@mikemhenry : I'd vote for a separate PR (if one isn't already open?) I was noticing the same as I'm finishing one of the storage PRs here.

mikemhenry:
One last thing I need to do is load up Ivan's notebook and make sure things still work then this will get good to merge (and fix mypy and sphinx)

mikemhenry:
Ivan's notebooks works! 
```
{
  "__class__": "Settings",
  "__module__": "gufe.settings.models",
  ":is_custom:": true,
  "settings_version": 0,
  "forcefield_settings": {
    "__class__": "OpenMMSystemGeneratorFFSettings",
    "__module__": "gufe.settings.models",
    ":is_custom:": true,
    "forcefield_kwargs": {
      "constraints": "app.HBonds",
      "rigidWater": true,
      "removeCMMotion": false,
      "hydrogenMass": "4*unit.amu"
    },
    "forcefields": [
      "amber/ff14SB.xml",
      "amber/tip3p_standard.xml",
      "amber/tip3p_HFE_multivalent.xml",
      "amber/phosaa10.xml"
    ],
    "small_molecule_forcefield": "openff-2.0.0"
  },
  "thermo_settings": {
    "__class__": "ThermoSettings",
    "__module__": "gufe.settings.models",
    ":is_custom:": true,
    "temperature": {
      "magnitude": 300,
      "unit": "kelvin",
      ":is_custom:": true,
      "pint_unit_registry": "openff_units"
    },
    "pressure": null,
    "ph": null,
    "redox_potential": null
  },
  "protocol_settings": {
    "__class__": "NonEqCyclingSettings",
    "__module__": "perses.protocols.settings",
    ":is_custom:": true,
    "lambda_functions": {
      "lambda_sterics_core": "lambda",
      "lambda_electrostatics_core": "lambda",
      "lambda_sterics_insert": "select(step(lambda - 0.5), 1.0, 2.0 * lambda)",
      "lambda_sterics_delete": "select(step(lambda - 0.5), 2.0 * (lambda - 0.5), 0.0)",
      "lambda_electrostatics_insert": "select(step(lambda - 0.5), 2.0 * (lambda - 0.5), 0.0)",
      "lambda_electrostatics_delete": "select(step(lambda - 0.5), 1.0, 2.0 * lambda)",
      "lambda_bonds": "lambda",
      "lambda_angles": "lambda",
      "lambda_torsions": "lambda"
    },
    "softcore_LJ_v2": true,
    "interpolate_old_and_new_14s": false,
    "timestep": {
      "magnitude": 4.0,
      "unit": "femtosecond",
      ":is_custom:": true,
      "pint_unit_registry": "openff_units"
    },
    "neq_splitting": "V R H O R V",
    "eq_steps": 1000,
    "neq_steps": 100,
    "platform": "CUDA",
    "save_frequency": 100
  }
}
```

The serialization code correctly points to the right module in perses!

mikemhenry:
Ready for review!

gufe/custom_codecs.py
```diff
@@ -3,16 +3,54 @@
 
 # custom_codecs.py: A place to keep various custom JSONCodec instances
 
+import functools
 import pathlib
+
 import numpy as np
+import openff.units
```
dwhswenson, line 10: this can be removed (see other comment)

gufe/custom_codecs.py
```diff
@@ -33,4 +71,36 @@
...
+
+OPENFF_QUANTITY_CODEC = JSONCodec(
+    cls=None,
+    to_dict=lambda obj: {
+        "magnitude": obj.m,
+        "unit": str(obj.u),
+        ":is_custom:": True,
+        "pint_unit_registry": "openff_units",
+    },
+    from_dict=lambda dct: openff.units.DEFAULT_UNIT_REGISTRY(
```
dwhswenson, line 89: drop `openff.units`, since we've already imported `DEFAULT_UNIT_REGISTRY` (allows us to drop `openff.units` import, as well)

gufe/custom_codecs.py
```diff
@@ -33,4 +71,36 @@
...
+)
 
+OPENFF_UNIT_CODEC = JSONCodec(
+    cls=None,
+    to_dict=lambda unit: {
+        ":is_custom:": True,
+        "pint_unit_registry": "openff_units",
+        "unit_name": str(unit),
+    },
+    from_dict=lambda dct: openff.units.DEFAULT_UNIT_REGISTRY(dct["unit_name"]).u,
```
dwhswenson, line 103: drop `openff.units`, since we've already imported `DEFAULT_UNIT_REGISTRY` (allows us to drop `openff.units` import, as well)

gufe/settings/models.py
```diff
@@ -3,130 +3,77 @@
...
-    vdW: VdWSettings
-    electrostatics: ElectrostaticSettings
-    gbsa: GBSASettings
+    ...
 
 
-class ThermoSettings(SettingsBaseModel):
-    """Settings for thermodynamic parameters.
+class OpenMMSystemGeneratorFFSettings(BaseForcefieldSettings):
+    """This class is a WIP.
```
dwhswenson, line 57: does this need to be completed?

gufe/settings/models.py
```diff
@@ -3,130 +3,77 @@
...
-
-class ProtocolSettings(SettingsBaseModel, abc.ABC):
-    """Protocol-specific settings; this is a base class for protocol
-    developers to use for building any settings not included elsewhere.
-
-    """
-    ...
+    # I don't want app.HBonds or 4*unit.amu, so will string-ify those for now
+    forcefield_kwargs: dict = {
+        "constraints": "app.HBonds",
```
dwhswenson, line 65: given that this is OpenMM-specific, shouldn't this have a way to either take in the OpenMM singletons or create them from a string?

gufe/settings/models.py
```diff
@@ -3,130 +3,77 @@
...
+        "rigidWater": True,
+        "removeCMMotion": False,
+        "hydrogenMass": "4*unit.amu",
```
dwhswenson, line 68: why not actually use units here, since we should be able to now?

gufe/tests/test_custom_json.py
```diff
@@ -116,13 +123,125 @@ class TestPathCodec(CustomJSONCodingTest):
...
+
+class TestSettingsCodec(CustomJSONCodingTest):
+    def setup(self):
+        self.codec = SETTINGS_CODEC
+        self.objs = [
+            models.Settings.get_defaults(),
+        ]
+        true = True
+        null = None
+        false = False
```
dwhswenson, line 146: can these be removed / replaced where needed?

dwhswenson (changes requested):
A few small changes to make.

Two bigger issues here, which probably go beyond scope of this PR, but are things we should stay on top of:

1. If our goal is to create engine-independent settings, that's not at all what these are. These are closely tied to OpenMM. That might be fine as a first step with a major rewrite for engine-independence as a later goal, but it doesn't look like we're planning engine independence from the start.
2. We will desperately need some mechanism to distinguish the different "levels" of settings. There is currently no mechanism to encode that data in these objects, and that will be necessary for analysis.

gufe/tests/test_custom_json.py
mikemhenry: Removed!

mikemhenry:
> A few small changes to make.
> 
> Two bigger issues here, which probably go beyond scope of this PR, but are things we should stay on top of:
> 
>     1. If our goal is to create engine-independent settings, that's not at all what these are. These are closely tied to OpenMM. That might be fine as a first step with a major rewrite for engine-independence as a later goal, but it doesn't look like we're planning engine independence from the start.

Really the only openmm focused bit is the `OpenMMSystemGeneratorFFSettings` class which really is only a wrapper for [`openmmforcefields`](https://github.com/openmm/openmmforcefields#automating-force-field-management-with-systemgenerator) system generator.
Once we have another protocol implementation, we can generalize it. 


>     2. We will desperately need some mechanism to distinguish the different "levels" of settings. There is currently no mechanism to encode that data in these objects, and that will be necessary for analysis.

This is tracked with https://github.com/OpenFreeEnergy/gufe/issues/119

gufe/settings/models.py
mikemhenry: This should be fine for now! Until we get another protocol forcefield application method/use case, this should be okay.

gufe/settings/models.py
mikemhenry: I will have to look into [`openmmforcefields`](https://github.com/openmm/openmmforcefields#automating-force-field-management-with-systemgenerator) to see what we can do, I didn't want to require `openmm` for testing which is why I left it as a string.

gufe/settings/models.py
mikemhenry: I don't think `openmmforcefields` won't work with the way we have units right now, I need to make an update to the system generator for that to work.

dwhswenson (approved):
works for now... I added issue #121 to do so follow-up on a few things here, and assigned it to @mikemhenry. But those don't need to block the rest of this PR. Will merge now.

PR #111, opened 2022-12-29
Implementation of topological approach for relationships between `ProtocolDAGResult`s

dotsdl:
Implementation of the "topological" approach to relationships between `ProtocolDAGResult`s from the same `Transformation`.

- rename `extend_from` -> `extends`


gufe/protocols/protocoldag.py
```diff
@@ -85,14 +86,22 @@ class ProtocolDAGResult(GufeTokenizable, DAGMixin):
     _unit_result_mapping: dict[ProtocolUnit, list[ProtocolUnitResult]]
     _result_unit_mapping: dict[ProtocolUnitResult, ProtocolUnit]
 
+    _transformation: Optional[GufeKey]
```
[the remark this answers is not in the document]
dotsdl: I put `_transformation` here as an optional attribute so that it isn't strictly required to run a `ProtocolDAG` and get back a `ProtocolDAGResult`. I still prefer the idea that a `Protocol` developer can write and test their `Protocol` without having to invoke a `Transformation` object at all, but I recognize that this is getting to be a small difference at this point.

dotsdl:
@richardjgowers is the OE license expired?

dotsdl:
I believe this is now ready for review; please let me know if I missed anything from our discussion on Friday.

dwhswenson (changes requested):
Overall, I like the idea, but I think we'll want actual objects instead of just the gufe keys.

Since the protocol only has the *key* of the result it extends, what is the picture for how it knows where to get other results? For example, imagine a protocol that keeps running until some convergence criterion is met, but in order to analyze convergence, it needs all previous data (no checkpointing possible).

With the current setup, I don't see how it would get that information. @richardjgowers and I chatted about this today, and I think the best solution here is to actually include the object, not the key. This does require including the whole history chain, but other solutions I thought of seemed messier:

* making storage part of `Settings`
* also passing storage to `ProtocolUnit`s (and thus requiring an extra that all storage use an identical API)
* making it so the object passed here is some wrapper around both storage and the key, with a defined API to load the key if needed

dotsdl:
Thanks @dwhswenson and @richardjgowers! I've added clarifying details on docstrings and in the codebase around our use of `transformation_key` and `extends_key` as labels.

I know we also discussed changes to `Protocol.create` to consume a `Transformation` object earlier today, but in trying to implement these I realized this ends up giving us a more awkward experience than current state. 

This becomes clear when trying to refactor the test suite to accommodate this. For example, in `test_protocol.py`, getting a `ProtocolDAG` goes from:

```python
        protocol = DummyProtocol(settings=None)
        dag = protocol.create(
            stateA=solvated_ligand, stateB=vacuum_ligand, name="a dummy run",
            mapping=None,
        )
```

to
```python
        protocol = DummyProtocol(settings=None)
        transformation = Transformation(
            stateA=solvated_ligand, 
            stateB=vacuum_ligand,
            mapping=None,
            protocol=protocol)
        dag = protocol.create(
            transformation=transformation,
            name="a dummy run",
        )
```
We end up having to specify everything as before, but now need to invoke a level up (transformations) on the data model just to use `Protocol.create`. I think this doesn't actually help us in practice and feels rather contorted.

I personally don't want to pursue that route, and would prefer that be a separate PR if it's still desired by others.

gufe/protocols/protocoldag.py
```diff
@@ -43,18 +54,38 @@ def name(self):
...
+        return self._transformation_key
+
+    @property
+    def extends_key(self) -> Union[GufeKey, None]:
+        """The `GufeKey` of the `ProtocolDAGResult` this object extends.
+
+        This functions as a label, indicating where this object came from.
+        It can be used to reconstruct the set of extension relationships
+        between a collection of ProtocolDAG
+
```
richardjgowers, line 85: also needs documenting that this (and above) are optional.  Extends being `None` indicating it's not an extension, `transformation_key` being None indicating ????

richardjgowers (approved):
lgtm barring some doc fixes

gufe/protocols/protocoldag.py
```diff
@@ -78,21 +109,38 @@ class ProtocolDAGResult(GufeTokenizable, DAGMixin):
...
         `ProtocolUnitResult`'s dependencies.
+    transformation_key : Union[GufeKey, None]
+        Key of the `Transformation` that this `ProtocolDAGResult` corresponds
+        to, if applicable. This functions as a label for identifying the source
+        of this `ProtocolDAGResult`.
+    extends_key : Optional[GufeKey]
+        Key of the `ProtocolDAGResult` that this `ProtocolDAGResult` extends from.
+        This functions as a label for identifying the source of this `ProtocolDAGResult`;
+        it can be used to reconstruct the tree of extensions from a collection
+        of `ProtocolUnitResult`s.
```
dwhswenson, line 120:
```suggestion
        of `ProtocolUnitResult`\s.
```

This is RST, not Markdown. (Which also probably means that none of this will render how you expect, but this change will at least get docs to build without the warning.)

dwhswenson (approved):
lgtm, after the one fix to get RTD to stop complaining.One question: did we decide against simplifying `Protocol.create` to take  `Transformation`? Or is that for a future PR?

gufe/protocols/protocoldag.py
dotsdl: Oh thanks, I didn't realize this was a thing. Does this apply to any character following the backtick (e.g., punctuation)?

dotsdl: Oh yeah, for RST we typically need double backticks, yes?

gufe/protocols/protocoldag.py
dotsdl: Done! Thanks for this!

dotsdl:
@dwhswenson see my comment here: https://github.com/OpenFreeEnergy/gufe/pull/111#issuecomment-1414519126

I really would prefer a change to `Protocol.create` as a separate PR. I tried to pursue that line of thought but it just looked a mess in practice, and felt really odd as a choice. Perhaps someone else can take a stab at it and find a better solution in a separate PR if it's still desirable.

PR #116, opened 2023-01-30
Issue 115 lam

### LINKED ISSUE #115: Move `openfe` `LigandAtomMapping` into `gufe`?
@ijpulidos and I are working on [nonequilibrium cycling](https://github.com/choderalab/perses/pull/1066) in `perses`, and were wondering if [`LigandAtomMapping`](https://github.com/OpenFreeEnergy/openfe/blob/82a6a6db4c22cc5b30ddf26253db10130a24662f/openfe/setup/atom_mapping/ligandatommapping.py#L13) from OpenFreeEnergy/openfe made more sense here in `gufe`? 

The `AtomMapping` base class is an abstract base class, so we can't directly use that, and think it best if we avoid implementing our own child class in `perses` for the mappings used there.

richardjgowers:
fixes #115

- Moves `LigandAtomMapping` class from openfe to gufe
- Adds visualization submodule


ijpulidos:
Thanks! This is great, I already have this branch working with my NEQ protocol implementation.

I think we want to be explicit as to what we mean by `componentA_to_componentB` when creating the `LigandAtomMapping` instances. Is the key the index in componentA and the value the index in componentB or is it the other way around?

richardjgowers:
@ijpulidos great thanks.  yes `A_to_B` is the key is an index of A that maps to an (value) index in B, I think this is identical to the naming in perses?

PR #126, opened 2023-02-17
[WIP]: Proposal for a LigandNetwork class in gufe in stead of OpenFE

RiesBen:
I think we should like with ligand atom mapping move the ligandNetwork to gufe. (more consistent)


richardjgowers:
I think this got done elsewhere at some point

PR #127, opened 2023-02-17
type annotations for ProtocolResult

richardjgowers:
should return openff units


dotsdl (approved):
Looks great to me @richardjgowers! Merging!

PR #129, opened 2023-02-17
Fix #128

### LINKED ISSUE #128: `Settings` object not hashable
Our current `Settings` object does not appear to be hashable. This becomes a problem when trying to add a `Transformation` that includes a `Settings` object via its attached `Protocol` to an `AlchemicalNetwork`, as the `AlchemicalNetwork` deduplicates `Transformation`s via a frozenset.

Currently we have individual `__hash__` implementations for many `GufeTokenizable`s in `gufe`, such as `Transformation`, `ChemicalSystem` and `Protocol`. These all should probably use the same `__hash__` implementation as `GufeTokenizable` itself, and doing so should avoid the above.

This first surfaced in updating the [example notebook for building an AlchemicalNetwork](https://github.com/OpenFreeEnergy/ExampleNotebooks/pull/36#issuecomment-1433402491).

dotsdl:
Sets the `__hash__` method for each major `GufeTokenizable` to that of `GufeTokenizable` itself. This removes the redundancy of existing implementations and gives consistent behavior for these objects.

### DIFF
--- a/gufe/chemicalsystem.py
+++ b/gufe/chemicalsystem.py
@@ -1,7 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
-import abc
+from collections import abc
 from typing import Dict, Optional
 
 import numpy as np
@@ -10,7 +10,7 @@ from .tokenization import GufeTokenizable
 from .components import Component
 
 
-class ChemicalSystem(abc.Mapping, GufeTokenizable):
+class ChemicalSystem(GufeTokenizable, abc.Mapping):
     """A node of an alchemical network.
 
     Attributes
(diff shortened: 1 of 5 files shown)


dotsdl:
Closes #128.

gufe/chemicalsystem.py
```diff
@@ -10,7 +9,7 @@
 from .components import Component
 
 
-class ChemicalSystem(abc.Mapping, GufeTokenizable):
```
richardjgowers, line 13: @dotsdl was there a reason you removed the mapping-like behaviour here?

dotsdl: It wasn't clear to me that we were really using it, and it was messing with class resolution for our call to `super().__hash__()` in `__hash__`. I've re-added it and specified which parent class to `super` on in 71d83c9.

dotsdl: Switched the order to avoid having to do a `super(abc.Mapping, self).__hash__()`, and instead we keep `super().__hash__()` here.

richardjgowers (approved):

PR #130, opened 2023-02-20
Update Protocol tests to use Settings object

richardjgowers:
Updates the tests around `DummyProtocol` to use a `Settings` object internally, making the tests more relevant for real world cases

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -73,9 +73,9 @@ class ProtocolResult(GufeTokenizable):
 class Protocol(GufeTokenizable):
     """A protocol that implements an alchemical transformation.
 
-    Takes a `ProtocolSettings` object specific to the protocol on init.
-    This configures the protocol for repeated execution on `ChemicalSystem`
-    objects.
+    Takes a `Settings` object customised for this protocol on init.
+    This configures the protocol for repeated execution on (pairs of)
+    `ChemicalSystem` objects.
 
     This is an abstract base class; individual Protocol implementations
     should be subclasses of this class. The following methods should be
@@ -139,7 +139,7 @@ class Protocol(GufeTokenizable):
     def _default_settings(cls) -> Settings:
         """Method to override in custom `Protocol` subclasses.
 
-        Gives a usable instance of `ProtocolSettings` that function as
+        Gives a usable instance of `Settings` that function as
         reasonable defaults for this `Protocol` subclass.
 
         """
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -9,7 +9,7 @@ import abc
 from typing import Optional, Iterable, Any, Union
 from openff.units import Quantity
 
-from .. import Settings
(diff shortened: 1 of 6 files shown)


mikemhenry:
Really weird that only RTD check shows up here, but the CI action and the mypy action ran and passed

richardjgowers:
@mikemhenry might be because I was PRing onto a different branch than main?

mikemhenry:
Yes! That was it, we have it (correctly) set up to only run on PRs pointed at main. https://github.com/OpenFreeEnergy/gufe/blob/main/.github/workflows/ci.yaml#L5

dotsdl:
Reviewing!

gufe/protocols/protocol.py
```diff
@@ -73,9 +73,9 @@ def get_rate_of_convergence(self):
 class Protocol(GufeTokenizable):
     """A protocol that implements an alchemical transformation.
 
-    Takes a `ProtocolSettings` object specific to the protocol on init.
-    This configures the protocol for repeated execution on `ChemicalSystem`
-    objects.
+    Takes a `Settings` object customised for this protocol on init.
```
dwhswenson, line 76:
```suggestion
    Takes a ``Settings`` object customised for this protocol on init.
```

dwhswenson (approved):
Suggested some md -> rst fixes in the docstrings; otherwise LGTM.

I'm assuming updating to main will fix the CodeCov project complaints.

Also, note that `openfecli` re-uses `DummyProtocol` from here, so we'll likely need to update the saved JSON files there. There's a script to do that in `openfecli/tests/dev`.

dotsdl (approved):
It's a yes from me dawg.

gufe/protocols/protocol.py
```diff
@@ -91,23 +91,22 @@ class Protocol(GufeTokenizable):
         Corresponding `ProtocolResult` subclass
 
     """
-
+    _settings: Settings
     result_cls: type[ProtocolResult]
 
-    def __init__(self, settings=None):
-        """Create a new `Protocol` instance.
+    def __init__(self, settings: Settings):
```
richardjgowers, line 97: `Settings` now not optional, I couldn't think of a case (besides lazy test writing, which I'm fixing here) when we'd have a Settings-less Protocol

gufe/settings/models.py
```diff
@@ -85,7 +87,7 @@ class Settings(SettingsBaseModel):
     settings_version: int = 0
     forcefield_settings: OpenMMSystemGeneratorFFSettings
     thermo_settings: ThermoSettings
-    protocol_settings: Union[ProtocolSettings, None]
+    protocol_settings: type[ProtocolSettings]
```
richardjgowers, line 90: Also `Settings` must include the Protocol specific section.  Again, I couldn't think of a good example of when you'd not want to have Protocol specific settings.  If there was one, you could always define an empty dict.

richardjgowers:
@dwhswenson @dotsdl I took it all a bit further, sorry

Issue #120, opened 2023-02-03
Change `shared` from directory to an object


dwhswenson:
This is an API-breaking change that I think should be included before releasing a 1.0. We're currently assuming that `shared` is a directory. The argument I'm making is that there are reasonable approaches where this will not be valid, e.g., if `shared` is accessed securely (`scp`, `sftp`, etc.). So `shared`, as a `Protocol` author uses it, should be abstracted out.

What is needed here is a subset of the API already implemented in `externalresource`. Therefore, the code shouldn't be hard to write, and for now we can just use the existing `externalresource` objects. Eventually we may split out the smaller, required API, from aspects of `externalresource` that will only be required for `openfe`-specific storage.

PR #131, opened 2023-02-21
Storage migration: Remove from `gufe`

dwhswenson:
This removes some of the storage code, which we've determined will be more `openfe`-specific, from `gufe`.

This removes:

* `metadatastore.py`
* `resultclient.py`
* `resultserver.py`

and relevant tests.

This does *not* remove the `externalresource` structures. The intent is to reuse those as the solution to #120, although the API needs there are probable smaller than what is provided by `ExternalResource`s. Further discussion on that can go in #120.

A later PR will move/restructure imports for `ExternalResource`s, since the current nested structure is nonsense when the rest of `storage` is gone.

* Sibling PR on `openfe`: OpenFreeEnergy/openfe#266


dwhswenson:
The failed check here codecov/project, which is failing because the PR removes fully-covered code. That shouldn't be a blocker to review/merge.

PR #132, opened 2023-02-22
Add test data to MANIFEST.in

dwhswenson:
We had not been including test data in MANIFEST.in, meaning that `pytest --pyargs gufe` would fail because the files weren't including in the installation.

I know that @richardjgowers has been talking about moving to a separate package to manage test data. That may be a better solution than this, but it would be nice to push out a release with this quickly: (1) running tests is one of the first things I recommend when a user has an installation issue (can often point to problems like this); (2) I think this is blocking OpenFreeEnergy/openfe#266.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,8 @@ include versioneer.py
 include gufe/_version.py
 
 include gufe/vendor/pdb_file/data/*.xml
+recursive-include gufe/tests/data/ *.pdb
+recursive-include gufe/tests/data/ *.sdf
+recursive-include gufe/tests/data/ *.cif
+recursive-include gufe/tests/data/ *.mol2
+recursive-include gufe/tests/data/ *.json


IAlibay (approved):

IAlibay:
Maybe a CI test to build wheels, install from them, and then run CI from installed wheels would be good?

dwhswenson:
> Maybe a CI test to build wheels, install from them, and then run CI from installed wheels would be good?

This is essentially what the GitHub Actions workflows in [autorelease](https://github.com/dwhswenson/autorelease) do. The release (to pypi) process is:

1. Trigger the release process by making a PR into the branch `stable`
2. If tests pass, do a test release on test.pypi
3. Download the test release on a fresh machine; run tests (e.g., `pytest --pyargs $PACKAGE`)
4. If that set of checks passes, you merge the PR and it automatically cuts the GitHub release (using PR top post as release notes) and then deploys to PyPI.

The idea is to avoid exactly this situation. Might be a pain if we don't want to actually release on PyPI here, though (although we've parked it with what appears to by gufe 0.4 -- maybe better to actually release than to park an old release).

IAlibay:
I guess I'm just advocating for just a quick `pipx build wheel` -> `pip install <wheel>` -> `pytest --pyargs` sequence?

dwhswenson:
I guess my thought it that you really only want to do this before a release, although I suppose an extra check wouldn't hurt. 🤷‍♂️

IAlibay:
> I guess my thought it that you really only want to do this before a release, although I suppose an extra check wouldn't hurt. 🤷‍♂️ 

Yeah so basically you just want to catch out cases where someone adds a new file and doesn't update the manifest

richardjgowers:
Yeah I think I want something fancier later, but this'll do for now

Issue #119, opened 2023-02-03
gufe settings need to check compatibility with transformations 


mikemhenry:
We need to add some machinery so that when a user defines a transformation, we can ensure the two different settings objects that are included are compatible. For example, running at two different temperatures.

We should leverage the hierarchy of our settings to accomplish this. For example, we could compare a hash of the thermo settings and ensure that they are the same for both ends of the transformation.

PR #134, opened 2023-02-22
WIP: Protocol.validate API point

richardjgowers:
Not sure what the failure return should be;
- plain `False` isn't constructive
- raising an `Error` would prevent multiple failures being reported
- maybe a list of strings, each an individual error message? (so empty list is success?)
- maybe an Error, but with all the error messages concatenated?

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -85,6 +85,10 @@ class Protocol(GufeTokenizable):
     - `_gather`
     - `_default_settings`
 
+    Optionally, these methods can be overridden:
+
+    - `validate`
+
     Attributes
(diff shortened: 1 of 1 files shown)


dwhswenson:
It might be worth planning for this to eventually use [`ExceptionGroups`](https://docs.python.org/3/library/exceptions.html#exception-groups) (added in 3.11: [PEP](https://peps.python.org/pep-0654/)).

mikemhenry:
I agree with @dwhswenson so we should ideally raise exceptions and then once our min version is 3.11, we can switch to using exception groups.

richardjgowers:
see #119

PR #146, opened 2023-03-01
Make `GufeTokenizable` use a true flyweight pattern

### LINKED ISSUE #145: RDKit properties are not stored in `SmallMoleculeComponent.to_dict`
Problem illustrated below. However, there's a more fundamental issue that we were originally using RDKit properties to label the version of `gufe` that the molecule came from. So if we do include those in `to_dict`, `SmallMoleculeComponent` would need a custom gufe-key representation that excludes the gufe version properties from the hash.

My proposed solution would be to:

1. Include properties in `to_dict`
2. Remove the additional properties we add for gufe versioning.

Alternate solutions: 

* Custom `_tokenization` method for `SmallMoleculeComponent` that filters the specific properties we add 
* Drop the guarantee that we preserve properties attached to an rdkit molecule -- I'm not sure whether these properties would survive a round-trip to another molecular representation anyway (openff, oechem).

MCVE below (note that this has to be done in separate processes for the error to occur because our flyweight will reuse the same object in the same process).

```python
# dump_smc.py
import gufe
from rdkit import Chem
import json

mol = Chem.MolFromSmiles("CC")
Chem.AllChem.Compute2DCoords(mol)
mol.SetProp("foo", "bar")
smc = gufe.SmallMoleculeComponent.from_rdkit(mol)

assert smc.to_rdkit().GetProp("foo") == "bar"

dct = smc.to_dict()
with open("dump.json", mode='w') as f:
    f.write(json.dumps(dct, cls=JSON_HANDLER.encoder))
```

```python
# load_smc.py
import gufe
import json

with open("dump.json", mode='r') as f:
    dct = json.load(f)

smc = gufe.SmallMoleculeComponent.from_dict(dct)
assert smc.to_rdkit().GetProp("foo") == "bar"
```

```bash
$ python dump_smc.py
$ python load_smc.py
```

```pytb
Traceback (most recent call last):
  File "/Users/dwhs/Dropbox/omsf/tmp-rdkit-prop-issue/load_smc.py", line 9, in <module>
    assert smc.to_rdkit().GetProp("foo") == "bar"
KeyError: 'foo'
```

dwhswenson:
This ensures that only one `GufeTokenizable` exists in memory for any given gufe key.

Consider this more a proposal than a request to merge.

Significant changed behavior:

* `tokenizable.copy_with_replacements()` (with no args) used to give a distinct copy in memory. It no longer does this, since no two different objects in memory can have the same gufe key.
* several tests in `SmallMoleculeComponent` are now meaningless -- there were a bunch of tests predicated on the idea that two molecules were equal but not the same in memory. This is now testing tautologies (I had put a test in to ensure that those tests were not testing tautologies; that extra test is now failing).
* removed concept of ofe-version from SMC
* serialise the `_rdkit` mol rather than going via openff.  this should preserve Props better

fixes  #145.

This would also require changing the docs on [deduplication in memory](https://gufe.readthedocs.io/en/latest/guide/serialization.html#deduplication-in-memory).

### DIFF
--- a/gufe/tokenization.py
+++ b/gufe/tokenization.py
@@ -54,7 +54,9 @@ class _GufeTokenizableMeta(type):
     def __call__(cls, *args, **kwargs):
         instance = super().__call__(*args, **kwargs)
         # add to registry if not already present
-        TOKENIZABLE_REGISTRY.setdefault(instance.key, instance)
+        key = instance.key
+        TOKENIZABLE_REGISTRY.setdefault(key, instance)
+        instance = TOKENIZABLE_REGISTRY[key]
         return instance
 
 
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -49,6 +49,8 @@ class ExplicitMoleculeComponent(Component):
     representations. Specific file formats, such as SDF for small molecules
     or PDB for proteins, should be implemented in subclasses.
     """
+    _rdkit: Chem.Mol
+    _name: str
 
     def __init__(self, rdkit: RDKitMol, name: str = ""):
         name = _ensure_ofe_name(rdkit, name)
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -37,11 +37,6 @@ def _ensure_ofe_name(mol: RDKitMol, name: str) -> str:
     return name
 
 
-def _ensure_ofe_version(mol: RDKitMol):
-    """Ensure the rdkit representation has the current version associated"""
-    mol.SetProp("ofe-version", __version__)
-
-
 class ExplicitMoleculeComponent(Component):
     """Base class for explicit molecules.
 
@@ -54,7 +49,6 @@ class ExplicitMoleculeComponent(Component):
 
     def __init__(self, rdkit: RDKitMol, name: str = ""):
         name = _ensure_ofe_name(rdkit, name)
-        _ensure_ofe_version(rdkit)
         conformers = list(rdkit.GetConformers())
         if not conformers:
             raise ValueError("Molecule was provided with no conformers.")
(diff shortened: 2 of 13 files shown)


dwhswenson:
@richardjgowers : Pull these changes into #141 and it should fix the tests (except the 2 that are failing here)

gufe/tokenization.py
```diff
@@ -54,7 +54,9 @@ def __new__(cls, name, bases, classdict):
     def __call__(cls, *args, **kwargs):
         instance = super().__call__(*args, **kwargs)
         # add to registry if not already present
-        TOKENIZABLE_REGISTRY.setdefault(instance.key, instance)
+        key = instance.key
```
mikemhenry, line 57: You were not kidding when you said that it would take 2 lines of code to change this :rofl:

dotsdl:
I'm for this change. 

Originally we had reservations on going full [flyweight](https://python-patterns.guide/gang-of-four/flyweight/) for `GufeTokenizable` instantiation via `__init__`, since the expectation from some users may be that this always gets you a new object. The downside with this approach, as we've seen, is that duplicate objects are then allowed, and can lead to surprising consequences, with no clear benefit.

I think resolving this inconsistency is worth breaking any expectation that `__init__` gets a new, distinct object.

dotsdl (approved):
Looks great to me @dwhswenson; if we can fix the remaining failed tests, I'm for merging when satisfied.

richardjgowers (approved):

richardjgowers (changes requested):
hmm, I'm not 1,000% confident that the rdkit tobinary will be as stable in output as we need, so I might reroll that

richardjgowers (approved):

dwhswenson:
> hmm, I'm not 1,000% confident that the rdkit tobinary will be as stable in output as we need, so I might reroll that

@richardjgowers : I take it you are now 1000% confident?

richardjgowers:
Yeah wrote something manual that was similar to what we had previously, but
also solves #145 as it is rdkit centric

PR #148, opened 2023-03-04
Added scratch_basedir override to `execute_DAG`

dotsdl:
Working adjustments to accommodate work on compute services in `alchemiscale`.


gufe/protocols/protocoldag.py
```diff
@@ -355,6 +355,8 @@ def _from_dict(cls, dct: dict):
 
 def execute_DAG(protocoldag: ProtocolDAG, *,
                 shared: Optional[PathLike] = None,
+                scratch_basedir: Optional[PathLike] = None,
```
richardjgowers, line 358: just `scratch`, or change `shared` to be `shared_basedir` (or `shared_dir`)

gufe/tokenization.py
```diff
@@ -360,6 +360,15 @@ def __repr__(self):   # pragma: no cover
     def to_dict(self):
         return {':gufe-key:': str(self)}
 
+    @property
+    def prefix(self):
+        return self.split('-')[0]
+
+    @property
+    def token(self):
+        return self.split('-')[1]
```
richardjgowers, line 369: not bad, but tests

Prefix is always the classname though right?

richardjgowers (changes requested):

gufe/protocols/protocoldag.py
dotsdl: Adjusted this in our live session today. `shared` is expected to already be established by caller to be fed into this function, while `scratch_basedir` is needed to spawn scratch spaces for the `ProtocolUnit`s executed in the course of this function. These scratch spaces are cleared after each unit's execution.

gufe/tokenization.py
dotsdl: Yes, prefix is always the classname. Perhaps instead of these properties pulling apart the string representation, we hold onto these components separately by overriding `__init__`?

gufe/tests/test_protocoldag.py
```diff
@@ -0,0 +1,88 @@
...
+        my_id = inputs['identity']
+
+        with open(os.path.join(ctx.shared, f'unit_{my_id}_shared.txt'), 'w') as out:
+            out.write(f'unit {my_id} existed!\n')
+        with open(os.path.join(ctx.scratch, f'unit_{my_id}_scratch.txt'), 'w') as out:
+            out.write(f'unit {my_id} was here\n')
+
+        return {
+            'log': 'finished',
+        }
```
richardjgowers, line 24: I think this is where @dwhswenson would say these files need tagging in the output to exist

richardjgowers (commented):
Code is looking a lot tighter with the changes!

gufe/tokenization.py
mikemhenry: :pray: Yes, much more clear than remembering what `[0]` and `[1]` are

dotsdl:
@richardjgowers I think this is now ready for review!

I now have `execute_DAG` treating `shared_basedir` and `scratch_basedir` the same way, with subdirs created per `ProtocolUnit`. This expresses an idea discussed earlier today with @dwhswenson, and helps to ensure that there are no name clashes in the `shared` space between `ProtocolUnit`s. 

Under this scheme, protocol authors are forced to include paths to `shared` files they want to pass to dependent `ProtocolUnit`s in the return dict, since relying on names alone will be insufficient. I think this is a good pattern, as we wanted authors to do this anyway instead of creating "implicit" file dependencies between `ProtocolUnit`s.

dotsdl:
I don't understand why `sphinx` is throwing a warning in the docs build. I can't see any difference between the docstring of `ProtocolUnit` in this PR compared to the others inflight that it doesn't throw the warning on.

@dwhswenson any insights here?

richardjgowers (approved):

gufe/protocols/protocolunit.py
```diff
@@ -270,11 +270,10 @@ def execute(self, *,
         ----------
         context : Context
             Execution context for this `ProtocolUnit`; includes e.g. ``shared``
-            and ``scratch`` `Path`s.
+            and ``scratch`` `Path` s.
```
dotsdl, line 273: Oh there's the one you mentioned. What's puzzling to me is all our use of "`ProtocolUnit`s" in other docstrings on the same object doesn't appear to trigger sphinx's warning at all. So weird. I guess because they aren't defined yet by the time the strings are parsed?

PR #150, opened 2023-03-15
Alchemiscale compute 2

richardjgowers:
@dotsdl I tried to write tests for your PR

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -1,14 +1,13 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
-import abc
 from copy import copy
 from collections import defaultdict
 import os
 from typing import Iterable, Optional, Union, Any
 from os import PathLike
 from pathlib import Path
-import tempfile
+import shutil
 
 import networkx as nx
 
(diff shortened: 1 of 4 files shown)


gufe/protocols/protocoldag.py
```diff
@@ -402,23 +401,16 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
         # `ProtocolUnitResult`
         inputs = _pu_to_pur(unit.inputs, results)
 
-        if scratch_basedir is not None:
```
richardjgowers, line 405: @dotsdl the logic here seemed wrong.  If `scratch_basedir is not None`, then you probably don't want to wrap it as a `TemporaryDirectory` if you then want to keep it.

Also, `ProtocolUnit.execute()` **also** has logic for dealing with scratch=None, so we can omit the second branch.

dwhswenson:
Note: As you're diving into some of this, let's keep in mind that #120 tells us we need to make changes that will change the API around `shared`. I'm about to put a PR up for that.

gufe/tests/test_execute_dag.py
```diff
@@ -0,0 +1,106 @@
...
+        my_id = inputs['identity']
+
+        with open(os.path.join(ctx.shared, f'unit_{my_id}_shared.txt'), 'w') as out:
+            out.write(f'unit {my_id} existed!\n')
+        with open(os.path.join(ctx.scratch, f'unit_{my_id}_scratch.txt'), 'w') as out:
+            out.write(f'unit {my_id} was here\n')
+
+        return {
+            'log': 'finished',
+        }
```
dwhswenson, line 22: If the intent is to have a protocol unit that outputs a file to be saved, I think you'll need to include that as part of the return here.

dotsdl:
Thanks @richardjgowers! Sorry for the delay; been focused downstream on `alchemiscale` bits that use this, but will loop back around to this once I get things tied off more there.

dotsdl:
Merging this into #148 to hammer on in one place. Thanks @richardjgowers!

PR #151, opened 2023-03-15
Make `shared` into an `ExternalStorage` object

### LINKED ISSUE #120: Change `shared` from directory to an object

dwhswenson:
Resolves #120.

I'm not 100% sure that this is the right way to do it, but it is the best thing I've come up with. A couple major points:

1. As said in the title, this uses our existing `ExternalStorage` for `shared`. That API is probably bigger than we need (I think we only need `store_path` and `get_filename`), but we already have it in place.
2. This now takes all `pathlib.Path` outputs from the dict returned by `ProtocolUnit._execute` and ensures that they are saved to `shared`.
3. The label used in `shared.store_path` is worth discussion. We've previously decided that permanent storage does not use a labelling system in common with the protocol system, so I've opted for a very simple label here. The downside is that we'll need to copy files around a lot.

This is a significant API break. `ProtocoUnit`s that use files generated by other units will need to get the filename with `shared.get_filename(label)`.

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -354,10 +354,11 @@ class ProtocolDAG(GufeTokenizable, DAGMixin):
         return cls(**dct)
 
 
-def execute_DAG(protocoldag: ProtocolDAG, *,
-                shared: Optional[ExternalStorage] = None,
-                raise_error: bool = True,
-                ) -> ProtocolDAGResult:
+def execute_DAG(
+    protocoldag: ProtocolDAG, *,
+    shared: Optional[Union[ExternalStorage, PathLike, str]] = None,
+    raise_error: bool = True,
+) -> ProtocolDAGResult:
     """Execute the full DAG in-serial, in process.
 
     This is intended for debug use for Protocol developers.
@@ -367,7 +368,7 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
     ----------
     protocoldag : ProtocolDAG
         The `ProtocolDAG` to execute.
-    shared : Optiona[ExternalStorage]
+    shared : Optional[ExternalStorage]
        Storage space that persists across whole DAG execution, but may
        be removed after. Used by some `ProtocolUnit`s to pass file
        contents to dependent `ProtocolUnit` objects.
@@ -385,6 +386,9 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
     """
     if shared is None:
         shared = FileStorage(os.getcwd())
+
+    if isinstance(shared, (PathLike, str)):
+        shared = FileStorage(shared)
 
     # iterate in DAG order
     results: dict[GufeKey, ProtocolUnitResult] = {}
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -16,6 +16,7 @@ from ..tokenization import GufeTokenizable, GufeKey
 from .protocolunit import (
     ProtocolUnit, ProtocolUnitResult, ProtocolUnitFailure,
 )
+from ..storage import ExternalStorage, FileStorage
 
 
 class DAGMixin:
@@ -354,7 +355,7 @@ class ProtocolDAG(GufeTokenizable, DAGMixin):
 
 
 def execute_DAG(protocoldag: ProtocolDAG, *,
-                shared: Optional[PathLike] = None,
+                shared: Optional[ExternalStorage] = None,
                 raise_error: bool = True,
                 ) -> ProtocolDAGResult:
     """Execute the full DAG in-serial, in process.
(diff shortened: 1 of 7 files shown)


gufe/protocols/protocolunit.py
```diff
@@ -302,22 +304,27 @@ def execute(self, *,
...
             )
 
-        # TODO: change this part once we have clearer ideas on how to inject
-        # persistent storage use
+        output_files = [f for f in outputs.values() if isinstance(f, Path)]
+        for path in output_files:
+            # label is worth further consideration before merge
+            label = str(self.key / path)
+            shared.store_path(label, path)
+
```
richardjgowers, line 327: so this is scanning the outputs for files that are indicated in the outputs, then explicitly storing them in the `ExternalResource`? if so this deserves a comment to explain that

then also a test to check that this is working

gufe/protocols/protocolunit.py
```diff
@@ -22,6 +22,8 @@
     GufeTokenizable, GufeKey, TOKENIZABLE_REGISTRY
 )
 
+from ..storage import ExternalStorage
+
 
 @dataclass
 class Context:
```
richardjgowers, line 29: since we're tinkering here, what's the point of the `Context` object here?  it converts two arguments to one, and we use it once?  can we remove it and just have two `ExternalStorage` arguments (shared/scratch) everywhere?

gufe/protocols/protocoldag.py
```diff
@@ -366,10 +367,10 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
     ----------
     protocoldag : ProtocolDAG
         The `ProtocolDAG` to execute.
-    shared : Optional[PathLike]
-       Path to scratch space that persists across whole DAG execution, but
-       is removed after. Used by some `ProtocolUnit`s to pass file contents
-       to dependent `ProtocolUnit`s.
+    shared : Optiona[ExternalStorage]
```
richardjgowers, line 370: typo on `Optional`

gufe/protocols/protocoldag.py
```diff
@@ -354,7 +355,7 @@ def _from_dict(cls, dct: dict):
 
 
 def execute_DAG(protocoldag: ProtocolDAG, *,
-                shared: Optional[PathLike] = None,
+                shared: Optional[ExternalStorage] = None,
```
richardjgowers, line 358: could we make this `Optional[Union[PathLike, ExternalStorage]]` so it "just worked" for people that don't want to have to figure out the `ExternalStorage` concept.  I'm just seeing this as another step on the quickstart learning curve.  Internally we'd just immediately convert to `FileStorage` and all is well right?

gufe/protocols/protocolunit.py
```diff
@@ -30,7 +32,7 @@ class Context:
 
     """
     scratch: PathLike
-    shared: PathLike
+    shared: ExternalStorage
```
richardjgowers, line 35: may as well make both `ExternalStorage` at this point

richardjgowers (changes requested):

gufe/protocols/protocolunit.py
dwhswenson: I'm fine either way here.

@dotsdl : is there a reason we need the `Context` object?

gufe/protocols/protocolunit.py
dwhswenson: Disagree. I think `scratch` is inherently a local directory. There is never a need for it to be anything else (a protocol unit must be run on a single node). Also, I think that requiring contributors to use `get_filename` for every file in their protocol is going to be tedious, especially when they are porting existing code to be gufe-compatible.

Current model is that they use `get_filename` to obtain files that are inputs to the protocol. We might be able to simplify that, though, so they don't even need that. Writing output is done because the outputs are identified from `pathlib.Path` objects in the return dict from `_execute`.

dwhswenson:
This should be ready for another look, although I think we're going to talk about it on Thursday anyway.

gufe/protocols/protocolunit.py
```diff
@@ -302,22 +304,30 @@ def execute(self, *,
...
-        # TODO: change this part once we have clearer ideas on how to inject
-        # persistent storage use
+        # extract any files from the values of the dict that _execute
+        # returns; ensure that we move these to shared so they can be used
+        # by later units
+        for label, path in outputs.items():
+            is_file = (isinstance(path, Path) and path.exists()
+                       and not path.is_dir())
+            if is_file:
+                shared.store_path(f"{self.key}/{label}", path)
```
richardjgowers, line 329: We discussed and didn't quite decide on if outputs should explicitly be listed to be copied up, or instead the shared dir passed to _execute can just be walked and blindly copied up. It might come down to if there's cases where items are required in shared (for archiving) but never used as inputs for future Units

richardjgowers (commented):
This doesn't seem to conflict with @dotsdl 's neighbouring PR

richardjgowers (changes requested):
this needs a rebase to accommodate changes to execute_DAG around scratch etc

richardjgowers:
@dwhswenson this is obsolete right?

PR #152, opened 2023-03-17
Settings revision

richardjgowers:


gufe/tests/test_protocol.py
```diff
@@ -67,7 +67,7 @@ def _execute(ctx, *, simulations, **inputs):
         )
 
 
-class DummySpecificSettings(settings.ProtocolSettings):
+class DummySpecificSettings(settings.Settings):
```
richardjgowers, line 70: @dotsdl @ijpulidos this is the key change here.  Rather than the protocol settings being inside `settings.Settings` , the protocol settings **are a subclass** of `settings.Settings`

gufe/settings/models.py
```diff
@@ -56,38 +57,57 @@ class BaseForcefieldSettings(SettingsBaseModel, abc.ABC):
...
 
-    # I don't want app.HBonds or 4*unit.amu, so will string-ify those for now
-    forcefield_kwargs: dict = {
-        "constraints": "app.HBonds",
-        "rigidWater": True,
-        "removeCMMotion": False,
-        "hydrogenMass": "4*unit.amu",
-    }
-    forcefields: list = [
+    forcefields: list[str] = [
```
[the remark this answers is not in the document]
ijpulidos: I like this approach, it makes it less cryptic as to what these fields are, in my opinion. Thanks.

gufe/tests/test_protocol.py
ijpulidos: I think I was using this approach before in the NEQ cycling protocol, but I now can't remember what were the caveats or limitations using this. I will double check how these changes affect the protocol as soon as possible.

Is the idea here that we will then need just to instantiate a `ProtocolSettings` object and this will have all the `Settings` fields as well? This sounds simpler if that's the case.

richardjgowers: @ijpulidos yes.  So before you were defining `NonEqCyclingSettings` and putting that inside a `settings.Settings` object as the `protocol_settings=` argument.

so before:
```python
class NonEqCyclingSettings(settings.ProtocolSettings):
    eq_steps: int
    ...

class NonEqProtocol:
    @classmethod
    def default_settings(cls):
        return settings.Settings(..., protocol_settings=NonEqCyclingSettings(eq_steps=2500))
```

and with this PR: (the custom settings class is inheriting from something else now) 
```python
class NonEqCyclingSettings(settings.Settings):
    eq_steps: int
    ...

class NonEqProtocol:
    @classmethod
    def default_settings(cls):
        return NonEqCyclingSettings(eq_steps=2500)
```

this should remove a level of indentation when you're modifying settings specific to the protocol.

IAlibay (approved):
This seems like a good direction ahead for settings.

PR #153, opened 2023-03-18
Optional offtk and remove oechem

richardjgowers:


.github/workflows/ci.yaml
```diff
@@ -44,14 +44,6 @@ jobs:
             environment-file: environment.yml
             activate-environment: gufe
 
-      - name: Decrypt OpenEye license
```
IAlibay, line 47: Don't you still want to test with the optional dependency?

IAlibay: nvm I'm missing the point

PR #157, opened 2023-03-21
Clarify docstring for componentA_to_componentB

### LINKED ISSUE #155: Specifying component B to A in ligand mapping
Currently the `LigandAtomMapping` class has a `componentA_to_componentB` parameter that we can use to define the mapping. Some packages or tools might expect a different "directionality" in the mapping, for example [in perses](https://github.com/choderalab/perses/blob/0355528d2332d47d0b1a396785321ada9f6d6fc4/perses/rjmc/topology_proposal.py#L2030) is inverted compared to the convention chosen by `gufe` (hopefully for convenience, but maybe it was just arbitrary on the perses side).

Do we want to have a `componentB_to_componentA` option for the constructor of a `LigandAtomMapping` object? That said, I'd think that we want at least to have it unambiguously defined in the docstring, even for the current parameter.

dwhswenson:
@ijpulidos: Does this clarify sufficiently?

If so, resolves #155.

### DIFF
--- a/gufe/mapping/atom_mapping.py
+++ b/gufe/mapping/atom_mapping.py
@@ -30,6 +30,9 @@ class AtomMapping(ComponentMapping, abc.ABC):
     def componentA_to_componentB(self) -> Mapping[int, int]:
         """Maps the index of an item from Component A onto Component B
 
+        Keys are indices from Component A, values are indices from Component
+        B.
+
         Not all indices will be resolvable, these items have no corresponding
         entity in the other component (e.g. the atom disappears), therefore
         resulting in a KeyError on query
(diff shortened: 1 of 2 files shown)


ijpulidos:
Perfect, thanks!

ijpulidos:
Oh wait, maybe we also want to document it this way in https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/mapping/ligandatommapping.py#L33

dwhswenson:
> Oh wait, maybe we also want to document it this way in https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/mapping/ligandatommapping.py#L33

Good catch -- added.

richardjgowers (approved):

PR #161, opened 2023-03-29
Clarify / fix some of the settings documentation

IAlibay:
Small things, but if we're going to inherit these downstream, we should make sure they are as clean as can be.


IAlibay:
Also not to re-ignite this debate, but "force field" > "forcefield", otherwise it'd be OpenF instead of OpenFF.

richardjgowers (approved):

PR #163, opened 2023-04-05
A hill I'm ready to die upon [nomenclature changes]

IAlibay:
Last I heard it was  OpenFF not OpenF 🙃

### DIFF
--- a/gufe/settings/__init__.py
+++ b/gufe/settings/__init__.py
@@ -3,7 +3,7 @@
 from .models import (
     Settings,
     ThermoSettings,
-    BaseForcefieldSettings,
+    BaseForceFieldSettings,
     OpenMMSystemGeneratorFFSettings,
     SettingsBaseModel,
 )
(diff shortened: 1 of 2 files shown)


IAlibay:
🤦🏽 this goes upstream doesn't it

mikemhenry:
Is this a joke?

dwhswenson (approved):
I don't want @IAlibay to die on a hill!

I mean, I don't particularly want @IAlibay to die at all, but definitely not on a hill.

(Also, yes, this should be correct for consistency.)

mikemhenry (approved):

mikemhenry:
Depending on how you install gufe @ijpulidos you will need to update once this is merged in.

IAlibay:
It _shouldn't_ affect downstream unless you directly access the base class?

richardjgowers:
Normally a class name change would cause some hashes to change, but since this is a base class this won't break anything.

richardjgowers:
@IAlibay if you really want this hill, shouldn't we use `force_field_settings` not `forcefield_settings`?

IAlibay:
technically yeah, but that'll actually make downstream worse imho

PR #165, opened 2023-04-05
Just checking version import

IAlibay:
In this PR:
- Check version import in CI
- fixes universal wheel

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,9 +43,6 @@ license-files = ["LICENSE"]
 [tool.setuptools.packages]
 find = {namespaces = false}
 
-[tool.distutils.bdist_wheel]
-universal = 1
-
 [tool.mypy]
 files = "gufe"
 ignore_missing_imports = true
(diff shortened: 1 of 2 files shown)


IAlibay:
@mikemhenry the version installs as `0.0.0` - this seems weird? (https://github.com/OpenFreeEnergy/gufe/actions/runs/4621160585/jobs/8172233914?pr=165)

IAlibay:
nevermind, this is a checkout depth issue

pyproject.toml
```diff
@@ -43,9 +43,6 @@ license-files = ["LICENSE"]
 [tool.setuptools.packages]
 find = {namespaces = false}
 
-[tool.distutils.bdist_wheel]
```
IAlibay, line 46: @mikemhenry - universal bdist_wheel entries are for pure Python projects that are both py2 and py3 compliant (you can see the effect when you build a wheel, it'll output "py2py3" in the name)

mikemhenry:
And yes, I have it setup to use 0.0.0 if I can't find tag info instead of error out, so this behavior is expected. A depth of 0 should be fine since our repo is small

mikemhenry (approved):

mikemhenry:
Also @IAlibay would you mind doing this for openfe as well?

IAlibay:
yup will do

PR #169, opened 2023-04-06
Readonly settings

richardjgowers:
Adds `Settings.frozen_copy` and `Settings.unfrozen_copy` to create immutable/mutable versions of Settings object.  These freezes are recursive and apply to all contained Settings too.

Protocols now freeze their Settings on creation.  This should stop mistakes around modifying a Protocol's Settings in place.

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -112,7 +112,7 @@ class Protocol(GufeTokenizable):
 
     @classmethod
     @abc.abstractmethod
-    def _default_settings(cls) -> type[Settings]:
+    def _default_settings(cls) -> Settings:
         """Method to override in custom `Protocol` subclasses.
 
         Gives a usable instance of ``Settings`` that function as
@@ -122,7 +122,7 @@ class Protocol(GufeTokenizable):
         raise NotImplementedError()
 
     @classmethod
-    def default_settings(cls) -> type[Settings]:
+    def default_settings(cls) -> Settings:
         """Get the default settings for this `Protocol`.
 
         These can be modified and passed in as the `settings` for a new
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -103,7 +103,7 @@ class Protocol(GufeTokenizable):
         make_readonly(self._settings)
 
     @property
-    def settings(self) -> type[Settings]:
+    def settings(self) -> Settings:
         """The full settings for this ``Protocol`` instance.  This is read only"""
         return self._settings
 
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -92,12 +92,12 @@ class Protocol(GufeTokenizable):
             The full settings for this ``Protocol`` instance.  Must be passed an instance of Settings or a
             subclass which is specialised for a particular Protocol
         """
-        self._settings = settings.copy(deep=True)
+        self._settings = settings.frozen_copy()
 
     @property
     def settings(self) -> Settings:
         """A copy of the full settings for this ``Protocol`` instance.  This is read only"""
-        return self._settings.copy(deep=True)
+        return self._settings.frozen_copy()
 
     @classmethod
     def _defaults(cls):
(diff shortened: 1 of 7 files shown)


richardjgowers:
This seems to be a terrible idea... I'm getting weird errors where other Settings objects now think they're readonly when they're not.

dwhswenson (changes requested):
I think we should use the pydantic functionality to make settings immutable. That way we raise an error if a user mutates something.

I think that 

```python
>>> protocol.settings.mysetting
'foo'
>>> protocol.settings.mysetting = "bar"
>>> protocol.settings.mysetting
'foo'
```

is much riskier for users than

```python
>>> protocol.settings.mysetting
'foo'
>>> protocol.settings.mysetting = "bar"
TypeError: 'Settings' is immutable and does not support type assignment
```

richardjgowers:
@dwhswenson I tried that originally.

I had this in place:

https://github.com/OpenFreeEnergy/gufe/pull/169/commits/f07790f0e0c5c8823c7eea42dd135a4888ea0d82

But you ended up with very strange behaviour, where "brand new" Settings objects were readonly, see here:

https://github.com/OpenFreeEnergy/gufe/actions/runs/4628556994/jobs/8187777193#step:7:886

I never quite figured it out, but it felt like the readonly was applying to **all** instances of the class, rather than a given instance of the class

richardjgowers:
@dwhswenson thanks for the improvement.  I'm waiting on merging this as it causes a few headaches downstream in openfe.  In `TestResultClient` I get a lot of: `E       TypeError: Object of type DummySpecificSettings is not JSON serializable` from within `self._store_gufe_tokenizable` and I've not had time to dig through and see what's happening there.

richardjgowers:
todo: add unfreeze

gufe/tests/test_alchemicalnetwork.py
```diff
@@ -13,8 +13,8 @@
 class TestAlchemicalNetwork(GufeTokenizableTestsMixin):
 
     cls = AlchemicalNetwork
-    key = "AlchemicalNetwork-8c6df17d7ecf5902e2e338984cc11140"
-    repr = "<AlchemicalNetwork-8c6df17d7ecf5902e2e338984cc11140>"
+    key = "AlchemicalNetwork-d1035e11493ca60ff7bac5171eddfee3"
```
dwhswenson, line 16: Just checking: is this changing the gufe key because you've changed the fixture to use non-`None` settings? (Otherwise, nothing in this PR should affect gufe keys, right?)

gufe/tests/test_models.py
```diff
@@ -53,3 +53,45 @@ def test_invalid_constraint(value, good):
     else:
         with pytest.raises(ValueError):
             _ = OpenMMSystemGeneratorFFSettings(constraints=value)
+
+
+class TestFreezing:
```
dwhswenson, line 58: Current tests check that it works we do `settings.subsettings.item = ...`. Could you add a test that does `settings.subsettings = ...` ? (example: you attempt to replace the entire `thermo_settings` in one go).

dwhswenson (changes requested):
Mostly LGTM -- requesting small additional tests that aren't testing nested behavior, mainly so that no future refactors accidentally break things.

gufe/tests/test_alchemicalnetwork.py
richardjgowers: yep it's this

gufe/protocols/protocol.py
```diff
@@ -89,14 +89,15 @@ def __init__(self, settings: Settings):
...
         """
-        self._settings = settings
+        self._settings = settings.frozen_copy()
 
     @property
     def settings(self) -> Settings:
-        """The full settings for this ``Protocol`` instance."""
-        return self._settings
+        """A copy of the full settings for this ``Protocol`` instance.  This is read only"""
+        return self._settings.frozen_copy()
```
richardjgowers, line 100: this is probably overzealous and causes a copy on each access right?

dwhswenson (approved):
With the extra test added, LGTM!

PR #171, opened 2023-04-07
Add intersphinx and speed up builds

mikemhenry:


mikemhenry:
Still need to actually add intersphinx

mikemhenry:
Build time with this pr: Build took 139 seconds
Build time on main: Build took 235 seconds 

I'll do the same for openfe.

PR #172, opened 2023-04-07
Docs cleanup

dwhswenson:
This does a little cleanup, mainly things that fix up RST (missing escape to fix plurals; lack of roles defined).

This will fix some issues in openfreeenergy/openfe#325.

### DIFF
--- a/gufe/network.py
+++ b/gufe/network.py
@@ -11,14 +11,19 @@ from .transformations import Transformation
 
 
 class AlchemicalNetwork(GufeTokenizable):
-    """A network of `ChemicalSystem`s as nodes, `Transformation`s as edges.
+    """A network with all the information needed for a simulation campaign.
+
+    Nodes are :class:`.ChemicalSystem`\ s and edges are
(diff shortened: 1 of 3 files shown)


mikemhenry (approved):

gufe/network.py
```diff
@@ -11,14 +11,19 @@
...
+
+    Nodes are :class:`.ChemicalSystem`\ s and edges are
+    :class:`.Transformation`\ s.
 
     Attributes
     ----------
     edges : FrozenSet[Transformation]
-        The edges of the network, given as a `frozenset` of `Transformation`s.
+        The edges of the network, given as a ``frozenset`` of
+        :class:`.Transformation`\ s.
```
richardjgowers, line 23: I think somewhere there's a compromise between it being rendered nicely in docs and the notebook inspection docs being readable, and we're losing here.  This seems like the notebook docs got a lot worse for a link to `Transformation` that was already given above.  I don't think every single mention of a class needs to be a cross reference.

PR #173, opened 2023-04-10
Package test on CI

IAlibay:
Included in this PR:
  - package build & test on CI
  - daily cron conda install & test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
       - main
   schedule:
     # At 07:00 UTC on Monday and Thursday.
-    - cron: "0 7 * * 1,4"
+    - cron: "0 7 * * *"
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -97,7 +97,7 @@ jobs:
         run: pip install pipx wheel twine readme-renderer
 
       - name: "build sdistt"
-        run: pipx ruun build --sdist --outidr dist
+        run: pipx run build --sdist --outidr dist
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -8,7 +8,7 @@ on:
       - main
   schedule:
     # At 07:00 UTC on Monday and Thursday.
-    - cron: "0 7 * * 1,4"
+    - cron: "0 7 * * *"
```
IAlibay, line 11: As per today's discussion - run cron nightly now

.github/workflows/conda_cron.yaml
```diff
@@ -0,0 +1,54 @@
...
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  condacheck:
```
IAlibay, line 16: This should hopefully meet everything on your wishlist @richardjgowers - including a rough "raise an issue if it fails" (which we can iterate though / refine at a later point if folks want dashboards, etc...).

dwhswenson (approved):
LGTM for now (caught one unimportant typo).

I think we may iterate on the issue raising script in the future (e.g, would love to have it link to the failed run, if we can figure that out) but this will already be useful as is.

IAlibay:
> LGTM for now (caught one unimportant typo).
> 
> I think we may iterate on the issue raising script in the future (e.g, would love to have it link to the failed run, if we can figure that out) but this will already be useful as is.

So linking to the fail run is doable (we do so here: https://github.com/MDAnalysis/mdanalysis/blob/develop/maintainer/ci/darker-outcomes.py#L64), I just didn't want to make this complicated until we're really bought into the solution.

PR #174, opened 2023-04-10
remove box_vector from ChemicalSystem

richardjgowers:
was causing trouble in eq situations


dwhswenson (approved):

dotsdl:
Not used in `perses` `NonEquilibriumCyclingProtocol` as far as I can tell; should be good?

IAlibay:
It's going to affect hashes downstream afaik, so we'll have to cycle the `openfe` CI.

PR #177, opened 2023-04-10
Improved readonly settings

dwhswenson:
PR into the changes already in #169; @richardjgowers's readonly settings. Now raises `TypeError` if you try to mutate after settings get attached to a protocol.

Might be nice for something with more pydantic experience than me to look this over? But tests pass locally, at least.

### DIFF
--- a/gufe/settings/models.py
+++ b/gufe/settings/models.py
@@ -12,15 +12,52 @@ from openff.models.models import DefaultModel
 from openff.models.types import FloatQuantity
 from openff.units import unit
 from pydantic import Extra, Field, PositiveFloat
+from pydantic import PrivateAttr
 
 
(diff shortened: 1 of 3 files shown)


gufe/settings/models.py
```diff
@@ -12,16 +12,53 @@
...
+
+            if not model._is_frozen:
+                model._is_frozen = True
+
+        freeze_model(copied)
+        return copied
+
+    def __setattr__(self, name, value):
+        if self._is_frozen:
+            raise TypeError(f"Cannot set '{name}': Settings are immutable "
```
IAlibay, line 46: For the sake of your regular non-programmer user, I'd recommend toning down the word immutable to "can't be changed" or something?

PR #178, opened 2023-04-10
Delete cache from closed PRs

mikemhenry:
We only get 10gb of cache per repo, and conda envs are chonks so we are close to the limit. When we hit the limit, the oldest entry gets evicted which could cause some cache thrashing. This PR at least will delete cache entries we don't need anymore.


IAlibay (approved):
lgtm - pretty much follows the docs

PR #182, opened 2023-04-24
Change default hydrogen mass to 3.0 amu

### LINKED ISSUE #181: Reconsider default hydrogen_mass setting
Originally we used to run a hydrogen mass of 3 amu with a 4 fs timestep. Since we've switched to using the gufe settings, this has changed to 4 amu. Coincidentally I'm also seeing an increase in sporadic early simulation crashes.

It's possible that the two aren't related (could also be because we halved the number of minimization steps by default), but it's something worth keeping an eye out for. Hopefully post-workshop we can use something like alchemiscale to get to the bottom of this.

IAlibay:
Fixes #181 

This will change defaujlt settings in downstream protocols, so pinging @ijpulidos and @dotsdl too.

### DIFF
--- a/gufe/settings/models.py
+++ b/gufe/settings/models.py
@@ -67,9 +67,9 @@ class OpenMMSystemGeneratorFFSettings(BaseForceFieldSettings):
 
     rigid_water: bool = True
     remove_com: bool = False
-    hydrogen_mass: float = 4.0
+    hydrogen_mass: float = 3.0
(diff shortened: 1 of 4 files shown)


IAlibay:
An unreasonable amount of jobs failed in the initial alchemiscale hydration free energy run - it is very likely that this is the source of current problems.

richardjgowers (approved):

mikemhenry (approved):

PR #186, opened 2023-04-26
PathLike wrapper/cache for ExternalStorage

dwhswenson:
This is a start to solving #120. We've identified the following goals:

1. We want to be able to use an arbitrary external backend to `shared` (as well as `permanent`), which we already have via `ExternalStorage`, instead of it needing to be a directory on the same host.
2. We want the user experience to mirror conveniences of `pathlib.Path`, which will (we hope) minimize frustration with the transition for users.

So far, this PR doesn't integrate that into any of the execution process. That can be either part of this PR or a future one.

The basic approach here involves 2 classes:

* `SharedRoot`: represents the root directory for a given unit. By default, this will exist at `$SCRATCH/.holding/$UNIT_LABEL`. This is where most of the actual logic is. It contains references to all the paths that have been registered with it. On deletion (or on demand) this transfers all paths registered with it to the `ExternalStorage`.
* `SharedPath`: represents any path within `SharedRoot`. Contains information on the actual file path on the system, as well as the label used as the key for the external key-value store. Creation is typically from `root / "some_path"` or `some_path / "subpath"`. A `SharedPath` can be used like any other `PathLike` in `open`, and can be converted via `pathlib.Path(shared_path)` to get all the functionality of pathlib.

As I'm opening this, code is still in progress. But at least the tests should give a picture of usage; please take a look!

---

Update for some final to-do items.

- [x] Remove `prefix`; I believe it is no longer used
- [ ] Reconsider when `dag_label` needs to be given
- [x] Switch `delete_empty_dirs` to `keep_empty_dirs` (consistent with other `keep_*`)
- [x] Check special cases where a directory gets registered as a StagedPath (due to `__fspath__` getting called)
- [x] `StagingDirectory` => `StagingRegistry`
- [ ] Re-read docs and comments for accuracy
- [x] Make `StagingRegistry` objects non-FileLike (potential footgun on `os.path.join(context.shared, "filename.txt")`).


gufe/storage/pseudodirectory.py
```diff
@@ -0,0 +1,156 @@
...
+        temporarily stored; default is '.holding'. This must be the same for
+        all units within a DAG.
+    delete_holding : bool
+        whether to delete the contents of the $SCRATCH/$HOLDING/$PREFIX
+        directory when this object is deleted
+    read_only : bool
+        write to prevent NEW files from being written within this shared
+        directory. NOTE: This will not prevent overwrite of existing files,
+        in scratch space, but it will prevent changed files from uploading
+        to the external storage.
```
dwhswenson, line 30: This `read_only` is mainly to prevent the following mistake:

```python
shared = root.get_other_shared(other_unit_prefix)
with open(shared / "oldfile.txt") as f:
    ... # good so far!

with open(shared  / "newfile.txt", mode='wb') as f:
    ... # uh-oh -- now we're writing into the wrong unit's space!
```

However, it *won't* stop `open(shared / "oldfile.txt", mode='wb')`. You can still overwrite that data. If that data is in a separate permanent storage, it will be fine -- only overwrite locally. (If you set `shared` up such that the directory it uses is identical to the internal holding cache here, of course, it would overwrite that.)

gufe/storage/pseudodirectory.py
```diff
@@ -0,0 +1,156 @@
...
+            else:
+                logging.info(f"Transfering {path} to external storage")
+                self.external.store_path(obj.label, path)
+
+    def __del__(self):
+        # take everything in self.shared_dir and write to it shared; keeping
+        # our prefix
+        self.transfer_holding_to_external()
+        if self.delete_holding:
+            shutil.rmtree(self.shared_dir)
```
richardjgowers, line 99: iirc `__del__` isn't called on `del obj` but instead when garbage collection happens.  Maybe instead make `transfer_holding_to_external` always explicit and don't rely on garbage collection related things

gufe/storage/pseudodirectory.py
```diff
@@ -0,0 +1,156 @@
+from typing import Union, Optional
+from pathlib import Path
+from os import PathLike
+from .externalresource import ExternalStorage
+
+import logging
+_logger = logging.getLogger(__name__)
+
+class SharedRoot:
```
richardjgowers, line 9: comments, it's not obvious what a `SharedRoot` is etc

gufe/storage/pseudodirectory.py
```diff
@@ -0,0 +1,156 @@
...
+                self.external.store_path(obj.label, path)
+
+    def __del__(self):
+        # take everything in self.shared_dir and write to it shared; keeping
+        # our prefix
+        self.transfer_holding_to_external()
+        if self.delete_holding:
+            shutil.rmtree(self.shared_dir)
+
+    def register_path(self, shared_path):
```
richardjgowers, line 101: docs, what does register path mean etc?

gufe/storage/pseudodirectory.py
```diff
@@ -0,0 +1,156 @@
...
+        return SharedPath(root=self, path=path)
+
+    def __fspath__(self):
+        return str(self.shared_dir)
+
+    def __repr__(self):
+        return f"SharedRoot({self.scratch}, {self.external}, {self.prefix})"
+
+
+class SharedPath:
```
richardjgowers, line 134: why is this necessary?  Could `SharedRoot` just hand out real pathlikes?  Is this class just to avoid literally giving out Paths?

dwhswenson:
```python
filepath = root / "directory" / "file.txt"
```

if root/directory is a `pathlib.Path`, then root/directory is registered, but root/directory/file.txt is not. So we need the purpose is to use our `__truediv__` operator and register with the root.

gufe/storage/pseudodirectory.py
dwhswenson: works for me ... it's an executor-level thing anyway, and I don't mind asking the executor writer to think about how files get moved around (don't want to ask the protocol writer to think about that)

gufe/storage/pseudodirectory.py
dwhswenson: I'm up for name changes if there's a suggestion. I don't particularly like these names -- I originally was calling `SharedRoot` a `PseudoDirectory`, but I thought that would be more confusing.

richardjgowers: I don't mind what it's called, but it'll need a one liner to explain what the class is/does either way

gufe/storage/pseudodirectory.py
dwhswenson: This started out as a context manager. I switched it away because it makes actual usage kind of weird -- the entirety of the execution block where this is used gets indented by one level, which I don't particularly like. It gets worse if you have one of these for `shared` and one for `permanent` -- there's a syntax for setting multiple context managers at once, but I've also always found that kind of weird.

But I can switch it back to a context manager if you want 🤷‍♂️

dwhswenson:
Still in progress, but here's a quick update to what is included in this PR now. Before I dive any further into writing unit tests, take a quick look, @richardjgowers (at least at the contents of this comment).

Main idea: We have "staging" directories that are local. From a protocol author's perspective, `shared`, `permanent`, and `scratch` are `PathLike`s. However, the paths pointed to by `shared` and `permanent` are in the local scratch space, and the files there are transferred to external storage as part of the executor.

I've expanded this PR to include some tools for handling the entirety of the scratch / shared / permanent storage lifecycle. There are two context managers: one for the unit level that provides the `PathLike`s that are used by the protocol author, and which handles the cleanup at the end of that level, and one for the dag level that provides the thing that creates the unit-level context manager, and which cleans up at the dag level.

Here's an example from a protocol author's viewpoint (kind of faking our real units, but this gives the rough picture):

https://github.com/OpenFreeEnergy/gufe/blob/ed5e83c23e3ee03534f8cbe5af753cc4ac9baeae/gufe/tests/storage/test_storagemanager.py#L27-L53

For the executor, the basic picture is something like this:

```python
manager = StorageManager(scratch, shared_external, permanent_external)
with manager.running_dag(dag_label) as dag_ctx:
    for unit in ordered_dag_units:
        with dag_ctx.running_unit(unit.key) as (scratch, shared, permanent):
            unit_result = unit.run(scratch, shared, permanent)  # or however we run the unit
        # when this context ends, scratch will be deleted, shared external storage will be
        # updated, and shared holding will be deleted
# when this context ends, permanent external storage will be updated, and permanent holding 
# will be deleted
```

For much more detail, see this lifecycle test:

https://github.com/OpenFreeEnergy/gufe/blob/ed5e83c23e3ee03534f8cbe5af753cc4ac9baeae/gufe/tests/storage/test_storagemanager.py#L59-L131

Next tests will be to check the special cases that `shared` or `permanent` represent the same directories as the staging directory, or each other.

dotsdl:
From power hour:

```
         with shared.other_shared("unit1") as prev_shared: 
             with open(prev_shared / "bar.txt", mode='r') as f: 
                 bar = f.read() 
 ```

Instead of having `shared.other_shared("unit1")`, which would allow a e.g. `unitX` to access `unitA`'s shared space even if `unitA` wasn't an explicit dependency of `unitX`.

Instead, making `unitX` access `unitA`'s shared via its `ProtocolDAGResult` object passed into `run` would avoid this issue.

**Comment by dwhswenson** [comment]
@richardjgowers @dotsdl : This is pretty much ready for an initial review pass. There are a few lines that need coverage, but most of them are edge cases where I thought "not bad enough to raise an exception, but maybe I should log it," so I need to create those edge cases and ensure that the expected message gets logged.

New directory structure this uses:

```
local_root
├── scratch
│   └── dag
│       └── unit2
└── .staging
    └── dag
        ├── unit1
        └── unit2
```

Note that the name `.staging` is customizable; the executor can set it to `""`, in which case you have:

```
local_root
├── dag
│   ├── unit1
│   └── unit2
└── scratch
    └── dag
        └── unit2
```

In this way, the "staging" files can be final output location for local runs. (This is what I'd like to do with `quickrun`.) That way, if someone wants to check on the initial units of a DAG, the location is the same as checking those units after the DAG has completed (thinking of the people who just want to work with the filesystem).

File: gufe/storage/stagingdirectory.py [hunk 6]
```diff
@@ -0,0 +1,359 @@
...
+            scratch=self.scratch,
+            external=self.external,
+            prefix=prefix,
+            staging=self.staging,
+            delete_staging=delete_staging,
+            read_only=True,
+        )
+
+    @contextmanager
+    def other_shared(self, prefix: str,
```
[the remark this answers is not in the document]
dwhswenson: This part might not be necessary if `inputs` can give the `shared` from previous units (would require a context manager on the `ProtocolUnitResult` to handle that; not to be part of this PR). One of the problems I had was that the actual value for the `prefix` depends on executor, so if we can't agree that executors use a common labeling scheme, this becomes a problem.

gufe/storage/stagingdirectory.py
```diff
@@ -0,0 +1,359 @@
...
+        *,
+        staging: PathLike = Path(".staging"),
+        delete_staging: bool = True,
+        read_only: bool = False,
+    ):
+        super().__init__(scratch, external, prefix, staging=staging,
+                         delete_staging=delete_staging)
+        self.read_only = read_only
+
+    def get_other_shared(self, prefix: str,
```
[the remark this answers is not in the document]
dwhswenson: Recommended to use `other_shared`, but not required. If an executor wants to do manual cleanup, or if for some reason a context manager is a pain, I didn't want to disallow it.

gufe/storage/stagingdirectory.py
```diff
@@ -0,0 +1,359 @@
...
+    def __fspath__(self):
+        return str(self.staging_dir)
+
+    def __repr__(self):
+        return (
+            f"StagingDirectory({self.scratch}, {self.external}, "
+            f"{self.prefix})"
+        )
+
+    def __del__(self):  # -no-cov-
```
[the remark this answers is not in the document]
dwhswenson: Recommended, but not required. The `__del__` ensures cleanup in case an executor doesn't use the `StorageManager` and associated context managers.

gufe/tests/storage/test_storagemanager.py
```diff
@@ -0,0 +1,311 @@
...
+            "foo": root / "scratch/dag/unit1/foo.txt",
+            "foo2": root / "scratch/dag/unit2/foo2.txt",
+            "bar": root / staging / "dag/unit1/bar.txt",
+            "baz": root / staging / "dag/unit1/baz.txt",
+        }
+
+    def test_lifecycle(self, storage_manager, dag_units, tmp_path):
+        results = []
+        dag_label = "dag"
+        with storage_manager.running_dag(dag_label) as dag_ctx:
```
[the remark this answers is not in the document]
dwhswenson: Yes, at least this is the skeleton of it.

gufe/storage/stagingdirectory.py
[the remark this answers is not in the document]
richardjgowers: Can we stick to a single way of doing this in this PR?  Just make it work for execute_DAG and we'll go from there

gufe/storage/stagingdirectory.py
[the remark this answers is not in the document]
richardjgowers: again, let's pick a single way to use these and make that work rather than maintain two approaches

gufe/storage/stagingdirectory.py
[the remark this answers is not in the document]
dwhswenson: Moved `get_other_shared` to private in [0f2597](https://github.com/OpenFreeEnergy/gufe/pull/186/commits/90f2597d1af1cc4e5b2b8c70a57618d1feb889fb)

docs/guide/storage.rst
```diff
@@ -1,5 +1,106 @@
...
+``permanent`` can be external (remote) resources, using the
+:class:`.ExternalResource` API.
+
+As a practical matter, the GUFE storage system can be handled with a
+:class:`.StorageManager`. This automates some aspects of the transfer
+between stages of the GUFE storage system, and simplifies the API for
+protocol authors.  In detail, this provides protocol authors with
+``PathLike`` objects for ``scratch``, ``shared``, and ``permanent``. All
+three of these objects actually point to special subdirectories of the
+scratch space for a specific unit, but are managed by context manangers at
```
dotsdl, line 43:
```suggestion
local execution space for a specific unit, but are managed by context managers at
```

docs/guide/storage.rst
```diff
@@ -1,5 +1,106 @@
...
+  key-value store, as well as the actual local path to the file. On
+  creation, it registers itself with its :class:`.StagingDirectory`, which
+  handles managing it over its lifecycle.
+
+In practice, the executor uses the :class:`.StorageManager` to create a
+:class:`.DAGContextManager` at the level of a DAG, and then uses the
+:class:`.DAGContextManager` to create a context to run a unit. That context
+creates a :class:`.SharedStaging` and a :class:`.PermanentStaging`
+associated with the specific unit. Those staging directories, with the
+scratch directory, are provided to the :class:`.ProtocolDAGUnit`, so that
```
dotsdl, line 105:
```suggestion
scratch directory, are provided to the :class:`.ProtocolUnit`, so that
```

richardjgowers (commented):
this is partial, I'm still going to go through the tests and try breaking this locally myself

gufe/storage/stagingregistry.py
```diff
@@ -0,0 +1,397 @@
...
+        """
+        return _safe_to_delete_staging(
+            external=self.external,
+            path=self.staging_dir,
+            prefix=self.prefix,
+        )
+
+    def transfer_single_file_to_external(self, held_file: StagingPath):
+        """Transfer a given file from staging into external storage
+        """
```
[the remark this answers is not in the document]
dwhswenson: > if I've followed this, it returns the held_file back if it was transferred, but otherwise None? 

Yes. The real point here is that `transfer_staging_to_external`, which tries to transfer everything, returns the list of files that have been transferred. That was needed because it turned out that the successfully transferred files are needed by the `StorageManager` if `keep_shared is False`. I might be able to move that responsibility into a `delete_transferred` method on `StagingRegistry`, though.

> Can `store_path` ever (knowingly) fail?

I'm not sure what you mean here by "knowingly." If `store_path` does not successfully store the path, it should raise an exception. The only way it can communicate failure is via exception, and if it fails silently, that's a bug.

docs/guide/storage.rst
```diff
@@ -1,5 +1,106 @@
...
+
+   The ``shared`` space must be a resource where an uploaded object is
+   instantaneously available, otherwise later protocol units may fail if the
+   shared result is unavailable. This means that files or approaches based
+   on ``scp`` or ``sftp`` are fine, but things like cloud storage, where the
+   existence of a new document may take time to propagate through the
+   network, are not recommended for ``shared``.
+
+
+Details: Manangement of the Storage Lifetime
```
richardjgowers, line 70: typo: management

PR #191, opened 2023-04-30
updated off models pin to 0.0.5

richardjgowers:
FloatQuantities in Settings are now cast to float even if integers were used as input

updated related hashes


IAlibay (approved):

IAlibay:
hopefully this stops CI from failing

PR #193, opened 2023-05-05
Issue 168 basic telemetry

richardjgowers:
basic start/end times

probably not all we want to add, but definitely something we do want to add.


IAlibay (approved):

dotsdl:
Question: what happens if a user loads a `ProtocolUnitResult` that *doesn't* have `start_time` and `end_time` in its dict? From the code, it looks like the class will raise an exception.

Any objection to these being default `None`?

PR #200, opened 2023-05-31
Docs: Overview of gufe

dwhswenson:
The idea here is to give the very-big-picture view of what gufe involves. As docs for each of the headers in this get filled out,  these headers should link to that.

Audience here is incoming devs, to help orient them to the project.

### DIFF
--- a/docs/guide/overview.rst
+++ b/docs/guide/overview.rst
@@ -1,0 +1,83 @@
+GUFE Overview
+=============
+
+GUFE exists to define interoperable APIs for free energy calculations, which
+can be used by an ecosystem of Python packages to develop tools that focus
+on performing specific aspects of the free energy pipeline, while
+benefitting from other tools for other aspects.
+
+In order to do this, GUFE distinguishes several aspects of the process of
+calculating free energies, and define APIs for each of them. GUFE provides
+the underlying infrastructure; the real science is done in packages that
+implement parts of the GUFE APIs.
+
+Core data models
+----------------
+
+In order to ensure interoperability, GUFE defines objects that represent the
+core chemistry and alchemistry of a free energy pipeline, including
+molecules, chemical systems, and alchemical transformations. This provides a
+shared language that different tools use.
+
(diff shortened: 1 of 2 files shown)


docs/guide/overview.rst
```diff
@@ -0,0 +1,83 @@
...
+executor to determine how to/whether to use those.
+
+Strategies
+----------
+
+Strategies have yet to be implemented, but the GUFE design leaves a place
+for an object that, at the scale of an alchemical network, can dynamically
+decide where to focus more simulation effort based on the results that have
+been received so far. This will be useful for adaptive approaches to
+sampling a network.
```
richardjgowers, line 73: I'd maybe leave this out if it doesn't exist?  Or maybe have a separate page of planned future scope?

docs/guide/overview.rst
```diff
@@ -0,0 +1,83 @@
...
+implement parts of the GUFE APIs.
+
+Core data models
+----------------
+
+In order to ensure interoperability, GUFE defines objects that represent the
+core chemistry and alchemistry of a free energy pipeline, including
+molecules, chemical systems, and alchemical transformations. This provides a
+shared language that different tools use.
+
```
richardjgowers, line 21: Is it worth having a brief section on `Components` here?

docs/guide/overview.rst
dwhswenson: I struggled with the same. We need to document this concept *somewhere*, because it informs other architectural decisions. I thought that putting a very brief paragraph here was the least-bad option, but I'd be open to other options. But I like this page being a 1-page overview of the whole architecture, as a place for people to start.

docs/guide/overview.rst
dwhswenson: I think the next page in the guide is going to be a discussion of the `Component`s approach (we already have a placeholder for it). My goal for this page is to be an extended table of contents for the rest of the "guide" section; details will come later.

PR #203, opened 2023-06-08
Issue 192 execute dag restarts

richardjgowers:


gufe/protocols/protocoldag.py
```diff
@@ -359,6 +359,7 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
                 keep_shared: bool = False,
                 keep_scratch: bool = False,
                 raise_error: bool = True,
+                n_retries: int = 0,
```
IAlibay, line 362: Are we changing the default in quickrun?

richardjgowers: Yes, different repo but 2 retries would make sense?

IAlibay: Yeah my question was more "here or there".

dwhswenson (approved):
LGTM, although I'm not sure what's going on with the existing CI failure.

A lot of the details here will change when we integrate #186 into `execute_DAG`, but I don't see anything that will cause fundamental conflicts.

PR #204, opened 2023-06-08
LigandNetwork.to_*_alchemical_network

dwhswenson:
@richardjgowers asked for a method on `LigandNetwork` to attach the protocol. This effectively moves some of the easy-planner stuff into gufe, but this aspect of it is likely to be broadly useful.

So far, a first glance of what I see the API looking like. Rough code, completely untested. Please take a look and see if this is the direction I should continue!

### DIFF
--- a/gufe/ligandnetwork.py
+++ b/gufe/ligandnetwork.py
@@ -224,7 +224,8 @@ class LigandNetwork(GufeTokenizable):
                 sysA = sys_from_dict(edge.componentA)
                 sysB = sys_from_dict(edge.componentB)
                 if autoname:
-                    name = f"{autoname_prefix}_{sysA.name}_{sysB.name}"
+                    prefix = f"{autoname_prefix}_" if autoname_prefix else ""
+                    name = f"{prefix}{sysA.name}_{sysB.name}"
                 else:
                     name = ""
 
--- a/gufe/ligandnetwork.py
+++ b/gufe/ligandnetwork.py
@@ -213,12 +213,17 @@ class LigandNetwork(GufeTokenizable):
 
                 # define a helper func to avoid repeated code
                 def sys_from_dict(component):
+                    """
+                    Input component alchemically changing. Other info taken
+                    from the outer scope.
+                    """
                     syscomps = {alchemical_label: component}
+                    other_labels = set(labels) - {alchemical_label}
                     syscomps.update({label: components[label]
-                                     for label in labels})
+                                     for label in other_labels})
 
                     if autoname:
-                        name = f"{sys[alchemical_label].name}_{leg_name}"
+                        name = f"{component.name}_{leg_name}"
                     else:
                         name = ""
 
@@ -275,7 +280,8 @@ class LigandNetwork(GufeTokenizable):
         }
         leg_labels = {
             "solvent": ["ligand", "solvent"],
-            "complex": ["ligand"] + list(other_components),
+            "complex": (["ligand", "solvent", "protein"]
+                        + list(other_components)),
         }
         return self._to_rfe_alchemical_network(
             components=components,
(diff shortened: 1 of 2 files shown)


gufe/ligandnetwork.py
```diff
@@ -178,3 +177,116 @@ def enlarge_graph(self, *, edges=None, nodes=None) -> LigandNetwork:
...
+
+    def _to_rfe_alchemical_network(
+        self,
+        components: dict[str, gufe.Component],
+        leg_labels: dict[str, list[str]],
+        protocol: gufe.Protocol,
+        *,
+        alchemical_label="ligand",
+        autoname=True,
+        autoname_prefix=""
```
richardjgowers, line 189: since names are never harmful, how about compact these two kwargs into just `name_prefix=""`

richardjgowers (commented):
Yes this looks useful.

gufe/ligandnetwork.py
dwhswenson: Clarify; which of these to do you mean:

1. Something equivalent to `autoname = bool(name_prefix)`, where auto naming only happens if a `name_prefix` is given.
2. Always naming everything.

I think 1 is a nonstarter, because personally, I would prefer *not* to have a meaningless prefix attached to names. I know which campaign they are from because of the network they are in, and so I'd generally prefer an empty prefix. (current default prefix is to match current network planners).

The consequence of 2 that I don't particularly like is that, if I know I won't have unique names, it seems more natural to me to have the name field by empty, instead of `f"{prefix}__complex__complex"` for everything. But anything downstream that needs unique names should actually check for unique names and not assume that any non-empty name is unique. So forcing names wouldn't actually cause problems; it's just a question of user expectations.

dwhswenson: Close to done here. I need a review on the current state and clarification on this point re: naming UX; once I know what we want the UX to be around naming, I'll write tests to cover that. The rest should be ready.

richardjgowers: Yeah I'm thinking of 1).  I don't understand how `autoname_prefix=""` is different from `autoname=False`.  This can be the default, which I think means we don't by default have a meaningless prefix.

richardjgowers: Or another way of thinking of it, there's a 2x2 matrix of options for `autoname=False/True` and `autoname_prefix=""/"something"`, I don't understand what `autoname=True` and `autoname_prefix=""` and `autoname=False` and `autoname_prefix="something"` mean

dwhswenson: Three use cases:

1. I don't want names at all. `autoname=False`; `autoname_prefix` ignored. Neither system nor transition get named. Maybe you know that the ligand set you have doesn't have unique names (e.g., some names are missing) and so you can't guarantee unique names from the autonaming. IMO, no names are better than automatically-generated non-unique names.
2. I don't want some prefix (system names are enough for me). `autoname=True`, `autoname_prefix=""`. Systems get named, transformation name is `f"{sysA.name}_{sysB.name}"`. This is probably the use case I'd personally want.
3. I want to match exactly what we do currently, which is to add a prefix that could provide extra information, but by default doesn't add anything informative. `autoname=True`, `autoname_prefix="easy_rbfe"`. System names are the same as in 2, but transformation name is `f"{autoname_prefix}_{sysA.name}_{sysB.name}"`.

1 and 2 are distinct use cases, so you can't just say `autoname = bool(autoname_prefix)` and skip having `autoname` as a parameter. If you really want to only have one parameter, we can get rid of `autoname_prefix` and type `autoname` as `autoname: bool | str`, in which case `autoname=False` is 1, `autoname=True` is 2, and `autoname="easy_rbfe"` is 3.

richardjgowers (approved):

PR #213, opened 2023-08-02
Pin autodoc_pydantic

IAlibay:
Just saw our cron CI jobs failed last night.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,8 +7,8 @@ dependencies:
   - networkx
   - numpy
   - openff-toolkit >=0.13.0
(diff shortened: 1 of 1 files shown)


IAlibay:
conda_cron came back clean so I assume our gufe conda-forge recipe is adequately pinned

IAlibay:
Turns out autodoc_pydantic was the culprit all along.

dwhswenson (approved):
Do we have any sense how hard it is going to be to support both pydantic 1.x and pydantic 2? Especially whether the switch will break anything related to serialization?

Pinning definitely works for now, of course.

PR #215, opened 2023-08-08
Always raise, don't swallow exception, if exception is `KeyboardInterrupt` in `ProtocolUnit.execute`

dotsdl:
This PR is an attempt to address openforcefield/alchemiscale#132.

@mikemhenry raised a good point that for the case of a `KeyboardInterrupt`, even if `raise_error = False` for `ProtocolUnit.execute`, we probably always want to raise.

Is there any downside to this?

### DIFF
--- a/gufe/protocols/protocolunit.py
+++ b/gufe/protocols/protocolunit.py
@@ -322,6 +322,9 @@ class ProtocolUnit(GufeTokenizable):
                 start_time=start, end_time=datetime.datetime.now(),
             )
 
+        except KeyboardInterrupt:
+            # if we "fail" due to a KeyboardInterrupt, we always want to raise
+            raise
         except Exception as e:
             if raise_error:
                 raise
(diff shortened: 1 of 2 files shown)


dwhswenson:
My only concern would be if `INT` is being used elsewhere, e.g., by a queueing system, in which case we'd want to catch it and terminate cleanly. (Which also might be an option here, anyway.) 

The queueing systems I know send `TERM` or `USR2`, followed by a `KILL`; I'm not aware of one using `INT`, but I don't claim to know them all!

If we ever handle mid-unit restarts, we'll need to handle `KeyboardInterrupt` with a little more logic, but that's easily added. Otherwise, seems reasonable -- either needs a test or exclude from coverage.

dotsdl:
@dwhswenson thanks for this! I've added an explicit test for the new behavior.

As for the concern you raised about queueing systems, I think you're right that SIGINT generally isn't used for terminating jobs, so I think we're still good here.

PR #216, opened 2023-08-10
Prevent `_GufeTozenizableMeta.__call__` from overwriting signature of `__init__`

Yoshanuikabundi:
When the `inspect` module of the standard library is used to get the signature of a `GufeTokenizable` class, it sees that the metaclass has an implementation of `__call__` and takes the signature from that instead of from the `__new__` or `__init__` method of the child class:

```python
>>> from gufe import SmallMoleculeComponent
>>> import inspect
>>> inspect.signature(SmallMoleculeComponent)
<Signature (*args, **kwargs)>
>>> inspect.signature(SmallMoleculeComponent.__init__)
<Signature (self, rdkit: rdkit.Chem.rdchem.Mol, name: str = '')>
```

This results in unhelpful API documentation:

SmallMoleculeComponent_signature_before

But also may affect the signature displayed by IDEs, the Jupyter `help` magic, and so on.

This PR restores the signature from `__init__` or `__new__` when a child class is created:

SmallMoleculeComponent_signature_after

(I've taken screenshots from the GUFE docs, but this problem affects the OpenFE docs as well, and this solution fixes them too)

See https://stackoverflow.com/questions/49740290/call-from-metaclass-shadows-signature-of-init. I originally wrote a solution defining a `__signature__` property on the metaclass, but this solution seems more... canonical.

### DIFF
--- a/gufe/tokenization.py
+++ b/gufe/tokenization.py
@@ -150,7 +150,6 @@ class GufeTokenizable(abc.ABC, metaclass=_ABCGufeClassMeta):
     def __hash__(self):
         return hash(self.key)
 
-
     def _gufe_tokenize(self):
         """Return a list of normalized inputs for `gufe.base.tokenize`.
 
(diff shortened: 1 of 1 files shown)


richardjgowers (approved):
thanks @Yoshanuikabundi this had been bugging me

Issue #139, opened 2023-02-23
rename ProteinComponent to PDBComponent?


richardjgowers:

richardjgowers:
@IAlibay the thinking is that currently we're making a lot of PDB specific assumptions about the content of the underlying model (resnames, resnums + icodes, chainids) so that it's not a general protein model but specifically a PDB-format view of a protein model.  

What other ways to load things might we add (that would include the above special fields)?

IAlibay:
don't we mostly make these assumptions on load? I feel like elsewhere we just assume it's something that can be digested by rdkit and eventually passed along to PDBFile/Modeller?

My main view here re: naming here is 1) the intent of this component is that it's a polymer that can get parametrised using off the shelf additive force fields, 2) technically there's nothing that should stop you from passing in a customised / self built rdkit molecule to this component, 3) pretty sure something like a gro file or more programmatically an MDA Universe would include all the necessary information you'd need - in the long run it'd be great to just go "here are attributes you need, make this an rdkit molecule and it'll work"

richardjgowers:
So one key difference between SmallMoleculeComponent and ProteinComponent is that the underlying rdkit molecule in ProteinComponent is expected to have the MonomerInfo() attribute (which holds resname etc) populated.  This for example then populates the Hierarchy system in an OpenFF Molecule/Topology

IAlibay:
I think I'd vote for PolymerComponent here, or something like that, and then make it clear that the rdkit molecule needs that information. Maybe I'm just overthinking the cool possibility of creating arbitrary polymers without ever seeing an input file though.

IAlibay:
I see what you mean there, but what if you went the other way round, i.e. you create a component from an OpenFF polymer molecule? (created through some arbitrary means that has this information)

IAlibay:
My main point here is - are we naming for input or for what the object represents. Both are ok, but then we should probably consider having the same naming scheme across the board, otherwise it's confusing for users.

ijpulidos:
Arguably, there must be some "standards" for computer-friendly representation of proteins. I'd argue PDB is one of those, I think whatever we chose here must to adhere to some standard way of formatting the protein information and we just have to enforce it in our code base. Is there a way around that regardless of what we choose?

I agree our current component relies heavily on the PDB way of doing things, for better or for worse.  Maybe when we have the need to support other formats and standards we could create specific classes/models for them, not sure how API breaking this could be in the end, though.

PR #220, opened 2023-08-22
Improve docstrings as needed for OpenFE

Yoshanuikabundi:
This PR moves more docstrings out of `attribute` sections of docstrings and onto the corresponding object itself. This helps comprehensive API documentation in OpenFE. Required by https://github.com/OpenFreeEnergy/openfe/pull/527

### DIFF
--- a/docs/api/tokenization.rst
+++ b/docs/api/tokenization.rst
@@ -1,0 +1,4 @@
+GUFE Tokenization API
+---------------------
+
+.. autoclass:: gufe.tokenization.GufeTokenizable
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -56,7 +56,8 @@ positive_ions = ["NA", "MG", "ZN"]
 
 
 class ProteinComponent(ExplicitMoleculeComponent):
-    """Wrapper around a Protein representation.
+    """
+    :class:`Component` representing a protein.
 
     In comparison to a SmallMoleculeComponent, this representation additionally contains information
     relating to the residue and chain information.  This is achievable by having the MonomerInfo attributes
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -57,12 +57,13 @@ positive_ions = ["NA", "MG", "ZN"]
 
 class ProteinComponent(ExplicitMoleculeComponent):
     """
-    :class:`Component` representing a protein.
+    ``Component`` representing the contents of a PDB file, such as a protein.
 
-    In comparison to a SmallMoleculeComponent, this representation additionally contains information
-    relating to the residue and chain information.  This is achievable by having the MonomerInfo attributes
-    present on each atom of the input RDKit molecule, which is done when reading from either PDB or `.mae`
-    file inputs.
+    In comparison to a SmallMoleculeComponent, this representation additionally
+    contains information relating to the residue and chain information.  This
+    is achievable by having the ``MonomerInfo`` attributes present on each atom
+    of the input RDKit molecule, which is done when reading from either PDB or
+    ``.mae`` file inputs.
 
     Note
     ----
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -49,33 +49,30 @@ class DAGMixin:
         return reversed(list(nx.topological_sort(graph)))
 
     @property
-    def name(self):
+    def name(self) -> Optional[str]:
+        """Optional identifier."""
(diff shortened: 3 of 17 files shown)


docs/api/tokenization.rst
```diff
@@ -0,0 +1,4 @@
+GUFE Tokenization API
```
dwhswenson, line 1: Maybe something more like just `GufeTokenizable` here? This is the base class from which everything else inherits. The relevant API points aren't really about "tokenization" (indeed, all the tokenization is done in bare functions that aren't listed here)

gufe/tokenization.py
```diff
@@ -199,7 +199,7 @@ def defaults(cls):
         """Dict of default key-value pairs for this `GufeTokenizable` object.
 
         These defaults are stripped from the dict form of this object produced
-        with `to_dict(include_defaults=False) where default values are present.
+        with `to_dict(include_defaults=False)` where default values are present.
```
dwhswenson, line 202:
```suggestion
        with ``to_dict(include_defaults=False)`` where default values are present.
```

whoever wrote these docs forgot that docs should be in RST.

dwhswenson (changes requested):
Mostly looks good to me. One section title that I think will be very misleading, and so needs to be renamed, and a tiny fix that is because @Yoshanuikabundi fixed *half* of someone else's mistake, otherwise it would have gone unnoticed.

gufe/tokenization.py
Yoshanuikabundi: Ok so I wasn't sure about this one because the NumPy docstring spec has this insane thing where you're supposed to put variable names in single backticks, so I never know whether to correct it. This particular case is not a variable so it should definitely be in double backticks... But this kind of thing is alllll over the docstrings. I'll quickly do a grep to see if I can find them all!

docs/api/tokenization.rst
Yoshanuikabundi: Tokenization -> Serialization

Yoshanuikabundi: Make sure there are no filename clashes.

Yoshanuikabundi:
I think this is ready to go!

richardjgowers (approved):

gufe/components/proteincomponent.py
```diff
@@ -56,7 +56,8 @@
 
 
 class ProteinComponent(ExplicitMoleculeComponent):
-    """Wrapper around a Protein representation.
+    """
+    :class:`Component` representing a protein.
```
IAlibay, line 60: We really should rename this component - in reality this is just "a wrapper around something you placed in a PDB".

Not sure if we at least want to fix this in the docs?

dwhswenson (commented):
Small change recommended: a couple of times we talk about the DAG dependency order; we should be very explicit in which order that is.

I think it is in the order that follows the arrow of time, is that correct, @dotsdl @richardjgowers? In the sense that a "parent" node will come before a "child" node, i.e., that this order is a possible execution order for the DAG.

I know @Yoshanuikabundi is mainly moving existing text around, but as long as we're taking a closer look, let's clarify things.

gufe/components/proteincomponent.py
Yoshanuikabundi: Fixed!

Yoshanuikabundi:
@dwhswenson I completely agree that these docstrings are unclear, but can we leave it for another PR? It'll need input from someone who knows what is going on and this PR is currently holding up https://github.com/OpenFreeEnergy/openfe/pull/527. I've raised an #223 as a reminder.

dwhswenson (approved):

dwhswenson:
Yep... with #223 to remind, I think this is good. Sorry if this slows you down at all. I'm trying to take advantage of a new set of eyes on the docs as a chance to improve aspects that are unclear when we look at them with our own fresh eyes.

gufe/components/proteincomponent.py
richardjgowers: c.f. https://github.com/OpenFreeEnergy/gufe/issues/139

gufe/protocols/protocoldag.py
```diff
@@ -181,10 +152,18 @@ def _from_dict(cls, dct: dict):
 
     @property
     def result_graph(self) -> nx.DiGraph:
+        """
+        DAG of `ProtocolUnitResult` nodes with edges denoting dependencies.
```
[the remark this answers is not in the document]
richardjgowers: edges point to previous

PR #227, opened 2023-09-06
Versioning and migration for `GufeTokenizable`s

dwhswenson:
- [x] Migration for adding keys
- [x] Migration for removing keys
- [x] Migration for renaming keys
- [x] Migration for reorganizing keys within nested structures
- [x] Full tests for nested structure migrations functions
- [x] Try to preserve existing gufe keys (if possible, ensure that existing keys are not broken by this PR)
- [x] Add `GufeTokenizableTestMixin` to tests of migration processes

We decided that `GufeTokenizable`s need to be versioned, and that we need to provide a migration path to load old serialized representations of the `GufeTokenizable`. This PR provides the core functionality to do that.

### Changes to `GufeTokenizable`

* classmethod-property `_version`: Should return a sortable marking the serialization version.
* classmethod `serialization_migration`: Takes a dict (possibly from another version of serialization, but otherwise suitable to be ingested by `GufeTokenizable._from_dict`) and a version, and converts the dict to ensure that it can be handled by the current version of the code.

### Changes to the serialized representation of `GufeTokenizable`s

Now serialized versions of `GufeTokenizable`s (described as a Python dict/JSON object) will carry the additional key `":version:"`, which describes the version. Serialized representations without that key will be treated as if its value was `1`. Like `__module__` and `__qualname__`, this is added/popped without intervention from contributors.

### Labelling versions

I'm proposing we use versions that are simple increasing integers. In principle, users could use anything. However, I think that semantic versioning has no meaning here (the only purpose would be to distinguish breaks that can no longer be handled by the code, which is probably as easily handled with a simple `version < oldest_supported_version`). Tying to the version of the package also makes it more tedious to handle updates: first, this means that it requires more careful attention to the version each new thing was added in. Second, this is a nightmare for dev installs, because they lie about version number.

### On reorganizing keys within nested structures

Although we identified this as target, I don't think this will be easily done with general `GufeTokenizable`s. The issue is that, in deserialization, we can use gufe keys to shortcut through the tree of nested object, gaining some performance. However, if we want to rearrange the tree of the nested structure, we need to reconstruct the entire structure, and the make the modifications. This is a significantly different process than the one  we use now, and may be slower.

That said, moving an attribute within a digraph is probably an indication that the original design was significantly flawed, which is not something I think we need to worry about with `GufeTokenizable`s. For example, I have a hard time imagining that we would want to transfer `Transformation.protocol.settings` to `Transformation.chemical_systems.protocol_settings` or transfer a hypothetical `AlchemicalNetworkCreator.mapper.timeout` to `AlchemicalNetworkCreator.timeout`.

Where this will matter is in `Settings`, where we expect rapid flux in the near future. However, `Settings` are not `GufeTokenizable`s. Therefore, their "key-encoded" representation is still a nested dict, and the parent `GufeTokenizable` (the `Protocol`) can manage migrations. As of opening this PR, tools to facilitate that haven't been added; that will be a future step, if this PR isn't nixed.

With this PR, changes in settings will require changes in all parent protocols. We could additionally create similar versioning of `Settings`, which would duck-type the same as the approach here, but build on tools within the JSON codecs instead of the core `GufeTokenizable`. This can be done, but would be part of a separate PR. That said, a change in organization (moving an attribute from cousin to cousin) will always require that the correction be handled at a common ancestor (grandparent). Purely local changes are not possible for a change that is fundamentally nonlocal.

### Sequential migrations

This isn't illustrated here, but the pattern to use is actually something like this:

```python
if version <= 1:
    dct = ... # do stuff for v1 -> v2
if version <= 2:
    dct = ... # do stuff for v2 -> v3
if version <= 3:
    # etc
```

This does a series of sequential migrations, and continues to support version 1, 2, and 3 with minimal code. If we find that the chain of versions gets too long, it might be worth creating some skips in there, but I kind of doubt it will ever be worth going to the "correct" solution of creating the directed network and finding the minimal path.


gufe/tokenization.py
```diff
@@ -539,8 +571,13 @@ def from_dict(dct) -> GufeTokenizable:
 def _from_dict(dct: Dict) -> GufeTokenizable:
     module = dct.pop('__module__')
     qualname = dct.pop('__qualname__')
+    if ':version:' in dct:
+        version = dct.pop(':version:')
```
richardjgowers, line 575: isn't this just `dct.pop(':version:', 1)`?

richardjgowers (commented):
@dwhswenson this looks good to me.  I agree wrt the rearrangement case, it sounds like this can be done by the tokenizable that holds Settings?

gufe/tokenization.py
dwhswenson: So, I literally looked that up because I thought there should be such a thing, but didn't see it in the docs. I think I accidentally hit the docs for `list.pop`.

Yes, this is what it should have been. Fixed.

dwhswenson:
>  I agree wrt the rearrangement case, it sounds like this can be done by the tokenizable that holds Settings?

Yes; there's a test that serves as example in [0161938](https://github.com/OpenFreeEnergy/gufe/pull/227/commits/0161938d04552fde8d2dcb73ba3ca0525d94a1c6). This is also the downside of this approach: Any change we make to common settings will need to need to be changed in *all* protocols. And since it is the version of that parent object that matters, there's not much we can do within this approach to make it easier (a settings change that happens in between v2 and v3 OpenFE RBFE protocol could affect v1 to v2 in a Perses non-eq protocol; so the version number isn't the same and can't be used in for migration in a superclass). 

Basically, think about the tree structure we have of objects nested in other objects: Protocol, Settings, Subsettings. For a reorganizational change, the version has to be changed somewhere on the tree that knows about both the "move from" and "move to" components -- Protocol knows everything, but earlier parts of the settings tree also might know both.

Now color each of the nodes in that tree by who owns them. Because we don't have a strict separation of ownership at some level of the tree (some settings/subsettings are gufe, some are protocol dev; Protocol is owned by protocol dev) what happens is that any reorganization change we make could affect versioning of something that we don't own.

In general, we've discussed the need to redesign settings in the future. I think we're going to have to settle for an imperfect solution here, include these issues in our settings redesign.

richardjgowers:
@dwhswenson give me a ping when this is ready for review

dwhswenson:
@richardjgowers : This should be ready for review.

Additional notes on changes here:

* The `':version:'` key is considered as a "default value" for serialization purposes. Since it is constant within a version of the code, this makes sense to me. This also made it very easy to preserve gufe keys for most objects.
* Gufe keys on `LigandNetwork`s are not preserved. The reason for this is that the ligand network `_to_dict` explicitly includes the full serialization of the ligands. This includes the version number for the SMC serialization. We *could* preserve the gufe key now by not including the version (doing that was -- you guessed it -- about 4 lines of code) but the downside of that approach was that a change to the serialization format of `SmallMoleculeComponent` would then need us to also change the version of the `LigandNetwork` serialization. With this approach, it should be easier to just use the SMC serialization versions to deserialize.
* I walked back the classmethod-property -- apparently that's deprecated starting in Python 3.11. Doing it as just a classmethod (needs to be called from another classmethod, so can't be an instance method.)

gufe/tokenization.py
```diff
@@ -126,6 +128,59 @@ def process(self, msg, kwargs):
         return msg, kwargs
 
 
+def new_key_added(dct, new_key, default):
```
richardjgowers, line 131: all of these functions need a docstring to both document what they do and when they're used

gufe/tokenization.py
```diff
@@ -135,6 +190,10 @@ class GufeTokenizable(abc.ABC, metaclass=_ABCGufeClassMeta):
     This extra work in serializing is important for hashes that are stable
     *across different Python sessions*.
     """
+    @classmethod
+    def _version(cls):
```
richardjgowers, line 194: typehint with `->int`

gufe/tokenization.py
```diff
@@ -157,6 +216,10 @@ def _gufe_tokenize(self):
         return tokenize(self)
         # return normalize(self.to_keyed_dict(include_defaults=False))
 
+    @classmethod
+    def serialization_migration(cls, old_dict, version):
+        return old_dict
```
richardjgowers, line 221: can this get documented?  From this I wouldn't know how to implement this in child classes...

dwhswenson:
@richardjgowers : This is ready for another look. Changes you requested were all docstrings/typing, so this should be good.

richardjgowers (approved):
Looks good.  One last question I have is if we want to stick with `_version`/`':version:'` or if we want something more descriptive like `_schema_version` (`_serialisation_version`, `_model_version` etc) to distinguish it from potential confusion with the package version (which we're really trying to avoid confusion with).  Last chance to make this change.

dwhswenson:
Yeah, maybe `schema_version`? (I would normally prefer `serialization_version`, but since spelling disagreements might lead to confusion....) 😆

richardjgowers:
Yeah schema version works for me

richardjgowers:
@dwhswenson looks like we have consensus on `_schema_version` moving forwards

richardjgowers (approved):

PR #228, opened 2023-09-11
Add NPY dtype codec

IAlibay:
TODO:
- [x] Fix issue with np.bool_
- [ ] Fix issue where reconstructed object comes back as float rather than a dtype.

### DIFF
--- a/gufe/custom_codecs.py
+++ b/gufe/custom_codecs.py
@@ -74,7 +74,7 @@ DATETIME_CODEC = JSONCodec(
 
 
 NPY_DTYPE_CODEC = JSONCodec(
-    cls=None,
+    cls=np.generic,
     to_dict=lambda obj: {
         'dtype': str(obj.dtype),
         'bytes': obj.tobytes(),
--- a/gufe/custom_codecs.py
+++ b/gufe/custom_codecs.py
@@ -33,10 +33,12 @@ def inherited_is_my_dict(dct, cls):
     stored = gufe.tokenization.get_class(module, classname)
     return cls in stored.mro()
 
+
 def is_npy_dtype_dict(dct):
     expected = ["dtype", "bytes"]
     is_custom = all(exp in dct for exp in expected)
     return is_custom and ("shape" not in dct)
+
 
 def is_openff_unit_dict(dct):
     expected = ["pint_unit_registry", "unit_name", ":is_custom:"]
@@ -80,6 +82,7 @@ NPY_DTYPE_CODEC = JSONCodec(
     from_dict=lambda dct: np.frombuffer(
         dct['bytes'], dtype=np.dtype(dct['dtype'])
     )[0],
+    is_my_obj=lambda obj: isinstance(obj, np.generic),
     is_my_dict=is_npy_dtype_dict,
 )
 
--- a/gufe/custom_codecs.py
+++ b/gufe/custom_codecs.py
@@ -73,6 +73,10 @@ DATETIME_CODEC = JSONCodec(
 )
 
 
+# Note that this has inconsistent behaviour for some generic types
+# which end up being handled by the default JSON encoder/decoder.
+# The main example of this is np.float64 which will be turned into
+# a float type on serialization.
 NPY_DTYPE_CODEC = JSONCodec(
     cls=np.generic,
     to_dict=lambda obj: {
--- a/gufe/custom_codecs.py
+++ b/gufe/custom_codecs.py
@@ -33,6 +33,10 @@ def inherited_is_my_dict(dct, cls):
     stored = gufe.tokenization.get_class(module, classname)
     return cls in stored.mro()
 
+def is_npy_dtype_dict(dct):
+    expected = ["dtype", "bytes"]
+    is_custom = all(exp in dct for exp in expected)
+    return is_custom and ("shape" not in dct)
 
 def is_openff_unit_dict(dct):
     expected = ["pint_unit_registry", "unit_name", ":is_custom:"]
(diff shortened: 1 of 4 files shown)


gufe/tests/test_custom_json.py
```diff
@@ -115,9 +116,28 @@ def test_round_trip(self):
...
             reconstructed = json.loads(json_str, cls=decoder)
             npt.assert_array_equal(reconstructed, obj)
+            assert reconstructed.dtype == obj.dtype
             json_str_2 = json.dumps(obj, cls=encoder)
             assert json_str == json_str_2
 
+class TestNumpyGenericCodec(TestNumpyCoding):
+    def setup_method(self):
+        self.codec = NPY_DTYPE_CODEC
+        self.objs = [np.float16(1.0), np.float32(1.0), np.float64(1.0),
```
IAlibay, line 126: How odd, for np.float64;

`{':is_custom:': True, '__class__': 'float64', '__module__': 'numpy', 'dtype': 'float64', 'bytes': b'\x00\x00\x00\x00\x00\x00\xf0?'}`

Just json.dumps to 1.0.

IAlibay: Ok so the problem is effectively based on this: https://github.com/python/cpython/issues/84655#issuecomment-1093870644

The default method is only called _if_ the dict can't be serialised, and unfortunately numpy.float64 is JSON serializable to a standard float.

Do we actually care? If not we can just have a bit of a lossy conversion.

Pinging @dwhswenson specifically.

IAlibay:
@dwhswenson what was the plan here again?

dwhswenson:
> @dwhswenson what was the plan here again?

Recommended best practice: Don't try to serialize numpy scalars, because it isn't possible to round-trip a float64. These should be converted to float/int by protocol authors before being returned.

Just in case: we should have these serialization paths to catch when people don't follow best practices. Downside is that float64 is a weird edge case that doesn't round-trip exactly.

TBH, I'd like to figure out why/how float64 is supported in basic stdlib json. But that's me being curious; nothing that should block.

IAlibay:
Ok I'll clean this up and add a warning?

dwhswenson:
> I'd like to figure out why/how float64 is supported in basic stdlib json.

This is why:

```python
In [1]: import numpy as np

In [2]: issubclass(np.float64, float)
Out[2]: True

In [3]: issubclass(np.float32, float)
Out[3]: False
```

gufe/custom_codecs.py
```diff
@@ -52,18 +58,35 @@ def is_openff_quantity_dict(dct):
...
+
 DATETIME_CODEC = JSONCodec(
     cls=datetime.datetime,
     to_dict=lambda obj: {'isotime': obj.isoformat()},
     from_dict=lambda dct: datetime.datetime.fromisoformat(dct['isotime']),
 )
 
+
+NPY_DTYPE_CODEC = JSONCodec(
+    cls=np.generic,
```
dwhswenson, line 77: can use `cls=None` here: `cls` is only used in default functions; for this codec, you've overridden all4 funds.

dwhswenson (changes requested):
Looks good, but could you add a couple small tests to check that the order of the `NUMPY_CODEC` and `NPY_DTYPE_CODEC` doesn't make a difference? Use `JSONSerializerDeserializer` with different codec orders; you'll probably also need to include the `BYTES_CODEC`. Check a serialization cycle for a `np.generic` and for a `np.ndarray`.

This is because the `is_my_*` methods are checked in codec registration order. It looks to me like the code will be fine, but since the serialization dicts of those two are so similar, it would be good to double check that these are order-independent -- possible footgun for future changes.

gufe/custom_codecs.py
IAlibay: I noticed that the np.bool types are getting squashed to bool if this isn't set, not sure what's going on here.

IAlibay:
@dwhswenson - adding a warning for np.float64 would be really useful here, but I'm not sure where we would put it - my understanding is that `get_default` won't get called in that case?

dwhswenson:
> @dwhswenson - adding a warning for np.float64 would be really useful here, but I'm not sure where we would put it - my understanding is that `get_default` won't get called in that case?

That's correct, and I don't think there's a good fix. Three somewhat distinct thoughts on this:

To fix this, I think we'd have to sacrifice a lot of speed (most encoding is done in C; there's backup code in Python, but since these algos are recursive....) It also would be an annoying fix: it would require copying and modifying the `JSONEncoder.iterencode` method to support a custom `floatstr` that has a hook to check for this. This adds a bit of code that we would need to keep in sync with CPython; there's no easy way to `super()` our way out of that.

I'm against trying to do that. I think the main concern here is that, on a save-reload cycle, we lose the `dtype` on `np.float64`. I think we can document that `AttributeError: 'float' object has no attribute 'dtype'` means that (1) they should have been saving scalar floats as `float`; (2) they can get it back to having a `dtype` by wrapping it in `np.float64`.

---

It seems that the explicit intention in CPython's JSON module is to be lossy in these cases:

> Subclasses of int/float may override `__repr__`, but we still want to encode them as integers/floats in JSON. One example within the standard library is IntEnum.

https://github.com/python/cpython/blob/f2636d2c45aae0a04960dcfbc7d9a2a8a36ba3bc/Lib/json/encoder.py#L311-L313

I suppose one could argue that `np.generic`s are "conceptual" subclasses, even if only `float64` is a literal subclass. Not sure I'm in favor of this, but it would give a kind of consistency (and some safety in case NumPy ever decides to make one of its `np.intXX` objects a subclass of Python's `int`, although I don't think that's possible).

For example, I think we'd already have the same round-tripping problem if anyone saved a `np.str_`:

```python
In [1]: import numpy as np

In [2]: issubclass(np.str_, str)
Out[2]: True
```

---

One final option here is to do a recursive scan for `np.generic`s before we enter the JSON `iterencode`. I'm not sure what the performance impact would be, but that could catch most problems. However, it would still miss cases where `np.generic`s were contained inside other objects that require custom serialization; e.g., `np.float64(300) * unit.kelvin`.

tl;dr: My vote is to leave as is, try to document the issues, and steer people toward the best practice of not storing NumPy scalars.

IAlibay:
Yeah ok, adding a warning here seems kinda over the top. Do we have docs somewhere we can populate with a warning note?

IAlibay:
guide/serialization.rst?

IAlibay:
Seems like the need for this is popping up against in my solvation free energy benchmarks - I _think_ for some reason conformer generation is leading to us storing a complex128 when serialising SMCs. Using this branch things serialise/deserialise fine though.

dwhswenson (approved):
Yeah, with the docs updates to clarify, looks good to me. Merging this.

PR #230, opened 2023-09-11
Fixes ProtocolResult.from_dict

### LINKED ISSUE #229: RelativeHybridTopologyProtocolResult does not round trip through to_dict/from_dict
```python
>>> tpr = openmm_rfe.RelativeHybridTopologyProtocolResult(...)
>>> tpr == tpr.from_dict(tpr.to_dict())
False
```

richardjgowers:
fixes #229

### DIFF
--- a/gufe/tests/test_protocolresult.py
+++ b/gufe/tests/test_protocolresult.py
@@ -1,0 +1,28 @@
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
+from openff.units import unit
+import pytest
+
+import gufe
+from .test_tokenization import GufeTokenizableTestsMixin
+
+
+class DummyProtocolResult(gufe.ProtocolResult):
+    def get_estimate(self):
+        return self.data['estimate']
+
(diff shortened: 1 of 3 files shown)


gufe/tests/test_protocolresult.py
```diff
@@ -0,0 +1,28 @@
...
+        return self.data['estimate']
+
+    def get_uncertainty(self):
+        return self.data['uncertainty']
+
+
+class TestProtocolResult(GufeTokenizableTestsMixin):
+    cls = DummyProtocolResult
+    key = 'DummyProtocolResult-b7b854b39c1e37feabec58b4000680a0'
+    repr = f'<{key}>'
```
dwhswenson, line 21: ooo, so that my today-I-learned for the day... I should have known it would work, but hadn't thought of it!

dwhswenson (approved):
lgtm, as long as tests still pass when branch is brought up to date.

I'm assuming the reason to fix the `_from_dict` instead of the `_to_dict` is in case someone has already saved one of these? Otherwise, the extra nesting under `'data'` doesn't seem particularly useful.

PR #232, opened 2023-09-12
Fix conda-cron

IAlibay:


richardjgowers (approved):

.github/workflows/conda_cron.yaml
```diff
@@ -1,5 +1,8 @@
 name: "conda_cron"
 on:
+  pull_request:
+    branches:
+      - main
```
dwhswenson, line 5: Missed this when it happened, but @IAlibay, is there a reason we're now running `conda_cron` in PRs?

(This PR got opened and merged in about an hour, while I was sleeping, so I really didn't have a chance to review it at the time.)

IAlibay: I bet you this is a shim I added to test it worked in the PR and then we didn't remove before merge.

IAlibay: commit history indicates that this is the case

PR #234, opened 2023-09-13
[WIP DNM] Per-Unit Execution

dwhswenson:
Opening as PR for better discussion with @dotsdl 

---

Edit: This PR has been redirected so that it merges into #259. This allows you to only see the diffs specific to per-unit execution.

The original purpose of this PR (integrating with `execute_DAG` has been mainly moved into #258. As of #259, things should be enough in place that executors can update to use the new tools. The work left in this PR extends that to the draft the next generation of executors, where units can run in parallel on different machines.


atravitz:
closing, this work is being addressed in https://github.com/OpenFreeEnergy/openfe/pull/2113

PR #236, opened 2023-09-14
Fix quantity array codec

richardjgowers:
Roundtripping of **arrays** of Quantity objects would previously fail

### DIFF
--- a/gufe/tests/test_custom_json.py
+++ b/gufe/tests/test_custom_json.py
@@ -8,6 +8,7 @@ import pathlib
 
 import numpy as np
 import openff.units
+from openff.units import unit
 import pytest
 from numpy import testing as npt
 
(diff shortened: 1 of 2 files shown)


gufe/tests/test_custom_json.py
```diff
@@ -223,8 +225,21 @@ def setup(self):
         ]
 
 
+def test_openff_quantity_array_roundtrip():
```
richardjgowers, line 228: I tried at first to put this in the existing test harness, but because it uses a combination of two codecs (FF Unit and Array) I couldn't figure out the trick.  This test exemplifies the issue well

dwhswenson: Yeah, this is the way to do it. The harness treats each codec as a unit (as in "unit test", not as in "unit kelvin"); this is a problem in the integration of 2 units (as in "integration test", not as in "integration over $x$" -- why does language have to be so overloaded?)

dwhswenson (approved):
lgtm

PR #237, opened 2023-09-15
Adding multiMapping plots

### LINKED ISSUE #233: adding mol name to mapping figure?
https://github.com/OpenFreeEnergy/gufe/blob/441b70a6f716038dfa6c139c89c2ce2ebff6db14/gufe/visualization/mapping_visualization.py#L167C1-L173C6

@richardjgowers 
How about this addition, if we can assume ofe-name is always present?:
```
    d2d.DrawMolecules(
        copies,
        highlightAtoms=atoms_list,
        highlightBonds=bonds_list,
        highlightAtomColors=atom_colors,
        highlightBondColors=bond_colors,
        legends=[copies[i].GetProp("ofe-name"), copies[j].GetProp("ofe-name")]
    )
```
image

RiesBen:
This pull request adds obne nice small feature to gufe mapping visualization, such you can show multiple mappings on one figure:

- adding a Multi mapping grid img function `draw_multiple_mappings` - drafty right now.
image


gufe/visualization/mapping_visualization.py
```diff
@@ -305,3 +307,57 @@ def draw_unhighlighted_molecule(mol, d2d=None):
...
+        res.paste(img,(x,y))
+        x += img.width + buffer_spacing
+        
+        if(i%ncols == 0 and i!=0):
+            y+= img.height + buffer_spacing
+            x=0
+            
+    return res
+
+def draw_multiple_mappings(mappings:List, out_file_path:str=None)->Image:
```
jthorton, line 349: Update the type hints, for mappings and out_file_path, maybe change the second to be `output_file`. We should expose the grid layout settings as well to recreate the example.

jthorton (commented):
I really like this idea, it offers a quick way to see all the edges and mappings in the network without iterating through the network or using the CLI viewer! It just needs a bit more work to enable visuals like the great example. Do you have time to finish this @RiesBen or is this something that should be handed over to me and Hannah?

gufe/visualization/mapping_visualization.py
```diff
@@ -305,3 +307,57 @@ def draw_unhighlighted_molecule(mol, d2d=None):
...
+    """
+        This function generates an image, consisting of multiple mapping visualization.
```
IAlibay, line 351: Since this is a public facing method, it would need better docstring, especially on the "what does it do and what are my parameters".

gufe/visualization/mapping_visualization.py
```diff
@@ -305,3 +307,57 @@ def draw_unhighlighted_molecule(mol, d2d=None):
...
+    """
+    views = []
+    for m in mappings:
+        view = draw_mapping(m._compA_to_compB, m.componentA.to_rdkit(),
```
IAlibay, line 355: This would really need tests of some kind, even if they are smoke tests (although I do note that matplotlib has a series of methods to test image equality that could be useful here).

gufe/visualization/mapping_visualization.py
```diff
@@ -305,3 +307,57 @@ def draw_unhighlighted_molecule(mol, d2d=None):
         bond_colors=[{}],
         highlight_color=red,
     )
+
+
+def _concat_images(imgs, buffer_spacing=4, ncols=3)-> Image:
+    """
+        This helper function is constructing a Grid Image of given imgs.
```
IAlibay, line 314: Even if it's private, this really could use with a better explanation of what the method does.

gufe/visualization/mapping_visualization.py
```diff
@@ -1,7 +1,9 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
-from typing import Any, Collection, Optional
+from typing import Any, Collection, Optional, List
 from itertools import chain
+from PIL import Image, ImageOps
```
IAlibay, line 5: I believe pillow now replaces PIL, which should be used instead.

Additionally - we would need to work out what we want to do with this additional dependency. Probably would need to be an optional.

IAlibay (changes requested):
Various comments on dependencies, tests, and documentation.

gufe/visualization/mapping_visualization.py
jthorton: If we use pillow I think rdkit has that as part of its recipe.

gufe/visualization/mapping_visualization.py
```diff
@@ -305,3 +307,57 @@ def draw_unhighlighted_molecule(mol, d2d=None):
         bond_colors=[{}],
         highlight_color=red,
     )
+
+
+def _concat_images(imgs, buffer_spacing=4, ncols=3)-> Image:
```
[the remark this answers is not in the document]
RiesBen: I agree, this PR is a very old one and I wanted to bring it to your attention as I think it could be a handy feature. 

hm... I wonder if this function should be used in line 360 somehow? I also wonder if the code developed away from the initial implementation?

RiesBen:
> I really like this idea, it offers a quick way to see all the edges and mappings in the network without iterating through the network or using the CLI viewer! It just needs a bit more work to enable visuals like the great example. Do you have time to finish this @RiesBen or is this something that should be handed over to me and Hannah?

Hi @jthorton , I sadly will not have time to continue here :(

PR #241, opened 2023-09-27
adding mol name labels from ofe-name to mapping vis

RiesBen:
This pull request adds one nice small features to gufe mapping visualization:
- labeling mapped molecules with their `ofe-name`
image


mikemhenry (changes requested):
Can you update the documentation to add a note that `ofe-name` is used to label the molecules? Also is this something that users might want to just turn off? Could the `ofe-name` be really ugly or is that completely in the users hands?

richardjgowers:
@mikemhenry I think there's also some magic where `Mol.GetProp('_Name')` (the semi canonical way of labelling a name in rdkit) is converted to `ofe-name` (and vice versa)

richardjgowers (approved):

PR #249, opened 2023-11-07
Create python-publish.yml

### LINKED ISSUE #247: Update the PyPi release of gufe?
We have a PyPi indexed release but it's currently stuck at 0.4.

Given that some tools like Lomap might directly just require `gufe`, it might be nice to make pypi releases too so that other packages can have pypi only releases.

richardjgowers:
Fixes #247

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,39 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip versioningit
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
```
richardjgowers, line 24: something I wasn't sure on, should this just use the same micromamba checkout we use for CI?  Or is this good enough for package building?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        python -m pip install --upgrade pip versioningit
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
IAlibay, line 39: the recommend way to do this is to use OIDC now, so you don't use the token/user and just make it a trusted workflow on pypi

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip versioningit
```
IAlibay, line 31: you should not need this, this is pulled in by build_requires

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+        pip install build
```
IAlibay, line 32: I would just install `pipx` here and run `pipx run build` below, env isolation can be nice to make sure you didn't pull anything weird earlier (you shouldn't but I get really skeptical of github python setup actions).

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
```
IAlibay, line 26:
```suggestion
      uses: actions/setup-python@v4
```

.github/workflows/python-publish.yml
IAlibay, line 24:
```suggestion
    - uses: actions/checkout@v4
```

IAlibay: If you're building python wheels for pypi I'd got pure Python - so this route

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip versioningit
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
IAlibay, line 36: Any reason to not pin to a specific release version?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Upload Python Package
+
+on:
+  release:
```
IAlibay, line 12: Do you want to also do a testpypi release on tags?

.github/workflows/python-publish.yml
richardjgowers: I just clicked the action button and it gave me this

.github/workflows/python-publish.yml
richardjgowers: Doesn't seem necessary?

.github/workflows/python-publish.yml
richardjgowers: So this thing instead? https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/

IAlibay: yeah

.github/workflows/python-publish.yml
IAlibay: I'd only do it so you can save yourself a bugfix push, it's fine if you don't mess with packaging too often though.

I would however recommend that you do pull down the package and run tests post deploy though - that way you don't have to think about "will it work".

I wrote a re-usable action for this here: https://github.com/MDAnalysis/pypi-deployment/tree/main

You can see an example of it here: https://github.com/MDAnalysis/MDAnalysisData/blob/master/.github/workflows/deploy.yaml

(you can remove the test deployment if you don't want it).

Advantage of the re-usable action is that you only have to update a single place and everything will keep working.

PR #255, opened 2023-11-20
Settings setting shortcut thing

### LINKED ISSUE #254: setting settings shortcuts
Rather than remembering entire settings hierarchies, it would be nice to simply set a field name at a top level Settings object and have the Settings object figure out what I meant (if it were ambiguous)

richardjgowers:
fixes #254

Adds SettingsBaseModel.find_field and set_field to find and set a parameter name from anywhere within the nested Settings

### DIFF
--- a/gufe/settings/models.py
+++ b/gufe/settings/models.py
@@ -27,6 +27,15 @@ except ImportError:
     )
 
 
+def _find_key_walk(d: dict, target: str, prefix=()):
+    # generator for recursively walking Settings for find_field
+    for k, v in d.items():
+        if k == target:
+            yield prefix + (k,)
+        elif isinstance(v, dict):
+            yield from _find_key_walk(v, target, prefix=prefix + (k,))
+
+
 class SettingsBaseModel(DefaultModel):
     """Settings and modifications we want for all settings classes."""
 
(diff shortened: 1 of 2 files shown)


mikemhenry (approved):

dwhswenson (changes requested):
I'm not keen on this because I'm concerned about synonyms. I can imagine a user thinking they're setting a temperature with `settings.temp`, but this actually sets a temp file name (especially if they're setting temperature with a string.)

richardjgowers:
```
In [5]: class Foo(BaseModel):
   ...:     temp: FloatQuantity['kelvin']
   ...:

In [6]: f = Foo(temp='10 K')

In [7]: f
Out[7]: Foo(temp=<Quantity(10.0, 'kelvin')>)

In [8]: f = Foo(temp='10')
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[8], line 1
----> 1 f = Foo(temp='10')

File ~/mambaforge/envs/openfe_env/lib/python3.10/site-packages/pydantic/main.py:341, in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for Foo
temp
  Cannot convert from 'dimensionless' (dimensionless) to 'kelvin' ([temperature]) (type=type_error.dimensionality; units1=dimensionless; units2=kelvin; dim1=dimensionless; dim2=[temperature]; extra_msg=)
```

@dwhswenson I'm not sure we have to be worried about stray strings on Quantities.  It will only "work" if the string is also valid as a Quantity, so not paths, and simple ints/floats are "upcasted" into Quantities easily.  I think you're rightly worried about bad variable names here (temperature is much better than temp for example), but that's not really made much worse by this feature

dwhswenson:
@richardjgowers : You've got the mistake I'm concerned about backwards. I'm worried about stray "quantities" on strings (when given as string). If I thought `temp` controlled temperature, but it actually controlled a tempfile name, I could easily do this:

```python
In [1]: from pydantic import BaseModel

In [2]: class Foo(BaseModel):
   ...:     temp: str = "tempfile.dat"  # a filename
   ...:

In [3]: f = Foo(temp="10 K")

In [4]: f
Out[4]: Foo(temp='10 K')
```

That works just fine, and could lead to confusion and even incorrect science. Sure, it is user error, but we're giving them the gun and making it easier to point it at their foot.

PR #257, opened 2023-12-01
External resource updates

dwhswenson:
Two things here:

1. Add equality checks for `ExternalResource`s
2. Fix a potential problem when storing a path for `FileStorage` into a nested directory.

These were part of the overall work I've been doing on StagingPaths and the bigger storage tasks, but this is a small, self-contained PR that can be reviewed pretty easily. I thought it might be easier to review this first.

### DIFF
--- a/gufe/storage/externalresource/filestorage.py
+++ b/gufe/storage/externalresource/filestorage.py
@@ -20,6 +20,12 @@ class FileStorage(ExternalStorage):
     def _exists(self, location):
         return self._as_path(location).exists()
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, FileStorage)
+            and self.root_dir == other.root_dir
+        )
+
     def _store_bytes(self, location, byte_data):
         path = self._as_path(location)
         directory = path.parent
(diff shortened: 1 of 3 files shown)


gufe/storage/externalresource/filestorage.py
```diff
@@ -20,6 +20,12 @@ def __init__(self, root_dir: Union[pathlib.Path, str]):
     def _exists(self, location):
         return self._as_path(location).exists()
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, FileStorage)
```
richardjgowers, line 25: a bit nickpicky, but technically objects are meant to return `NotImplemented` in this case, so that the other object can have a shot at defining `__eq__`, see https://stackoverflow.com/a/42588737

gufe/storage/externalresource/memorystorage.py
```diff
@@ -26,6 +26,9 @@ def _delete(self, location):
                 f"Unable to delete '{location}': key does not exist"
             )
 
+    def __eq__(self, other):
+        return self is other
```
richardjgowers, line 30: iirc this is the default object behaviour, but I guess it isn't harming anyone having it here

richardjgowers (approved):

PR #258, opened 2023-12-01
[WIP] Use `StorageManager` in `execute_DAG`

dwhswenson:
This is a subset of the work in #234, split off to facilitate review. This PR will merge into #186 (making it easier to see the additions specific to the PR, but meaning that this should be merged before that). It also includes the work in #257 (one of the old tests fails without that).

This PR integrates `StorageManager` into `execute_DAG`. Note that this still isn't a complete replacement of the storage system: a follow-up PR will handle serialization of `StagingPath`s.

TODO:

- [x] Check that we can exactly reproduce the old behavior with the `ReproduceOldBehaviorStorageManager`; I think some paths aren't quite right.
- [x] Add some tests that the directories are as expected in the `ExecutionStorageDemoTest` subclasses
- [ ] Revisit when/where `dag_label` is given (also needed in [#186](https://github.com/OpenFreeEnergy/gufe/pull/186))
- [x] Switch it so the context managers return Context (executor shouldn't need to assemble that)


gufe/protocols/protocoldag.py
```diff
@@ -17,6 +17,13 @@
     ProtocolUnit, ProtocolUnitResult, ProtocolUnitFailure, Context
 )
 
+from ..storage.storagemanager import StorageManager
+from ..storage.externalresource.filestorage import FileStorage
+from ..storage.externalresource.base import ExternalStorage
```
richardjgowers, line 22: can we fix the namespace of `storage` so these are all available at the same level?  `from ..storage import StorageManager, FileStorage, etc`

gufe/protocols/protocoldag.py
```diff
@@ -345,9 +352,67 @@ def _from_dict(cls, dct: dict):
...
+            keep_scratch=keep_scratch,
+            keep_shared=keep_shared,
+            keep_staging=True,
+            keep_empty_dirs=True,
+            staging=Path(""),  # use the actual directories as the staging
+        )
+        return storage_manager
+
+
 def execute_DAG(protocoldag: ProtocolDAG, *,
```
richardjgowers, line 413: are you keeping this for comparison for the PR?  I don't think we need to keep the old behaviour around

gufe/protocols/protocoldag.py
```diff
@@ -385,52 +450,92 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
...
+        keep_scratch=keep_scratch
+    )
+    dag_label = shared_basedir.name
+    return new_execute_DAG(protocoldag, dag_label, storage_manager,
+                           raise_error, n_retries)
+
+
+def new_execute_DAG(  # TODO: this is a terrible name
+    protocoldag: ProtocolDAG,
+    dag_label: str,
```
richardjgowers, line 469: why is a DAG label now required?

and why can't the label just be taken off the input DAG?

gufe/protocols/protocoldag.py
dwhswenson: I think we told Joe that the API was stable enough that he could use it in workflows. So yeah, I spent **a lot** of time ensuring that there would be a code path that doesn't break the current API.

gufe/protocols/protocoldag.py
dwhswenson: Breaks current behavior, where the equivalent of the DAG label is given by the user as the `-o` from quickrun.

gufe/tests/test_protocol.py
```diff
@@ -696,7 +696,8 @@ def test_execute_DAG_retries(solvated_ligand, vacuum_ligand, tmpdir):
         # we did 3 retries, so 4 total failures
         assert len(r.protocol_unit_results) == 5
         assert len(r.protocol_unit_failures) == 4
-        assert len(list(shared.iterdir())) == 5
+        assert len(list(shared.iterdir())) == 0
+        # assert len(list(shared.iterdir())) == 5
```
dwhswenson, line 700: (Extremely minor) breaking change: Previously, this left behind 5 empty shared directories. In the current implementation, empty directories are, by default, automatically deleted, so we have no directories within the `shared` space. Had these directories contained any files, you would still find all 5.

There's a parameter to keep empty directories, but the effect will also be to keep empty subdirectories, which I didn't think was the desired default behavior. My assumption is that it will be relatively rare to keep `shared` anyway, because there should be something in settings to determine whether any given file is `scratch`/`shared`/`permanent`, and anything intended for long-term storage will be put in `permanent`.

dwhswenson (commented):
Commenting on a couple of breaking changes these PRs introduce that people should be aware of.

gufe/protocols/protocoldag.py
```diff
@@ -385,52 +450,92 @@ def execute_DAG(protocoldag: ProtocolDAG, *,
...
+                # translate each `ProtocolUnit` in input into corresponding
+                # `ProtocolUnitResult`
+                inputs = _pu_to_pur(unit.inputs, results)
+
+                label = storage_manager.make_label(dag_label, unit.key,
+                                                   attempt=attempt)
+                with dag_ctx.running_unit(
+                    dag_label, unit.key, attempt=attempt
+                ) as context:
+                    _logger.info("Starting unit {label}")
```
richardjgowers, line 523: missing an f-string thingy here

gufe/protocols/protocolunit.py
```diff
@@ -31,7 +33,8 @@ class Context:
 
     """
     scratch: PathLike
-    shared: PathLike
+    shared: StagingRegistry
+    permanent: StagingRegistry
```
richardjgowers, line 37: I think these are meant to be stagingpaths not registries?

atravitz:
closing, this work is being addressed in https://github.com/OpenFreeEnergy/openfe/pull/2113

PR #259, opened 2023-12-01
[WIP] Serialization of StagingPaths

dwhswenson:
This is a subset of the work in #234, split off to facilitate review. This PR will merge into #258 (making it easier to see the additions specific to the PR, but meaning that this should be merged before that)

This handles the serialization of `StagingPath` objects. (detailed description will be coming in a future edit)

TODO:

- [x] Add user story where the type of external storage is changed
- [x] Add user story where the a new JSONCodec is added after the `StagingSerialization` is created.
- [x] Docstrings
- [x] Attach to the StorageManager

### DIFF
--- a/gufe/storage/stagingserialization.py
+++ b/gufe/storage/stagingserialization.py
@@ -2,6 +2,7 @@ from gufe.tokenization import JSON_HANDLER
 from gufe.custom_json import JSONCodec, JSONSerializerDeserializer
 from .stagingdirectory import StagingPath
 
+
 class StagingPathSerialization:
     # TODO: where should this go? I think maybe on the storage manager
 
--- a/gufe/storage/stagingserialization.py
+++ b/gufe/storage/stagingserialization.py
@@ -1,6 +1,6 @@
 from gufe.tokenization import JSON_HANDLER
 from gufe.custom_json import JSONCodec, JSONSerializerDeserializer
-from .stagingdirectory import StagingPath
+from .stagingregistry import StagingPath
 
 
 class StagingPathSerialization:
--- a/gufe/storage/stagingserialization.py
+++ b/gufe/storage/stagingserialization.py
@@ -4,12 +4,20 @@ from .stagingregistry import StagingPath
 
 
 class StagingPathSerialization:
-    # TODO: where should this go? I think maybe on the storage manager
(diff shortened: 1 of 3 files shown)


gufe/storage/stagingserialization.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, manager):
+        self.manager = manager
+        self.codec = JSONCodec(
+            cls=StagingPath,
+            to_dict=self.to_dict,
+            from_dict=self.from_dict,
+        )
+        self.refresh_handler()
+
+    def refresh_handler(self):
```
richardjgowers, line 63: what's the need for the refresh_handler to be a method rather than just inlined into `__init__`?

or what's the harm in just adding to `JSONCodec` on the import of this module?

dwhswenson: > what's the need for the refresh_handler to be a method rather than just inlined into __init__?

In principle, the codecs in `JSON_HANDLER` can change. For example, an external package can register additional handlers. Since So we at least want the ability to refresh. In fact, I was thinking about whether the `encoder`/`decoder` properties maybe should call `refresh_handler`.

> or what's the harm in just adding to `JSONCodec` on the import of this module?

I'm not sure what you mean by "adding to `JSONCodec`." Could you clarify?

richardjgowers: Sorry I meant JSON_HANDLER.  Other Codecs are defined and just appended onto the HANDLER as they are defined, so the HANDLER is "batteries included".  Is there a reason this Codec can't have the same treatment?

dwhswenson: 1. We don't know the storage paths at import time, so this is literally impossible.
2. We may have more than one set of `shared`/`permanent` (especially `permanent`) defined in a given process. Imagine that I'm working with data from multiple campaigns. Imagine furthermore that I've use the `keep_shared` option so that my shared stuff is also still accessible. This precludes any kind of serialization that relies on global state. [This test](https://github.com/OpenFreeEnergy/gufe/blob/e39a79026c35b0b9a8643e4f8c1001bd81b74cc3/gufe/tests/storage/test_stagingserialization.py#L198) is specifically to prevent us from making this mistake.

PR #260, opened 2023-12-18
change signature of Protocol.create to have a list of ComponentMappings

richardjgowers:
was previously a `dict[str, ComponentMapping]` where the strings were arbitrary labels.  These arbitrary labels were superfluous.  Instead can use the arbitrary labels on ChemicalSystems to label components.  (ComponentMappings can (and should) be matched against components in ChemicalSystem using the Component eq operations.)

This now accepts either a single ComponentMapping (covers most cases), or None (covers ABFE), or a list (handles future things?)

also affects Transformation object, which was essentially wrapper around Protocol + ChemicalSystem

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -171,7 +171,7 @@ class Protocol(GufeTokenizable):
         *,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Union[dict[str, ComponentMapping], None],
+        mapping: Optional[Union[ComponentMapping, list[ComponentMapping], None]],
         extends: Optional[ProtocolDAGResult] = None,
         name: Optional[str] = None,
         transformation_key: Optional[GufeKey] = None
@@ -193,9 +193,9 @@ class Protocol(GufeTokenizable):
             The starting `ChemicalSystem` for the transformation.
         stateB : ChemicalSystem
             The ending `ChemicalSystem` for the transformation.
-        mapping : Optional[dict[str, ComponentMapping]]
+        mapping : Optional[ComponentMapping | list[ComponentMapping]]
             Mappings of e.g. atoms between a labelled component in the
-             stateA and stateB `ChemicalSystem` .
+            stateA and stateB `ChemicalSystem` .
         extends : Optional[ProtocolDAGResult]
             If provided, then the `ProtocolDAG` produced will start from the
             end state of the given `ProtocolDAGResult`. This allows for
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -135,15 +135,16 @@ class Protocol(GufeTokenizable):
         self,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Optional[dict[str, ComponentMapping]] = None,
+        mapping: Optional[Union[ComponentMapping, list[ComponentMapping]]] = None,
         extends: Optional[ProtocolDAGResult] = None,
     ) -> list[ProtocolUnit]:
         """Method to override in custom :class:`Protocol` subclasses.
 
-        This method should take two `ChemicalSystem`s, and optionally a
-        dict mapping string to ``ComponentMapping``, and prepare a collection of ``ProtocolUnit`` instances
-        that when executed in order give sufficient information to estimate the
-        free energy difference between those two `ChemicalSystem`s.
+        This method should take two `ChemicalSystem`s, and optionally one or
+        more ``ComponentMapping`` objects, and prepare a collection of
+        ``ProtocolUnit`` instances that when executed in order give sufficient
+        information to estimate the free energy difference between those two
+        `ChemicalSystem`s.
 
         This method should return a list of `ProtocolUnit` instances.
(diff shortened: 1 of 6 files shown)


gufe/tests/test_protocol.py
```diff
@@ -112,7 +112,7 @@ def _create(
         self,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Optional[dict[str, ComponentMapping]] = None,
+        mapping: Optional[list[ComponentMapping]] = None,
```
richardjgowers, line 115: Looking at the diff on this: https://github.com/OpenFreeEnergy/openfe/pull/666

Maybe we could go further and allow `Union[ComponentMapping, list[ComponentMapping]]`, i.e. one or more mappings.  It might be more intuitive to allow a single mapping to be provided, rather than more often than not a list with a single entry.

dwhswenson (commented):
Minor suggestion, although probably doesn't need to be blocking: Maybe rename `mapping` to `mappings`? This would solve confusion around the idea that this could take more than 1 but for existing protocols only takes 1.

richardjgowers:
@dwhswenson I think I'm going the other way, I'll make the `mapping` argument take a single instance (since this is currently all cases?) and not have to explain that technically we've thought about the idea that multiple mappings might one day be possible/necessary.

IAlibay:
Especially as we move towards multi-state perturbations, multiple mappings may be important here. I guess the question is - is flexibility better than not?

richardjgowers:
@IAlibay sorry should've been clearer, it's taking **either** a single Mapping or many, so zero, one or more.  Then you don't have to (but can) pass a list of a single mapping whenever you do a RFE.

IAlibay:
Ah so Optional[Union[Mapping, List[Mapping]]? That would be great!

IAlibay:
I should have just read the code instead of reading my emails, sorry...

gufe/protocols/protocol.py
```diff
@@ -135,15 +135,16 @@ def _create(
         self,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Optional[dict[str, ComponentMapping]] = None,
+        mapping: list[ComponentMapping],
```
IAlibay, line 138: Depending on the conversation below - it would be preferred if this was Optional i.e. expect None rather than a zero length list.

```suggestion
        mapping: Optional[list[ComponentMapping]],
```

gufe/protocols/protocol.py
```diff
@@ -170,7 +171,7 @@ def create(
         *,
         stateA: ChemicalSystem,
         stateB: ChemicalSystem,
-        mapping: Union[dict[str, ComponentMapping], None],
+        mapping: Optional[Union[ComponentMapping, list[ComponentMapping], None]],
```
IAlibay, line 174:
```suggestion
        mapping: Optional[Union[ComponentMapping, list[ComponentMapping]]],
```
Doesn't Optional by default define a `None` case?

gufe/protocols/protocol.py
```diff
@@ -192,9 +193,9 @@ def create(
             The starting `ChemicalSystem` for the transformation.
         stateB : ChemicalSystem
             The ending `ChemicalSystem` for the transformation.
-        mapping : Optional[dict[str, ComponentMapping]]
+        mapping : Optional[ComponentMapping | list[ComponentMapping]]
```
IAlibay, line 196:
```suggestion
        mapping : Optional[ComponentMapping, list[ComponentMapping]]
```

As per the Python docs Optional is a special type of Union, Union[A, B] is A | B.
I.e. Union[A | B] would be the same as saying Union[Union[A, B]] rather than Union[A, B]

gufe/protocols/protocol.py
```diff
@@ -214,6 +215,14 @@ def create(
             A directed, acyclic graph that can be executed by a `Scheduler`.
 
         """
+        # coerce mapping argument into list for _create signature
+        # top level "create" allows more intuitive input
+        # internal "_create" provides guaranteed type
+        if mapping is None:
```
IAlibay, line 221: Both the AFE and MD protocols expect `None` (although neither checks for it right now).

In practice there will be more Protocols without mappings than Protocols with mappings - i.e. it feels like we're bending things to be purely RFE orientated when it would make a lot more sense (especially in a Pythonic manner) for a protocol author to expect `None` rather than `empty list`

IAlibay (changes requested):
A couple of things - particularly I would much prefer here if mapping was Optional on create, not an enforced list of zero width when there are no mappings.

gufe/protocols/protocol.py
IAlibay: https://docs.python.org/3/library/typing.html#typing.Optional

gufe/protocols/protocol.py
richardjgowers: that syntax is 3.10 only, I think we're still supporting 3.9 per nep29

IAlibay: @richardjgowers https://docs.python.org/3.9/library/typing.html#typing.Union 

That's the correct syntax for py3.9?

IAlibay (approved):

PR #262, opened 2023-12-27
Update solventcomponent to encompass all composition properties

### LINKED ISSUE #261: Update SolventComponent properties to include those that define how the system is built
Currently we have a mismatch where we have some details about how solvation happens in the SolvationSettings and some in SolventComponent.

My proposal is that **anything that defines the molecular composition of the system** should be defined in the the SolventComponent.

This includes:
- The number of waters that will be added
  - Including options such as the `padding` & `mass density` used to add the waters
  - Including options such as how the box is defined
    - `Box vectors`
    - `Box shape`
- The type of `ions` & `concentration` that will be added (done)
- The solvent `smiles` (done)

By comparison, **anything that defines the implementation details** should be in the `SolvationSettings`:
- `water model`
- `tolerance` for any heuristic in water addition (see packmol)
- solvation backend (OpenMM modeller, packmol, leap, gmx solvate, etc...)
- centering the solute

IAlibay:
Fixes #261

### DIFF
--- a/gufe/components/solventcomponent.py
+++ b/gufe/components/solventcomponent.py
@@ -9,6 +9,7 @@ from .component import Component
 
 _CATIONS = {'Cs', 'K', 'Li', 'Na', 'Rb'}
 _ANIONS = {'Cl', 'Br', 'F', 'I'}
+_ALLOWED_BOX_TYPES = {'cube', 'dodecahedron', 'octahedron'}
 
 
 # really wanted to make this a dataclass but then can't sort & strip ion input
@@ -164,7 +165,7 @@ class SolventComponent(Component):
                 raise ValueError(errmsg)
 
             if solvent_padding.m < 0:
-                errmsg = ("solvent padding must be positive, "
+                errmsg = ("solvent_padding must be positive, "
                           f"got: {solvent_padding}")
                 raise ValueError(errmsg)
 
@@ -175,7 +176,7 @@ class SolventComponent(Component):
             if (not isinstance(solvent_density, unit.Quantity) or
                 not solvent_density.is_compatible_with(unit.gram / unit.milliliter)):
                 errmsg = ("solvent_density must be given in units compatible "
-                          f"g/mL, got: {solvent_density}")
+                          f"with g/mL, got: {solvent_density}")
                 raise ValueError(errmsg)
 
             if solvent_density.m < 0:
@@ -187,8 +188,10 @@ class SolventComponent(Component):
 
         # Validate box shape
         if box_shape is not None:
-            if box_shape.lower() not in ['cube', 'dodecahedron', 'octahedron']:
-                raise ValueError(f"Unknown box_shape passed, got: {box_shape}")
+            if box_shape.lower() not in _ALLOWED_BOX_TYPES:
+                errmsg = (f"Unknown box_shape passed, got: {box_shape}, "
+                          f"must be one of {', '.join(_ALLOWED_BOX_TYPES)}")
+                raise ValueError(errmsg)
 
         self._box_shape = box_shape
 
--- a/gufe/components/solventcomponent.py
+++ b/gufe/components/solventcomponent.py
@@ -78,7 +78,7 @@ class SolventComponent(Component):
         box_vectors : openff.units.unit.Quantity, optional
           Vectors defining the solvent box. Cannot be defined alongside
           `box_shape`. Cannot be defined alongside both `num_solvent` and
-          `solvent_density`. Default `None
+          `solvent_density`. Default `None`
 
         Raises
         ------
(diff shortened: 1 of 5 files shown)


gufe/tokenization.py
```diff
@@ -535,7 +537,7 @@ def to_keyed_dict(self, include_defaults=True) -> Dict:
 
         if not include_defaults:
             for key, value in self.defaults().items():
-                if dct.get(key) == value:
+                if np.all(dct.get(key) == value):
```
IAlibay, line 540: This is probably the wrong way to do this

gufe/components/solventcomponent.py
```diff
@@ -22,39 +56,94 @@ class SolventComponent(Component):
...
+        solvent_density : openff.units.unit.Quantity, optional
+          Target density of solvated systems. Must be defined with units
+          compatible with g / mL, e.g. "850 * unit.kilogram / unit.meter**3".
+          Cannot be defined alongside `num_solvent` if either `box_vector`
+          or `solvent_padding` are defined. Default `None`
+        box_shape : str, optional
+          Defined the shape of the solvent box being built. Can be one of
+          'cube', 'dodecahedron', and 'octahedron'. Cannot be defined alongside
+          `box_vectors`. Default 'cube'
+        box_vectors : openff.units.unit.Quantity, optional
```
IAlibay, line 111: missing box_size

gufe/components/solventcomponent.py
```diff
@@ -22,39 +56,94 @@ class SolventComponent(Component):
...
+                 num_solvent: Optional[int] = None,
                  positive_ion: str = 'Na+',
                  negative_ion: str = 'Cl-',
                  neutralize: bool = True,
-                 ion_concentration: unit.Quantity = 0.15 * unit.molar):
+                 ion_concentration: Optional[unit.Quantity] = 0.15 * unit.molar,
+                 solvent_padding: Optional[unit.Quantity] = 1.2 * unit.nanometer,
+                 solvent_density: Optional[unit.Quantity] = None,
+                 box_shape: Optional[Literal['cube', 'dodecahedron', 'octahedron']] = 'cube',
+                 box_vectors: Optional[unit.Quantity] = None,):
```
richardjgowers, line 74: Once we get to storing boxes, I'm not sure it makes sense to store the `box_shape` independently of the vectors, the vectors can/should define the shape.

gufe/components/solventcomponent.py
```diff
@@ -22,39 +56,94 @@ class SolventComponent(Component):
...
     def __init__(self, *,  # force kwarg usage
                  smiles: str = 'O',
+                 num_solvent: Optional[int] = None,
                  positive_ion: str = 'Na+',
                  negative_ion: str = 'Cl-',
                  neutralize: bool = True,
-                 ion_concentration: unit.Quantity = 0.15 * unit.molar):
+                 ion_concentration: Optional[unit.Quantity] = 0.15 * unit.molar,
+                 solvent_padding: Optional[unit.Quantity] = 1.2 * unit.nanometer,
+                 solvent_density: Optional[unit.Quantity] = None,
```
richardjgowers, line 72: this is probably worth including, as it *does* affect the reduced potential

richardjgowers (changes requested):
I'm not sure this is fitting the model we've got currently.  `ChemicalSystem` is meant to capture things that can affect the free energy differences between different `ChemicalSystem`s. This instead seems to be trying to capture a molecular simulation setup, which is different.  The solvation additions here shouldn't be affecting the measured (D)DG; padding should be enough that isn't affecting (D)DG, box shape should also be unimportant.

I think the fields num_solvent / padding / vectors are implementation details and should be in a Settings object not here.  E.g. if I ran `ChemicalSystem(w/ LigA) -> ChemicalSystem(w/ LigB)` using a triclinic box, and `ChemicalSystem(w/ LigA) -> ChemicalSystem(w/ LigB)` with a rectangular box, these two `ChemicalSystem(w/ LigA)` are identical.

gufe/components/solventcomponent.py
richardjgowers, line 111: Putting in `box_vectors` will be needed for explicit solvents, but I don't think we'll need to backport that field to this definition

IAlibay:
@richardjgowers solvation, particularly the number of waters *does* affect both the dH (which we will want to be able to calculate), and the *dG* under certain circumstances.

gufe/components/solventcomponent.py
IAlibay: Not everyone can handle dealing with both - it's too much to ask users to provide box vectors a priori.

gufe/components/solventcomponent.py
IAlibay: I'm not sure what you mean here - density is a function of N and volume, everything here defines both of those things.

PR #265, opened 2024-01-13
Adding CA ions to parsing list - proteincomponent.py

RiesBen:
I found this missing Ion reccently . We should add it,, as it is frequently used.

### DIFF
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -51,7 +51,7 @@ _CHIRALITY_STR_TO_RDKIT = {
 }
 
 
-negative_ions = ["F", "CL", "Br", "I"]
+negative_ions = ["F", "CL", "BR", "I"]
 positive_ions = ["NA", "MG", "CA", "ZN"]
 
 
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -52,7 +52,7 @@ _CHIRALITY_STR_TO_RDKIT = {
 
 
 negative_ions = ["F", "CL", "Br", "I"]
-positive_ions = ["NA", "MG", "ZN"]
+positive_ions = ["NA", "MG", "CA", "ZN"]
 
 
 class ProteinComponent(ExplicitMoleculeComponent):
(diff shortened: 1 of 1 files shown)


gufe/components/proteincomponent.py
```diff
@@ -51,8 +51,8 @@
 }
 
 
-negative_ions = ["F", "CL", "Br", "I"]
-positive_ions = ["NA", "MG", "ZN"]
+negative_ions = ["F", "CL", "BR", "I"]
+positive_ions = ["NA", "MG", "CA", "ZN"]
```
IAlibay, line 55: Would it make sense to just add every mono and divalent ions that are defined in the AMBER ffxml?

richardjgowers: We're probably replacing this entirely soon, so a small fix is ok for now. I'd be more interested in a test to encode the fix/behaviour

hannahbaumann:
I checked this new version of the ProteinComponent against the old one and both give the same formal charges for the atoms and same bond order for the bonds.

gufe/components/proteincomponent.py
```diff
@@ -204,15 +217,21 @@ def _from_openmmPDBFile(cls, openmm_PDBFile: Union[PDBFile, PDBxFile],
...
+                if atom_name.upper() in positive_ions:
                     fc = default_valence  # e.g. Sodium ions
-                elif atom_name in negative_ions:
+                elif atom_name.upper() in negative_ions:
                     fc = - default_valence  # e.g. Chlorine ions
+                elif atom_name.strip(string.digits).upper() in positive_ions:
+                    # catches cases like 'CL1' as name
+                    fc = default_valence
+                elif atom_name.strip(string.digits).upper() in negative_ions:
+                    fc = - default_valence
```
RiesBen, line 234: this could be simplified, right?

```
                if atom_name.strip(string.digits).upper() in positive_ions:
                    # catches cases like 'CL1' as name
                    fc = default_valence
                elif atom_name.strip(string.digits).upper() in negative_ions:
                    fc = - default_valence
```

RiesBen (commented):
looks pretty good to me only one question :)

gufe/components/proteincomponent.py
richardjgowers: hmm yeah you're right here

PR #266, opened 2024-01-15
makes AtomMappers subclass from GufeTokenizable

richardjgowers:


dwhswenson (approved):
I'm in favor of this API break (while we can do it), but note that it is a significant API break for all existing `AtomMapper`s. They must implement several additional methods that as `GufeTokenizable`s. I'd recommend immediately making PRs in downstream libraries that link back here.

PR #278, opened 2024-01-29
improve 2d vis of mappings

richardjgowers:
previously would independently calculate 2D vis for each molecule, then RMS align these

now constrains second molecule's 2d representation to match that of the first molecule

overall improves the 2d vis generated, esp. when dealing with rotamers

before: (here atom 9L maps to atom 9R.  rdkit however assigns a more traditional sulfonamide layout)

image (3)

after: (atom 9L & 9R are constrained.  This results in the H's having less layout room, but makes it clear why they didn't map)

mapping_fixed

### DIFF
--- a/gufe/visualization/mapping_visualization.py
+++ b/gufe/visualization/mapping_visualization.py
@@ -4,6 +4,7 @@ from typing import Any, Collection, Optional
 from itertools import chain
 
 from rdkit import Chem
+from rdkit import Geometry
 from rdkit.Chem import Draw
 from rdkit.Chem import AllChem
 
@@ -102,7 +103,7 @@ def _draw_molecules(
     atom_colors: Collection[dict[Any, tuple[float, float, float, float]]],
     bond_colors: Collection[dict[int, tuple[float, float, float, float]]],
     highlight_color: tuple[float, float, float, float],
-    atom_mapping: Optional[dict[tuple[int, int], dict[int, int]]] = None,
+    atom_mapping: Optional[dict[int, int]] = None,
 ) -> str:
     """
     Internal method to visualize a molecule, possibly with mapping info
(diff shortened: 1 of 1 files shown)


gufe/visualization/mapping_visualization.py
```diff
@@ -130,11 +131,10 @@ def _draw_molecules(
     highlight_color: Tuple[float, float, float, float]
         RGBA tuple for the default highlight color used in the mapping
         visualization
-    atom_mapping: Dict[Tuple[int,int], Dict[int, int]], optional
-        used to align the molecules to each othter for clearer visualization.
-        The structure contains the indices of the molecules in mols as key
-        Tuple[molA, molB] and maps the atom indices via the value Dict[
-        molA_atom_idx, molB_atom_idx]
+    atom_mapping: dict[int, int], optional
```
richardjgowers, line 134: I'm changing the signature of this, since it's an private/internal function.  Previously it would accept arbitrarily many molecules and mapping, but this was untested/unused.  I've narrowed this down to 1 molecule or 2 molecules with a mapping.

RiesBen: would we need to turn this back, when going multistate?

RiesBen: or would we implement a new function?

RiesBen: just a thought, to think about it.

RiesBen (approved):

IAlibay (commented):
Would it be possible to see an example with a larger R group please? So we know that the hydrogen "squishing" isn't going to lead to more worse visual problems long term?

gufe/visualization/mapping_visualization.py
```diff
@@ -153,15 +153,26 @@ def _draw_molecules(
...
-        AllChem.Compute2DCoords(mol)
-
-    # mol alignments if atom_mapping present
-    for (i, j), atomMap in atom_mapping.items():
-        AllChem.AlignMol(
-            copies[j], copies[i],
-            atomMap=[(k, v) for v, k in atomMap.items()]
-        )
+    # create layout of first molecule
+    # TODO:
```
IAlibay, line 157: Please raise an issue for this current problem and this todo.

I suspect someone's going to stumble upon this with an old version of gufe and we'll forget this ever happened.

IAlibay (changes requested):

Issue #263, opened 2024-01-05
Improve readability of the settings when printed out


IAlibay:
From a call today with @hannahbaumann - it feels like the way in which we look at settings is too messy to be accessible to users. Is there a way we could improve the `__repr__` or something like that to make it prettier?

richardjgowers:
I think `__repr__` is meant to be a single line string, so probably not with that.  `__str__` would make `print(str(settings))` better.  Are you also thinking about the file representation too?

IAlibay:
I think just changing the `print` behaviour would be fine - it's mostly a case of having a nice way to look at the output.

maybe just make `__str__` return `settings.dict()`? (at least pprint on that looks good)

richardjgowers:
yeah `__str__` being `pprint(settings.dict())` seems very useful.  If we're being fancy we could also make that the jupyter hook whatever that's called

PR #283, opened 2024-02-01
fixes issue #263

richardjgowers:
adds _ipython_display_ to SettingsBaseModel

this pretty prints the nested settings

Screenshot_20240201_124005

### DIFF
--- a/gufe/settings/models.py
+++ b/gufe/settings/models.py
@@ -43,7 +43,7 @@ class SettingsBaseModel(DefaultModel):
         arbitrary_types_allowed = False
         smart_union = True
 
-    def _ipython_display_(self, d2d=None):
+    def _ipython_display_(self):
         pprint.pprint(self.dict())
 
     def frozen_copy(self):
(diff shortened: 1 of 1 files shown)


gufe/settings/models.py
```diff
@@ -42,6 +43,9 @@ class Config:
         arbitrary_types_allowed = False
         smart_union = True
 
+    def _ipython_display_(self, d2d=None):
```
IAlibay, line 46: why the `d2d` option?

```suggestion
    def _ipython_display_(self):
```

IAlibay (changes requested):
Just the one question re: `d2d`

gufe/settings/models.py
richardjgowers: just copied this https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/mapping/ligandatommapping.py#L109

IAlibay (approved):

PR #284, opened 2024-02-01
Test against models Vec3/list patch

mattwthompson:
Context https://github.com/openforcefield/openff-models/pull/36


mattwthompson:
Cool, no surprises here

PR #287, opened 2024-02-08
Issue 170 lam validators

richardjgowers:
Adds a check to LigandAtomMapping that indices are possibly within the molecules.  E.g. if I provide index 100 to a molecule with 10 atoms then something isn't correct and I can provide an error immediately.

### DIFF
--- a/gufe/mapping/ligandatommapping.py
+++ b/gufe/mapping/ligandatommapping.py
@@ -45,7 +45,6 @@ class LigandAtomMapping(AtomMapping):
         self._compA_to_compB = componentA_to_componentB
 
         if annotations is None:
-            # TODO: this should be a frozen dict
             annotations = {}
 
         self._annotations = annotations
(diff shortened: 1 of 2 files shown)


gufe/mapping/ligandatommapping.py
```diff
@@ -34,18 +38,31 @@ def __init__(
...
         """
         super().__init__(componentA, componentB)
+
+        # validate compA_to_compB
+        nA = self.componentA.to_rdkit().GetNumAtoms()
+        nB = self.componentB.to_rdkit().GetNumAtoms()
+        for i, j in componentA_to_componentB.items():
+            if not (0 <= i < nA):
+                raise ValueError(f"Got invalid index for ComponentA ({i}); "
+                                 f"must be 0 <= n <= {nA}")
```
hannahbaumann, line 58: Should this be `must be 0 <= n < {nA}`?

gufe/mapping/ligandatommapping.py
```diff
@@ -34,18 +38,31 @@ def __init__(
...
+            if not (0 <= j < nB):
+                raise ValueError(f"Got invalid index for ComponentB ({i}); "
+                                 f"must be 0 <= n <= {nB}")
```
hannahbaumann, line 61: Same as above

gufe/mapping/ligandatommapping.py
richardjgowers: oops yes!

PR #296, opened 2024-03-13
Dep warnings for dict mapping

richardjgowers:
smoothes over a change made in PR #260


mikemhenry (approved):
LGTM! Do we want to add something about backwards compatibility being removed in version 1.x or 2.x? If we are not sure then we can leave it vague like this and add the version where we drop it later

richardjgowers:
@mikemhenry Yeah I wasn't sure when we'd ever have to remove these lines.  It's not that bad to keep in the shim for all of 1.x.  If/when we find a reason this can't stay we can add a deprecation removal date in the message

PR #306, opened 2024-04-17
Adding a generic atom mapping scorer

RiesBen:
this is a generic interface for all openFE scorers.

Todo:
- [x] define Interface
- [x] tests
- [x] tokenizable
- [x] User stories
- [x] Find Update places (LOMAP, Ruler, OpenFE-Setup, Example Scripts and Notebooks)

### DIFF
--- a/gufe/mapping/atom_mapping_scorer.py
+++ b/gufe/mapping/atom_mapping_scorer.py
@@ -1,0 +1,42 @@
+# This code is part of kartograf and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
+
+import abc
+
+from . import AtomMapping
+
+
+class AtomMappingScorer(abc.ABC):
+    """A generic class for scoring Atom mappings.
+    this class can be used for example to build graph algorithm based networks.
+
+    Implementations of this class can require an arbitrary and non-standardised
+    number of input arguments to create.
+
+    Implementations of this class provide the :meth:`.get_score` method
+
+    """
+
+    def __call__(self, mapping: AtomMapping, *args, **kwargs) -> float:
+        return self.get_score(mapping)
+
+    @abc.abstractmethod
+    def get_score(self, mapping: AtomMapping, *args, **kwargs) -> float:
+        """ calculate the score for an  :class:`.AtomMapping`
+            the scoring function returns a value between 0 and 1.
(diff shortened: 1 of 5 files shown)


gufe/mapping/atom_mapping_scorer.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Implementations of this class provide the :meth:`.get_score` method
+
+    """
+
+    def __call__(self, mapping: AtomMapping, *args, **kwargs) -> float:
+        return self.get_score(mapping)
+
+    @abc.abstractmethod
+    def get_score(self, mapping: AtomMapping, *args, **kwargs) -> float:
```
richardjgowers, line 24: rather than allow args/kwargs to this, could this instead be done by having all args/kwargs passed to the init?

so:
``` python
scorer = MyScorer()
score = scorer(mapping, alpha=0.1)

# instead
scorer = MyScorer(alpha=0.1)
score = scorer(mapping)
```
I think the latter is better, unless there's some use case that it doesn't allow?

RiesBen: aehm, yes you are very right here!

gufe/mapping/atom_mapping_scorer.py
```diff
@@ -0,0 +1,42 @@
...
+        ----------
+        mapping: AtomMapping
+            the mapping to be scored
+        args
+        kwargs
+
+        Returns
+        -------
+        float
+            a value between [0,1] where one is a very bad score and 0 a very good one.
```
richardjgowers, line 39: this is reverse right? 1.0 is good 0.0 is bad

RiesBen: right

IAlibay (changes requested):
block pending discussion

gufe/mapping/atom_mapping_scorer.py
```diff
@@ -0,0 +1,43 @@
...
+        ----------
+        mapping: AtomMapping
+            the mapping to be scored
+        args
+        kwargs
+
+        Returns
+        -------
+        float
+            a value between [0,1] where zero is a very bad score and one a very good one.
```
RiesBen, line 40: 1 is perfect, 0 is bad!

RiesBen:
moved to Fe setup structure #346

PR #307, opened 2024-04-17
[WIP]  Adding ligand network planner

RiesBen:
The goal here is to add a generic ligand network planner. 
to be discussed.


RiesBen:
Fe setup structure #346

Issue #314, opened 2024-04-27
ProtocolDAG._iterate_dag_order is unstable


LilDojd:
## Expected Behavior
`ProtocolDAG` dag order and key are stable on roundtrip from one process to another

## Current Behavior
When playing around with random graphs, I stumbled upon some odd behaviour.

There are some very special hard-to-reproduce cases when the `ProtocolDAG` key is unstable due to different ordering of protocol_units in `nx.topological_sort` on different hardware architectures or even different processes. This is related to an old networkx issue: https://github.com/networkx/networkx/issues/1181

## Steps to Reproduce
[test.json](https://github.com/OpenFreeEnergy/gufe/files/15137569/test.json)
```python
from typing import List
from gufe import ProtocolDAG, ProtocolUnit
from gufe.tokenization import is_gufe_dict, modify_dependencies, from_dict
import gufe

def roundtrip_key(pdag):
    return ProtocolDAG.from_dict(pdag.to_dict()).key


def roundtrip_pus(pdag) -> List[ProtocolUnit]:
    return ProtocolDAG.from_dict(pdag.to_dict()).protocol_units


def is_reordered(list1, list2):

    if len(list1) != len(list2):
        return False

    freq1 = {}
    freq2 = {}

    for item in list1:
        freq1[item] = freq1.get(item, 0) + 1

    for item in list2:
        freq2[item] = freq2.get(item, 0) + 1

    if freq1 == freq2:
        return True
    else:
        return False

class DummyUnit(ProtocolUnit):
    def _execute(self):
        return {"foo": "bar"}


with open("../test.json", "r") as rf:
    pus = json.load(rf)
    pus = modify_dependencies(pus, from_dict, is_gufe_dict, mode="decode")

pdag = ProtocolDAG(protocol_units=pus, transformation_key="somekey")


# Roundtrip on other process
import multiprocessing

pool = multiprocessing.Pool(processes=1)
print(f"Original key: {pdag.key}, Roundtrip key: {pool.map(roundtrip_key, [pdag])[0]}")

print("Is reordered:", is_reordered(pdag.protocol_units, pool.map(roundtrip_pus, [pdag])[0]))

pool.close()
```
Outputs:
```
Original key: ProtocolDAG-a05ef9eab9006e410eb4b3ad0e9a3664, Roundtrip key: ProtocolDAG-a7555d4f141783f28da2d2a172b0bf74
Is reordered: True
```

## Context (Environment)
python: 3.10
[...]

PR #315, opened 2024-04-27
Proposed fix to dag ordering

LilDojd:
Not sure how to add tests for this behavior. Please check if you can reproduce the original issue https://github.com/OpenFreeEnergy/gufe/issues/314

### DIFF
--- a/gufe/protocols/protocoldag.py
+++ b/gufe/protocols/protocoldag.py
@@ -46,7 +46,7 @@ class DAGMixin:
 
     @staticmethod
     def _iterate_dag_order(graph):
-        return reversed(list(nx.topological_sort(graph)))
+        return reversed(list(nx.lexicographical_topological_sort(graph, key=lambda pu: pu.key)))
 
     @property
     def name(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


ianmkenney:
I was able to reproduce the original error reported in #314 and the proposed changes do, in fact, correct the ordering (and `GufeKey`). Given that, the failing CI was a little odd.

`assert len(r.protocol_unit_results) == 5` failed with `26 != 5`

The five comes from the results for:

1. `InitializeUnit` (ok)
2. `BrokenSimulationUnit` (0, fail)
3. `BrokenSimulationUnit` (1, fail)
4. `BrokenSimulationUnit` (2, fail)
5. `BrokenSimulationUnit` (3, fail)

The `BrokenProtocol`, however, is defined by 21 `SimulationUnit`s and then a `BrokenSimulationUnit`. If you include "ok" results from the 21 `SimulationUnit`s, you get the above 26. There is an underlying loop within `execute_DAG` that used the `_iterate_dag_order`. Originally, `BrokenSimulationUnit` appeared first for execution and produced the above list of results. Now that the graphs are further sorted by the `GufeKey`s, `BrokenSimulationUnit` appears after (`reverse(sort(["B...", "S..."])`), allowing more results to collect before ultimately failing.

Assuming that the unit names (the key prefix) don't change, then 26 is the proper number to compare to.

ianmkenney (approved):
This LGTM. Nice catch and thanks for the contribution @LilDojd!

PR #320, opened 2024-05-07
adds `LigandNetwork.reduce_graph`

richardjgowers:
move this from Konnektor into gufe, it's very useful


news/added_LigandNetwork_delete_edges.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* added LigandNetwork.remove_edges
```
mikemhenry, line 3: +1 to this getting into the change log

gufe/ligandnetwork.py
```diff
@@ -184,6 +184,36 @@ def enlarge_graph(self, *, edges=None, nodes=None) -> LigandNetwork:
 
         return LigandNetwork(self.edges | set(edges), self.nodes | set(nodes))
 
+    def remove_edges(self, edges: Union[LigandAtomMapping, list[LigandAtomMapping]]) -> LigandNetwork:
+        """Create a new copy of this network with some edges removed
+
+        Note that this will not remove any nodes, potentially resulting in
```
mikemhenry, line 190: How big of deal is this? Is it worth adding the complexity of checking to make sure we don't allow that? I can't remember what the rules are, if we allow creating a network that is disconnected, then I don't think we need a check, but if we require a connected network, then we want to make sure we don't let users put things into a inconsistent state.

RiesBen: This is a good suggestion, I think :)

gufe/ligandnetwork.py
```diff
@@ -184,6 +184,36 @@ def enlarge_graph(self, *, edges=None, nodes=None) -> LigandNetwork:
...
+        to_remove = set(edges)
+        current = set(self.edges)
+
+        # check that all edges to remove are present
+        if extras := to_remove - current:
+            raise ValueError("Some edges weren't already present: "
+                             f"{extras}")
+
+        new_edges = current - to_remove
+
```
RiesBen, line 214:
```suggestion
new_net = LigandNetwork(new_edges, self.nodes)
if not new_net.graph().is_connected():
    raise ValueError("The result is a disconnected Network!")
```

gufe/ligandnetwork.py
richardjgowers: Yeah I was 50/50 on what is best here. Currently disconnected graphs are "allowed" in a data structure sense, but they're often undesirable. It wouldn't be impossible to find and remove orphan nodes, but then the method isn't strictly "remove edges". I think I'll chicken out and add a kwarg that allows either, default being remove orphans.

mikemhenry: > It wouldn't be impossible to find and remove orphan nodes

Ah yes, I was suggesting a more gentle "throw an error if removing this edge creates an orphan" or something so it can stay focused and do what it says on the tin, so I would be happy with kwargs that control things like, raising an error or not if it would create an orphan, and/or remove orphan(s) (this will be tricking if removing a edge creates 2 disconnect graphs, which one gets remove? the one with more nodes? -- maybe we can support the case where if it creates a single orphan, the default is to raise an error, but that can be suppressed, and additionally with another kwarg the orphan can be removed)

gufe/ligandnetwork.py
```diff
@@ -211,7 +211,10 @@ def remove_edges(self, edges: Union[LigandAtomMapping, list[LigandAtomMapping]])
                              f"{extras}")
 
         new_edges = current - to_remove
-
+new_net = LigandNetwork(new_edges, self.nodes)
+if not new_net.graph().is_connected():
```
RiesBen, line 215:
```suggestion
if not self.is_connected():
```

sorry just saw, there is the functionality inside the class

RiesBen (approved):
last suggestion, after that beautiful :)

gufe/ligandnetwork.py
RiesBen: Absolutely correct!

dotsdl:
@atravitz this is now ready for review! I added `LigandNetwork.reduce_graph` as the inverse of the existing `enlarge_graph` method, and this generalizes the functionality that would have been present in `remove_edges`.

gufe/ligandnetwork.py
```diff
@@ -313,9 +348,11 @@ def to_rbfe_alchemical_network(
     #     )
 
     def is_connected(self) -> bool:
-        """Are all ligands in the network (indirectly) connected to each other
+        """Indicates whether all ligands in the network are (indirectly)
```
atravitz, line 351:
```suggestion
        """Indicates whether all ligands in the network are (directly or indirectly)
```
Maybe unnecessary, but I think this is more accurate.

gufe/ligandnetwork.py
```diff
@@ -313,9 +348,11 @@ def to_rbfe_alchemical_network(
...
+        connected to each other.
+
+        A ``False`` value indicates that either some ligands have no edges or
```
atravitz, line 354:
```suggestion
        A ``False`` value indicates that either some nodes have no edges or
```
Just to keep nodes/edges language together.

atravitz (commented):
My hesitation about naming here is that "reduce" in graph theory sort of implies that nodes are preserved. Could this be named `remove` or `remove_from` (to mirror networkx's language here) to be more clear?

Also, is the reason for handling nodes and edges in the same function to avoid making copies of the graph unnecessarily?

dotsdl:
Thanks @atravitz! I'm trying to maintain some symmetry to `LigandNetwork.enlarge_graph`, with `reduce_graph` effectively being the inverse. I'm open to other verbs here; would either `shrink_graph` or `trim_graph` work?

atravitz:
> Thanks @atravitz! I'm trying to maintain some symmetry to `LigandNetwork.enlarge_graph`, with `reduce_graph` effectively being the inverse. I'm open to other verbs here; would either `shrink_graph` or `trim_graph` work?

"shrink" works, but I'm fine with keeping `reduce` as long as we're clear about the behavior!

dotsdl:
@atravitz after talking with @ianmkenney, we like `trim_graph` quite a bit. Any opposition to this?

atravitz:
> @atravitz after talking with @ianmkenney, we like `trim_graph` quite a bit. Any opposition to this?

that works for me!

dotsdl:
pre-commit.ci autofix

PR #335, opened 2024-08-19
Mention 1.0 Release

### LINKED ISSUE #334: Remove README note of this being "pre-alpha" work, since 1.0 release makes certain guarantees
See title. Since the 1.0 release is now in place, replace the "pre-alpha" note in the README with guarantees in place around stability, major changes in major releases, etc.

mikemhenry:
Fixes #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This library is made available under the [MIT](LICENSE) open source license.
 
 ## Important note
 
-This is pre-alpha work, it should not be considered ready for production and API changes are expected at any time without prior warning.
+Our 1.0 releases follow semantic versioning and we will make an effort to keep the 1.0 API stable.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ This library is made available under the [MIT](LICENSE) open source license.
 
 ## Important note
```
IAlibay, line 17:
```suggestion
## API stability
```
Happy to not go with this, but in my eyes it's no longer an "important note", so we probably should rename the title?

IAlibay (changes requested):
Thanks @mikemhenry - just the one thing for discussion

IAlibay (approved):
We need to fix CI but that's a different problem than this PR.

PR #344, opened 2024-09-11
Globally set RDKit to include all properties when pickling an RDKit Mol

### LINKED ISSUE #322: RDKit `Mol` doesn't preserve atom properties when pickled; impacts `ExplicitMoleculeComponent` and subclasses
Although we typically serialize `GufeTokenizable`s to `dict`, `keyed_dict`, or `KeyedChain` forms turned to JSON, it can also be convenient to pickle these objects in certain cases. When using `multiprocessing`, for example, `GufeTokenizable`s passed as arguments to functions executed on other processes are typically pickled/unpickled as the serialization approach.

This presents an issue for `ExplicitMoleculeComponent`s: RDKit `Mol` objects do not by default include atom properties when pickled (https://github.com/rdkit/rdkit/issues/6573#issuecomment-1781734093). This behavior causes issues especially for preserving partial charges, since we use RDKit `Mol` properties for holding on to these.

It's possible to change this behavior globally for RDKit with:
```python
from rdkit import Chem
Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
```

Should we set this within `gufe` so as to avoid this issue across the board, or will this have undesirable consequences?

dotsdl:
Closes #322.


IAlibay (commented):
I would like to see more tests on this please - it's not immediately clear to me that this doesn't change the behaviour and I would like to be sure it doesn't.

IAlibay (changes requested):
Meant this as a request changes.

IAlibay (approved):
@dotsdl I think we can go with this one, but it would be good to a) make sure we aren't creating big file (especially with ProteinComponents) - I could see this becoming an issue when you have large PDBs (maybe a fully solvated system?), b) it doesn't break any backwards compatibility.

I'm not 100% sure on the latter, it _shouldn't_ have an impact, but we'd need some kind of migration test (maybe something you already do in alchemiscale?). The former should be easy, we can take a big PDB file, serialize it, and see what breaks?

dotsdl:
@IAlibay thanks for this! This should have no impact on files, since we don't/can't support pickle files as a stable form of serialization written to disk. It is important, however, when using tools like `multiprocessing.ProcePool`, `concurrent.futures.ProcessPoolExecutor`, or `dask.distributed`, as these often rely on pickling for transmission of Python objects between processes.

dotsdl:
As an alternative, we could switch `ExplicitMoleculeComponent` to use `PropertyMol` instead of `Mol` as the class for its `_rdkit` attribute. That will always serialize its properties, and avoids us setting this behavior globally on `rdkit` itself for all `Mol` objects in the Python session.

gufe/components/explicitmoleculecomponent.py
```diff
@@ -9,6 +9,10 @@
 # typing
 from ..custom_typing import RDKitMol
 
+# globally set RDKit to include all properties when pickling an RDKit Mol
+# see issue #322: https://github.com/OpenFreeEnergy/gufe/issues/322
```
IAlibay, line 13: Can you add a pickle test please? If that's the use case we're targeting then we should have a test for it.

IAlibay (changes requested):
Probably needs a test.

gufe/components/explicitmoleculecomponent.py
dotsdl: Will do! Good call.

dotsdl:
I will add a test and a logger warning to the global settings change, and draft a follow-up issue proposing the use of `PropertyMol`s instead as a less-global, more targeted solution.

dotsdl:
@IAlibay this is ready for another review. Instead of setting global state, we decided to only issue a warning when pickling is attempted on an `ExplicitMoleculeComponent` and the `rdkit` pickle behavior is not set to preserve all properties.

IAlibay (approved):
Lgtm @dotsdl - however does it truly close #322? As in that issue remains active, would it be worth keeping it open?

dotsdl:
@IAlibay I'm in favor of letting this PR close #322, since it amounts to our answer to that issue at this time. If this problem rears its head again, we can consider another approach and reopen #322.

dotsdl:
pre-commit.ci autofix

PR #346, opened 2024-09-11
[WIP] Fe setup structure

RiesBen:
This is a draft for the future base class implemetnations for setting things up.
It allows the development of new protocols, such they fit seemingly into the concept.
This PR was discussed by the protocol devs.


dotsdl:
@RiesBen please let me know when you'd like a review from me! Happy to help you get this into `gufe`!

gufe/__init__.py
```diff
@@ -16,12 +16,6 @@
 
 from .chemicalsystem import ChemicalSystem
 
-from .mapping import (
```
IAlibay, line 19: We can't move all these imports, this will break the API

gufe/mapping/__init__.py
```diff
@@ -1,7 +0,0 @@
-# This code is part of gufe and is licensed under the MIT license.
-# For details, see https://github.com/OpenFreeEnergy/gufe
-"""Defining the relationship between different components"""
-from .componentmapping import ComponentMapping
```
IAlibay, line 4: Similarly, this can't go away until 2.0

gufe/setup/chemical_system_generator.py
```diff
@@ -0,0 +1,25 @@
...
+# For details, see https://github.com/OpenFreeEnergy/openfe
+import abc
+from enum import Enum
+
+from typing import Iterable
+from gufe import ChemicalSystem
+
+# Todo: connect to protocols - use this for labels?
+
+class RFEComponentLabels(str, Enum):
```
IAlibay, line 11: This needs more discusison, I don't remember this being in the plans for updating things, definitely using `labels` for components seems like something we don't want to do.

gufe/setup/network_planning/atom_mapping_based/atom_mapper.py
```diff
@@ -27,4 +27,5 @@ def suggest_mappings(self,
         Suggests zero or more :class:`.AtomMapping` objects, which are possible
         atom mappings between two :class:`.Component` objects.
         """
-        ...
+        raise NotImplementedError("This function was not implemented.")
```
IAlibay, line 30: I believe this (...) is a pattern we use quite widely for our abcs?

gufe/setup/network_planning/atom_mapping_based/atom_mapping.py
```diff
@@ -37,22 +37,23 @@ def componentA_to_componentB(self) -> Mapping[int, int]:
         entity in the other component (e.g. the atom disappears), therefore
         resulting in a KeyError on query
         """
-        ...
+        raise NotImplementedError("This function was not implemented.")
```
IAlibay, line 40: as above

gufe/setup/network_planning/atom_mapping_based/atom_mapping_scorer.py
```diff
@@ -0,0 +1,43 @@
+# This code is part of kartograf and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
+
+import abc
+from ....tokenization import GufeTokenizable
+
+from .atom_mapping import AtomMapping
+
+
+class AtomMappingScorer(GufeTokenizable):
```
IAlibay, line 10: This is in line with what we discussed, so this feature I think it ok to include.

gufe/setup/network_planning/component_mapper.py
```diff
@@ -0,0 +1,30 @@
...
+# For details, see https://github.com/OpenFreeEnergy/gufe
+import abc
+from collections.abc import Iterator
+import gufe
+
+from gufe.tokenization import GufeTokenizable
+from .component_mapping import ComponentMapping
+
+
+class ComponentMapper(GufeTokenizable):
```
IAlibay, line 11: This is also in line with what was discussed.

gufe/setup/network_planning/network_plan.py
```diff
@@ -0,0 +1,191 @@
...
+import networkx as nx
+from typing import FrozenSet, Iterable, Optional
+import gufe
+
+from gufe import Component
+from gufe.tokenization import GufeTokenizable
+
+from .component_mapping import ComponentMapping
+
+class NetworkPlan(GufeTokenizable):
```
IAlibay, line 16: Very briefly looking at class names, I don't think `Plan` is the right word here. It's not immediately clear what it does.

IAlibay (changes requested):

dotsdl:
@RiesBen can you offer a response to @IAlibay above? Can you give us a checklist of things you aim to do before you declare this PR ready for review?

gufe/__init__.py
RiesBen: shame :( I preferred that structure.

gufe/mapping/__init__.py
RiesBen: hm would it make sense to split this out into a new PR, or drop it completly? I m not sure of your intend here.

gufe/setup/chemical_system_generator.py
RiesBen: this is part of OpenFE Currently. however I wondered why, because it would make more sense here in my oppinion.
(I think at some time point we thought this is better then plain strings ;) )

gufe/setup/network_planning/atom_mapping_based/atom_mapper.py
RiesBen: I thought we tried to get rid of this pattern, as you can break the code with some vodoo (found that some time ago)

gufe/setup/network_planning/atom_mapping_based/atom_mapping.py
RiesBen: same

gufe/setup/network_planning/network_plan.py
RiesBen: do you have an alternative suggestion?

RiesBen: @dotsdl  maybe we can have a short look at this this evening in the gufe meetin?

RiesBen:
> @RiesBen can you offer a response to @IAlibay above? Can you give us a checklist of things you aim to do before you declare this PR ready for review?

Sure :) 

ToDos:

- [ ] discuss change suggestions in order to find a clear direction with some changes
  - [ ] Labels for `ChemicalSystems` transfer them from OpenFE to here or not?
  - [ ] `NetworkPlan`  name improvement
  - [ ] How to proceed with code/import structure.
- [ ] adapt the PR to the discussions.
- [ ] Check and adapt unit test

gufe/setup/chemical_system_generator.py
IAlibay: I don't think it should be migrated here, this seems like an implementation rather than an abstract class.

RiesBen: remove this again.

gufe/setup/network_planning/network_plan.py
RiesBen: This Class is required for this PR, I'm open for name suggestions, however a `NetworkPlan` is the result of a `NetworkPlanner` ;).

RiesBen:
Current Failure reasons are 2:
1. Gufe key stability: the key for `LigandNetwork` and `LigandAtomMapping` is not equal to the expected `GufeKey`. (affects: `test_key_stable`, `test_repr`
2. `test_graph_annotations`: The Annotations of the `LigandAtomMapping` is lost, after testing found this to have some weird behavior when you call `ligand_network.edges` and `ligand_network.graph..

gufe/mapping/__init__.py
RiesBen: I think now it should be not api breaking.

PR #348, opened 2024-09-17
Fix CI

IAlibay:
- Add in the shim for macos.

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -30,6 +30,7 @@ jobs:
           cache-downloads: true
           create-args: >-
             python=3.10
+            rdkit=2023.09.5
           init-shell: bash
 
       - name: "Install steps"
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -29,7 +29,7 @@ jobs:
           cache-environment: true
           cache-downloads: true
           create-args: >-
-            python=3.9
+            python=3.10
           init-shell: bash
 
       - name: "Install steps"
--- a/docs/environment.yaml
+++ b/docs/environment.yaml
@@ -9,4 +9,4 @@ dependencies:
 - sphinx
 - openmm
 - pip:
-  - git+https://github.com/OpenFreeEnergy/ofe-sphinx-theme@main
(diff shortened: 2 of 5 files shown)


docs/environment.yaml
```diff
@@ -9,4 +9,4 @@ dependencies:
 - sphinx
 - openmm
 - pip:
-  - git+https://github.com/OpenFreeEnergy/ofe-sphinx-theme@main
+  - git+https://github.com/OpenFreeEnergy/ofe-sphinx-theme@a45f3edd5bc3e973c1a01b577c71efa1b62a65d6
```
IAlibay, line 12: This is a known issue upstream

.github/workflows/mypy.yaml
```diff
@@ -29,7 +29,8 @@ jobs:
           cache-environment: true
           cache-downloads: true
           create-args: >-
-            python=3.9
+            python=3.10
+            rdkit=2023.09.5
```
IAlibay, line 33: I've never been able to work that one out.

dotsdl: @IAlibay can you elaborate? Work what out here?

docs/environment.yaml
dotsdl: Can you link to the issue? I'd like to learn what's going on here so we can follow up later.

dotsdl (approved):
@IAlibay some questions, but nothing blocking.

Do we know why some of the CI runs appear to be waiting endlessly?

.github/workflows/mypy.yaml
IAlibay: For some reason or another mypy has been complaining about a certain error with rdkit 2024+. Something about self being defined twice for a certain object.

Our approach has just been to pin to an older version of rdkit, but honestly we should try to fix it properly.

docs/environment.yaml
IAlibay: Here we go: https://github.com/OpenFreeEnergy/ofe-sphinx-theme/issues/5

It's a fixaeble packaging issue but I've not had time to look into it.

IAlibay:
> Do we know why some of the CI runs appear to be waiting endlessly?

It's because we've changed the matrix and some of the checks are enforced as "necessary" in the gufe repo settings, so github is waiting for runners that don't exist.

Issue #361, opened 2024-10-15
Add convenience methods to `GufeTokenizable`s for producing keyed chain representations


dotsdl:
After #286 is merged, it may make sense to add `to_keyed_chain` and `from_keyed_chain` methods to `GufeTokenizable` to make keyed chains a convenient representation in the same way as keyed dicts and shallow dicts.

We should also add corresponding developer docs giving guidance on when this representation is most advantageous.

PR #367, opened 2024-10-24
Add to/from keyed chain methods to `GufeTokenizable`

jthorton:
Implements #361 by adding `to_keyed_chain` and a `from_keyed_chain` classmethod to GufeTokenizable. This should enable more compact serialisation options of GufeTokenizable objects throughout the ecosystem.


dotsdl:
Thanks for this @jthorton! It looks like #368 also features these changes, so I recommend we proceed with that one and close this PR.

Issue #359, opened 2024-10-11
Convience methods to write to/from json


jthorton:
While writing tests for the industry benchmark I find myself writing the same two bits of code a lot to write and load `AlchemicalNetworks` to json file. I think it would make sense to add them as functions to avoid repeated code. I would suggest a `to_json_file` method and a `from_json_file` class method. In the future we might consider other supported file types like messagepack so a general `to_file` and `from_file` might be more appropriate with the file type inferred from the name.

I noticed that the `Transformation` object can already be written to json using its `dump` method so I think having some general serialisation options on the `GufeTokenizable` class would be very useful!

dotsdl:
Agree with this @jthorton! Although this gets a bit tricky, since there are multiple representations a `GufeTokenizable` can take, including `dict`, `shallow_dict`, `keyed_dict`, and now `keyed_chain`. All of these are readily JSON-serializable except `shallow_dict`, but which one should be used for e.g. `to_json`?

What may help is to create a `gufe.serialization` submodule with convenience functions, such as `to_json`, `from_json`, `to_msgpack`, `from_msgpack`, etc. These functions could by default return an in-memory representation, but could optionally take a filename or file-like object to write to.

Do you think that would be sufficient, or still be annoying as a downstream user of `GufeTokenizable`s?

jthorton:
> All of these are readily JSON-serializable except shallow_dict, but which one should be used for e.g. to_json?

I would say the most minimal one which can be used to roundtrip the same object currently I just use `dict` as I wasn't sure but if the `to/from` methods could handle this for me that would be fantastic!

>What may help is to create a gufe.serialization submodule with convenience functions, such as to_json, from_json, to_msgpack, from_msgpack, etc. These functions could by default return an in-memory representation, but could optionally take a filename or file-like object to write to.

Great idea, I think that would help a lot and save [downstream implementations](https://github.com/asapdiscovery/asapdiscovery/blob/7234302b61cb367e27b1ac458d7d637cac48c2b0/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/base.py#L15-L32).

PR #368, opened 2024-10-24
JSON serialization on `GufeTokenizable`

jthorton:
Implements #359.

This adds convenience methods to the `GufeTokenizable` class to aid easy serialisation to and from JSON using the `keyed_chain` pattern which keeps the JSON objects small. 

# Questions
@dotsdl I like the idea of having a serialisation module but noticed I would have to move the `JSON_HANDLER` which is a large API break so I went for the easier option for now. 

For the `from_json` method I wasn't quite sure how to handle filelike vs  JSON object input, Do you have any ideas? 

I noticed the [Transformation](https://docs.openfree.energy/en/stable/cookbook/dumping_transformation.html) object has saving and loading from JSON already with its `dump/load` methods do we want to keep these and route them through the new `to/from_json` methods and remove them later?

If we are happy with this I'll add some tests.

### DIFF
--- a/gufe/tests/test_tokenization.py
+++ b/gufe/tests/test_tokenization.py
@@ -257,18 +257,21 @@ class TestGufeTokenizable(GufeTokenizableTestsMixin):
 
     def test_to_json_string(self):
         raw_json = self.cont.to_json()
+
         # tuples are converted to lists in JSON so fix the expected result to use lists
         expected_key_chain = [list(tok) for tok in self.expected_keyed_chain]
         assert json.loads(raw_json, cls=JSON_HANDLER.decoder) == expected_key_chain
 
     def test_from_json_string(self):
         recreated = self.cls.from_json(content=json.dumps(self.expected_keyed_chain, cls=JSON_HANDLER.encoder))
+
         assert recreated == self.cont
         assert recreated is self.cont
 
     def test_to_json_file(self, tmpdir):
         file_path = tmpdir / "container.json"
         self.cont.to_json(file=file_path)
+
         # tuples are converted to lists in JSON so fix the expected result to use lists
         expected_key_chain = [list(tok) for tok in self.expected_keyed_chain]
         assert json.load(file_path.open(mode="r"), cls=JSON_HANDLER.decoder) == expected_key_chain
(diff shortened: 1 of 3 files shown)


dotsdl (changes requested):
Fantastic work @jthorton! I've made some adjustments, and also added the ability to use `from_json` on either string content or filepath/filelike objects.

Can you add some tests for `to_json` and `from_json`? I like what you've added for `to_keyed_chain` and `from_keyed_chain`, so these can just expand on those.

jthorton:
Thanks for the review @dotsdl the tests are included in [068d583](https://github.com/OpenFreeEnergy/gufe/pull/368/commits/068d58380a195ceb529ee0f7a079f54c51de8090). Should we also update the [documentation](https://gufe.readthedocs.io/en/latest/guide/serialization.html#dumping-arbitrary-objects-to-json) to advertise the new serialization method?

dotsdl:
> Thanks for the review @dotsdl the tests are included in [068d583](https://github.com/OpenFreeEnergy/gufe/pull/368/commits/068d58380a195ceb529ee0f7a079f54c51de8090). Should we also update the [documentation](https://gufe.readthedocs.io/en/latest/guide/serialization.html#dumping-arbitrary-objects-to-json) to advertise the new serialization method?

Yes! Good catch! Would you like to give it a go? Otherwise I'm happy to update the docs as an additional commit to this PR.

jthorton:
@david I had a go at the docs and the changes can be seen [here](https://gufe--368.org.readthedocs.build/en/368/guide/serialization.html#built-in-serialisation) please feel free to edit as you see fit.

gufe/tests/test_tokenization.py
```diff
@@ -232,6 +247,39 @@ def test_from_keyed_dict(self):
...
+
+    def test_from_keyed_chain(self):
+        recreated = self.cls.from_keyed_chain(self.expected_keyed_chain)
+        assert recreated == self.cont
+        assert recreated is self.cont
+
+    def test_to_json_string(self):
+        raw_json = self.cont.to_json()
+        # tuples are converted to lists in JSON so fix the expected result to use lists
+        expected_key_chain = [list(tok) for tok in self.expected_keyed_chain]
```
dotsdl, line 262: Ah, I guess this is a slight ugly spot in our keyed chain representation, but philosophically I do prefer the tuple usage since each element of a keyed chain has exactly two elements of its own.

dotsdl (approved):
Fantastic work @jthorton! Feel free to merge if satisfied!

gufe/tests/test_tokenization.py
jthorton: I agree the tuple is nicer I'll just leave this here so others know whats going on.

PR #369, opened 2024-10-24
Save edge annotations in `LigandNetwork.to_graphml`

### LINKED ISSUE #365: `LigandNetwork.to_graphml` does not save edge annotations
Round tripping a `LigandNetwork` to graphml does not save the edge annotations like the mapping score which users might want to review later.  See the example:

```python
from rdkit import Chem
from openfe import SmallMoleculeComponent
import openfe
from openfe.setup.ligand_network_planning import generate_minimal_spanning_network
from openfe.setup import LomapAtomMapper
from gufe import LigandNetwork



# Load ligands using RDKit
ligands_sdf = Chem.SDMolSupplier('tyk2_ligands.sdf', removeHs=False)

# Now pass these to form a list of Molecules
ligand_mols = [SmallMoleculeComponent(sdf) for sdf in ligands_sdf]

# plan a quick network
mst_network = generate_minimal_spanning_network(
    ligands=ligand_mols,
    scorer=openfe.lomap_scorers.default_lomap_score,
    mappers=[LomapAtomMapper(),])

# print the edge scores
for edge in mst_network.edges:
    print(edge.annotations["score"])

# now round trip the ligand network
graphml = mst_network.to_graphml()
del mst_network
network2 = LigandNetwork.from_graphml(graphml_str=graphml)
# check the scores again
for edge in network2.edges:
    print(edge.annotations["score"])

```
> 0.9048374180359595
0.951229424500714
0.33287108369807955
0.301194211912202
0.9048374180359595
0.951229424500714
0.9048374180359595
0.36787944117144233
0.951229424500714
Traceback (most recent call last):
  File "/Users/joshua/Documents/Software/openfe_dev/graphml_issue/test_graphml.py", line 32, in <module>
    print(edge.annotations["score"])
          ~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'score'

jthorton:
Fixes #365.

Tests will currently fail as the test file has no edge annotations, I just want to check we are happy with the method before I edit the test file.

### DIFF
--- a/gufe/tests/test_ligand_network.py
+++ b/gufe/tests/test_ligand_network.py
@@ -47,9 +47,9 @@ def mols():
 @pytest.fixture
 def std_edges(mols):
     mol1, mol2, mol3 = mols
-    edge12 = LigandAtomMapping(mol1, mol2, {0: 0, 1: 1})
-    edge23 = LigandAtomMapping(mol2, mol3, {0: 0})
-    edge13 = LigandAtomMapping(mol1, mol3, {0: 0, 2: 1})
+    edge12 = LigandAtomMapping(mol1, mol2, {0: 0, 1: 1}, {"score": 0.0})
+    edge23 = LigandAtomMapping(mol2, mol3, {0: 0}, {"score": 1.0})
+    edge13 = LigandAtomMapping(mol1, mol3, {0: 0, 2: 1}, {"score": 0.5})
     return edge12, edge23, edge13
 
 
--- a/gufe/ligandnetwork.py
+++ b/gufe/ligandnetwork.py
@@ -97,7 +97,8 @@ class LigandNetwork(GufeTokenizable):
             (
                 mol_to_label[edge.componentA],
                 mol_to_label[edge.componentB],
-                json.dumps(list(edge.componentA_to_componentB.items()))
+                json.dumps(list(edge.componentA_to_componentB.items())),
+                json.dumps(edge.annotations),
             )
             for edge in self.edges
         ])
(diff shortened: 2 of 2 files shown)


gufe/tests/test_ligand_network.py
```diff
@@ -47,9 +47,9 @@ def mols():
 @pytest.fixture
 def std_edges(mols):
     mol1, mol2, mol3 = mols
-    edge12 = LigandAtomMapping(mol1, mol2, {0: 0, 1: 1})
-    edge23 = LigandAtomMapping(mol2, mol3, {0: 0})
-    edge13 = LigandAtomMapping(mol1, mol3, {0: 0, 2: 1})
+    edge12 = LigandAtomMapping(mol1, mol2, {0: 0, 1: 1}, {"score": 0.0})
```
IAlibay, line 50: Could you test other types of annotations too? For example, what if you passed a united quantity - from the code, it looks like that might fail ungracefully.

IAlibay (changes requested):
A bit more testing might be worth it, that way we know the bounds of what's possible.

gufe/tests/test_ligand_network.py
jthorton: Good point we could add support for all of our JSON codec entries here, it might be worth considering transitioning to the new `to_json` and `from_json` methods to serialise these networks generally.

IAlibay: For now unfortunately we do have to fix the graphml serialisation because it's what we've been getting out users to use, so for the rest of 1.x we have an obligation to keep this going. 

Probably just trying to see if we can use the standard JSON decoder/encoder on the annotations might be worth it.

IAlibay (approved):

PR #374, opened 2024-10-29
Removed all hardcoded gufe keys from GufeTokenizable tests in test suite

dotsdl:
We do not consider `gufe` keys to be stable across versions, though they should be stable in-process and generally across machines running the same software versions. We only rely on them being stable in-process, however.

### DIFF
--- a/gufe/tests/test_tokenization.py
+++ b/gufe/tests/test_tokenization.py
@@ -82,7 +82,6 @@ class GufeTokenizableTestsMixin(abc.ABC):
 
     # set this to the `GufeTokenizable` subclass you are testing
     cls: type[GufeTokenizable]
-    key: Optional[str]
     repr: Optional[str]
 
     @pytest.fixture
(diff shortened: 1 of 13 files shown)


gufe/tests/test_tokenization.py
```diff
@@ -140,9 +139,6 @@ def test_to_shallow_dict_roundtrip(self, instance):
         # include `np.nan`s
         #assert ser == reser
 
-    def test_key_stable(self, instance):
```
IAlibay, line 143: We should do a key stability test - i.e. if you generate the key twice do you get the same value. 

Shall we change this test to that?

IAlibay (changes requested):
As you mention, keys should be stable in the same python call so we should probably have a small test that checks that.

gufe/tests/test_tokenization.py
dotsdl: Yes! This is a good idea; I'll implement this next.

dotsdl: @IAlibay please have a look at 1dca49d and let me know if this satisfies!

IAlibay (approved):
lgtm!

dotsdl:
Gracias @IAlibay! Merging!

PR #381, opened 2024-10-30
Add the `n_protocol_dag_results` property to `ProtocolResult`

ianmkenney:
This PR adds the `n_protocol_dag_results` property to the `ProtocolResult` class. It reflects the number of `ProtocolDAGResult`s that were input to the `_gather` call.


ianmkenney:
I'm probably being a little (very) pedantic here, but since `gather` takes in an Iterable, which can include non-terminating generators, it's possible for gather to hang forever (before running out of memory and exiting).

Since I'm using the `len` function in this PR, even well-behaved generators will no longer work. An alternative would be to use `list(protocol_dag_results)`, but this is still susceptible to the infinite generator case. I'm not sure what the best option for typing is, since there doesn't appear to be a good builtin that enforces a terminating iterable and I think defining and comparing to a custom type might be more extreme than just not allowing generators.

All of that said, I'm happy to revert the guardrail (and its test) if it seems like too much for this PR. Although I suspect type checkers will complain if I use `len` rather than `list`.

dotsdl:
I think the case of infinite `ProtocolDAGResult`s isn't remotely likely, but I see your point. We could slate changing the type hint for `gather` from `Iterable` to `list` for `gufe` 2.0, indicating that we explicitly expect a `list` (and therefore a finite number) of `ProtocolDAGResult`s fed to `gather`.

dotsdl:
You will also need to make changes to the `_to_dict` and `_from_dict` methods in `ProtocolResult` reflecting these changes.

gufe/protocols/protocol.py
```diff
@@ -32,8 +32,11 @@ class ProtocolResult(GufeTokenizable):
     - `get_uncertainty`
     """
 
-    def __init__(self, **data):
+    def __init__(self, n_protocol_dag_results: Optional[int] = None, **data):
         self._data = data
+        self._n_protocol_dag_results = (
+            n_protocol_dag_results if n_protocol_dag_results is not None else 0
```
IAlibay, line 38: I'm trying to wrap my head around the use case here, could you explain the case where None happens? I.e. at first glance setting None to zero seems incorrect if None means that the gather method did not pass through a value. In that case None means undefined and it's much cleaner than getting 0 with the idea that you got no results (but you do).

IAlibay (changes requested):
Also blocking on downstream testing and maybe a bit of a discussion on protocol needs.

gufe/protocols/protocol.py
ianmkenney: I think in the context of a `Protocol`'s `gather`, you'll never see a `None` passed in. I can see instances where `ProtocolResult`s are used directly, incorrectly or not, and a `None` would make sense. But of course I immediately throw away the `None` in the code you highlighted above. I suppose there is still some indecision on my part what the right signature should be and which values the `ProtocolResult` holds on to. I'm tempted to just drop `Optional[int]` and use `int` with default `0`.

Just from the tests, I see instances where someone would make a `ProtocolResult` directly and I'll probably do something similar in testing parts of `stratocaster` downstream:

```python
class DummyProtocolResult(gufe.ProtocolResult):
    def get_estimate(self):
        return self.data['estimate']

    def get_uncertainty(self):
        return self.data['uncertainty']


class TestProtocolResult(GufeTokenizableTestsMixin):
    cls = DummyProtocolResult
    key = 'DummyProtocolResult-b7b854b39c1e37feabec58b4000680a0'
    repr = f'<{key}>'

    @pytest.fixture
    def instance(self):
        return DummyProtocolResult(
            estimate=4.2 * unit.kilojoule_per_mole,
            uncertainty=0.2 * unit.kilojoule_per_mole,
        )
```

ianmkenney:
> I think the case of infinite `ProtocolDAGResult`s isn't remotely likely, but I see your point. We could slate changing the type hint for `gather` from `Iterable` to `list` for `gufe` 2.0, indicating that we explicitly expect a `list` (and therefore a finite number) of `ProtocolDAGResult`s fed to `gather`.
> 
> Do you agree with this?

Yeah that works, I think in the long term it will simplify typing issues.

gufe/protocols/protocol.py
ianmkenney: Not necessarily. I'll be testing stratocaster strategies based on this branch until it's merged in the hopes that we get some solution that's close in spirit to what we have here.

dotsdl:
@IAlibay downstream do you serialize `ProtocolResult` objects, or `ProtocolDAGResult` objects. The former will be impacted by this change, but the latter will not.

gufe/protocols/protocol.py
```diff
@@ -32,19 +31,28 @@ class ProtocolResult(GufeTokenizable):
...
         return {}
 
     def _to_dict(self):
-        return {'data': self.data}
+        return {'n_protocol_dag_results': self.n_protocol_dag_results, 'data': self.data}
 
     @classmethod
     def _from_dict(cls, dct: dict):
-        return cls(**dct['data'])
+        return cls(n_protocol_dag_results=dct['n_protocol_dag_results'], **dct['data'])
```
IAlibay, line 51:
```suggestion
        # TODO: remove in gufe 2.0!
        try:
            n_protocol_dag_results = dct['n_protocol_dag_results']
        except KeyError:
            n_protocol_dag_results = None
        return cls(n_protocol_dag_results=n_protocol_dag_results, **dct['data'])
```

IAlibay (commented):
How about this?

gufe/protocols/protocol.py
ianmkenney: Added!

IAlibay (approved):
lgtm thanks!

dotsdl:
@ianmkenney if you can resolve conflicts, I think this one is good to go for merge!

Issue #360, opened 2024-10-11
The `nodes` optional attributes to LigandNetworks is unclear / unexplained


IAlibay:
From a developer or user perspective, it's unclear what is the intent of the `nodes` kwarg.

My personal understanding is that it's to add nodes that aren't connected (i.e. if you had absolute calculations or similar). Making the intent clearer would be really helpful.

PR #390, opened 2024-11-07
adding to LigandNetwork docstring

atravitz:
addresses https://github.com/OpenFreeEnergy/gufe/issues/360. 

hopefully this is more clear (and also correct).

### DIFF
--- a/gufe/ligandnetwork.py
+++ b/gufe/ligandnetwork.py
@@ -14,14 +14,17 @@ from .tokenization import GufeTokenizable, JSON_HANDLER
 
 
 class LigandNetwork(GufeTokenizable):
-    """A directed graph connecting many ligands according to their atom mapping
+    """A directed graph connecting ligands according to their atom mapping.
+       A network can be defined by specifying only edges, in which case the nodes are implicitly added.
+      
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


gufe/ligandnetwork.py
```diff
@@ -14,14 +14,17 @@
...
 
     Parameters
     ----------
     edges : Iterable[LigandAtomMapping]
-        edges for this network
+        Edges for this network, each specified as a LigandAtomMapping between two nodes.
     nodes : Iterable[SmallMoleculeComponent]
-        nodes for this network
+        Nodes for this network. Any nodes already included as a part of the 'edges' will be ignored.
+        Nodes not already included in 'edges' will be added as isolated, unconnected nodes.
```
IAlibay, line 27: It comes to mind that we don't have a mechanism to handle the reverse case - i.e. what if we just wanted unconnected nodes for some reason or another? This is a conversation for another time.

IAlibay (approved):
Looks good to me thanks!

gufe/ligandnetwork.py
atravitz: opened issue here: https://github.com/OpenFreeEnergy/gufe/pull/390#discussion_r1833792370

PR #406, opened 2024-11-13
vendor file updates + test openmm 8.2 & 8.1.2

### LINKED ISSUE #410: PDBFile vendor no longer matches behaviour of OpenMM 8.2
OpenMM 8.2 now includes formal charges when writing PDBs with PDBFile. This differs from the current PDBFile vendor.

We should update our vendored files to match OpenMM behaviour.

Ideally we should also port all the tests from OpenMM upstream since those don't seem to be present.

mikemhenry:
- [x] Update vendored files
- [x] Test openmm 8.2 & 8.1.2
- [ ] fix tests

Fixes #410


mikemhenry:
Okay now we are getting the pdb file issues, this is different than the hybrid topo stuff

mikemhenry:
This is what the coverage looks like:
image
I will go back to skipping the coverage and just work on updating the files

mikemhenry:
Okay this is actually a bit of a mess. Now that I've updated our vendored files, we have a problem when using openmm 8.1.2, since we use openmm's objects here:
https://github.com/OpenFreeEnergy/gufe/blob/cd9b0f9cb53bd3c9fbde6dfe05019a25d1f0aa5a/gufe/components/proteincomponent.py#L416
but then our vendored objects here:
https://github.com/OpenFreeEnergy/gufe/blob/cd9b0f9cb53bd3c9fbde6dfe05019a25d1f0aa5a/gufe/components/proteincomponent.py#L500

So we get a `AttributeError: 'Atom' object has no attribute 'formalCharge'` error since we are using 8.1.2 to make the atom objects, but then our object (based on the openmm 8.2.0 code) to write out those objects, and since 8.2.0 is expecting the `Atom` attribute, we get an error.

A fix could be to use our vendored topology object instead (so we would use the vendored version of Topology instead of calling `self.to_openmm_topology()`)

mikemhenry:
I tried to use the vendored version of `Element` but b/c we check if it matches the openmm objects, we get this error:
```
    def assert_topology_equal(ref_top, top):
        assert ref_top.getNumAtoms() == top.getNumAtoms()
        for ref_atom, atom in zip(ref_top.atoms(), top.atoms()):
            assert ref_atom.name == atom.name
>           assert ref_atom.element == atom.element
E           assert <Element carbon> == <Element carbon>
E            +  where <Element carbon> = <Atom 0 (CH3) of chain 0 residue 0 (ACE)>.element
E            +  and   <Element carbon> = <Atom 0 (CH3) of chain 0 residue 0 (ACE)>.element
```

I don't want to modify the tests to fix this, so we can use openmm's version of the Element object. 

I did manage to use our vendored version of the bond objects.

gufe/vendor/pdb_file/pdbfile.py
```diff
@@ -40,6 +40,7 @@
 from datetime import date
 
 import numpy as np
+from openmm import Platform
```
IAlibay, line 43: It bafflles me as to why we vendor from OpenMM and then also import from it in the vendor 🤦🏽 (this is a complaint against the concept of this file, not this PR).

Again, something we're better off dealing with later.

.github/workflows/ci.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+        openmm-version:
+          - "8.2.0"
+          - "8.1.2"
```
IAlibay, line 38: Can we skip PDB checking tests for pre-8.2? That way CI won't fail.

IAlibay (approved):
This looks good to me, I do agree with the historical issues you've already raised @mikemhenry.. this vendor is quite messy.

My only requested change is in in terms of testing. As per yesterday's call, I'm of the opinion that it doesn't make sense to try to have tests pass against multiple openmm versions. Since it's mostly a PDB formatting things, matching 8.2 is "more correct" than matching pre-8.2. So my suggestion would be to only test PDB writing when testing with 8.2+.

IAlibay (changes requested):
Oops technically that was a "requested changes".

IAlibay:
It would be good to try to get this into a release of gufe so we can more easily do an openmm 8.2 rc release of OpenFE.

How feasible would it be to finish this PR in the next few days @mikemhenry ?

gufe/tests/conftest.py
```diff
@@ -67,7 +67,12 @@ def get_test_filename(filename):
     name: URLFileLike(url=_pl_benchmark_url_pattern.format(name=name)) for name in _benchmark_pdb_names
 }
 
-PDB_FILE_LOADERS = {name: lambda: get_test_filename(name) for name in ["181l.pdb"]}
+PDB_FILE_LOADERS = {
+    name: (lambda name=name: get_test_filename(name))
```
mikemhenry, line 71: this was fun https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

mikemhenry:
I will need to make another PR that vendors in openff-models since that will prevent us from being numpy 2 compatible

mikemhenry:
@IAlibay @dotsdl @atravitz We decided to just match openmm for now, and maybe raise an issue about it in openmm

gufe/tests/conftest.py
```diff
@@ -67,7 +67,12 @@ def get_test_filename(filename):
     name: URLFileLike(url=_pl_benchmark_url_pattern.format(name=name)) for name in _benchmark_pdb_names
```
IAlibay, line 67: Need to hard-code release version here (or better, just vendor all the PDBs into gufe)

atravitz:
addressed by https://github.com/OpenFreeEnergy/gufe/pull/539, closing.

PR #408, opened 2024-11-14
Save atom hybridization

### LINKED ISSUE #407: SmallMoleculeComponent round trip does not save atom hybridization
I noticed the hybridization of each  atom was set as `UNSPECIFIED` when working on a ligand from a graphml object I think this is due to the `to_dict` method not saving the hybridization of the atom. See this simplified example 

```python
from rdkit import Chem
from gufe import SmallMoleculeComponent

# generate an rdkit molecule
ethane = Chem.MolFromSmiles("CC")
ethane = Chem.AddHs(methane)
Chem.rdDistGeom.EmbedMolecule(methane, useRandomCoords=False, randomSeed=0)
# check the state of each carbon
for atom in methane.GetAtoms():
    if atom.GetAtomicNum() == 6:
        print(atom.GetHybridization())

# create the SMC
smc = SmallMoleculeComponent(methane, name="methane")
# round trip to dict
smc_dict = smc.to_dict()
# remove the original object
del smc
smc_2 = SmallMoleculeComponent.from_dict(smc_dict)
# check the state again
for atom in smc_2.to_rdkit().GetAtoms():
    if atom.GetAtomicNum() == 6:
        print(atom.GetHybridization())
```
> SP3
SP3
UNSPECIFIED
UNSPECIFIED

jthorton:
Fixes #407 by saving the atom hybridization in the` to_dict` method and uses it in the `from_dict` method. This should now cause an error when loading molecules which were serialised using the old `to_dict` method though so we might need to think about how to support backwards compatibility with old serialised objects.


IAlibay (approved):
lgtm thanks!

gufe/components/smallmoleculecomponent.py
```diff
@@ -247,6 +261,7 @@ def _from_dict(cls, d: dict):
             a.SetChiralTag(_INT_TO_ATOMCHIRAL[atom[4]])
             a.SetAtomMapNum(atom[5])
             _setprops(a, atom[6])
+            a.SetHybridization(_INT_TO_HYBRIDIZATION[atom[7]])
```
IAlibay, line 264: We should put this in a try/except with a warning.

IAlibay (changes requested):
Actually we'll need to do a migration I think.

I.e. what if you deserialize and there's only 7 entries in `atom`?

gufe/components/smallmoleculecomponent.py
jthorton: Yeah this is why the tests are currently failing as the old graphml files don't have the hybridization info, I think try-except is the best option so I'll add that in.

IAlibay: Thanks!

jthorton: done in [27bd913](https://github.com/OpenFreeEnergy/gufe/pull/408/commits/27bd913c3eaa01a4ee5183ffb9265d7a22978a24)

jthorton:
pre-commit.ci autofix

jthorton:
Oh dear looks like we need someone to run the lint in a different PR first!

jthorton:
Yeah it looks like it was done in https://github.com/OpenFreeEnergy/gufe/pull/421 but something might have gone wrong it looks like pre-commit is failing on main and it was to change a lot of files here, tagging @mikemhenry.

gufe/components/smallmoleculecomponent.py
```diff
@@ -223,6 +237,7 @@ def _to_dict(self) -> dict:
                     _ATOMCHIRAL_TO_INT[atom.GetChiralTag()],
                     atom.GetAtomMapNum(),
                     atom.GetPropsAsDict(includePrivate=False),
+                    _HYBRIDIZATION_TO_INT[atom.GetHybridization()],
```
IAlibay, line 240: What happens in this case if this isn't set?

I.e. if you read and old file (which now skips the hybridization reading), then write it out again - what does this pick up?

IAlibay (commented):
A question on what happens when you don't have the property set and you try to write.

gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -339,6 +339,35 @@ def test_to_dict(self, phenol):
...
+            # remove the hybridization atomic info which should be at index 7
+            new_atoms.append(tuple([atom_info for i, atom_info in enumerate(atom) if i != 7]))
+        phenol_dict["atoms"] = new_atoms
+        new_phenol = SmallMoleculeComponent.from_dict(phenol_dict)
+        # they should be different objects due to the missing hybridization info
+        assert new_phenol != phenol
+        # make sure the rdkit objects are different
+        for atom_hybrid, atom_no_hybrid in zip(phenol.to_rdkit().GetAtoms(), new_phenol.to_rdkit().GetAtoms()):
+            assert atom_hybrid.GetHybridization() != atom_no_hybrid.GetHybridization()
+
```
jthorton, line 370: @IAlibay this test covers loading an old SMC with the missing info it looks like Rdkit defaults to setting the hybridization as undefined.

IAlibay: Ok thanks - in that case we should let users know they've hit a bug (just so that they don't end up silently dealing with this). Could you add a warning please?

jthorton:
pre-commit.ci autofix

gufe/tests/test_smallmoleculecomponent.py
jthorton: Added the warning though this might become a bit much as every SMC loaded from file created with gufe<1.1.0 will trigger this.

IAlibay (approved):
lgtm thanks!

Issue #383, opened 2024-11-04
Get connected subgraphs in an  `AlchemicalNetwork` 


jthorton:
[From Irfan](https://github.com/OpenFreeEnergy/IndustryBenchmarks2024/pull/156#discussion_r1827612898) 
> This is a perfectly fine strategy. In the future we should see if we can extend the `.graph` method of AlchemicalNetwork objects to offer this kind of convenience. I.e. "get me all the connected subgraphs".

dotsdl:
This subgraphs method on an `AlchemicalNetwork` could return the subgraphs as `AlchemicalNetwork`s themselves.

PR #409, opened 2024-11-14
383 get connected subgraphs in an alchemicalnetwork

atravitz:
addresses #383

### DIFF
--- a/gufe/network.py
+++ b/gufe/network.py
@@ -1,7 +1,7 @@
 # This code is part of OpenFE and is licensed under the MIT license.
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
-from typing import Iterable, Optional
+from typing import Generator, Iterable, Optional, Self
 
 import networkx as nx
 from .tokenization import GufeTokenizable
@@ -102,7 +102,7 @@ class AlchemicalNetwork(GufeTokenizable):
                 "name": self.name}
 
     @classmethod
-    def _from_dict(cls, d: dict):
+    def _from_dict(cls, d: dict) -> Self:
         return cls(nodes=frozenset(d['nodes']),
                    edges=frozenset(d['edges']),
                    name=d.get('name'))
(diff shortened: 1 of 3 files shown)


atravitz:
I'm not convinced that `connected_subgraphs` is the clearest language here. Most technically accurate would be `connected_components`, but given that `components` has a different meaning in openfe-world than in network theory world, I don't like that either.

Maybe `subgraphs` really is the best option?

gufe/network.py
```diff
@@ -10,6 +11,7 @@
 from .transformations import Transformation
 
 
+
```
IAlibay, line 14:
```suggestion
```
We really should just apply `black` to gufe.

gufe/network.py
```diff
@@ -119,3 +121,18 @@ def to_graphml(self) -> str:
     def from_graphml(cls, str):
```
IAlibay, line 121:
```suggestion
    def from_graphml(cls, str) -> Self:
```

gufe/tests/test_alchemicalnetwork.py
```diff
@@ -47,3 +47,22 @@ def test_connectivity(self, benzene_variants_star_map):
             else:
                 edges = alnet.graph.edges(node)
                 assert len(edges) == 0
+
+    def test_connected_subgraphs(self, benzene_variants_star_map):
```
IAlibay, line 51: could you add a brief test that checks that this works for fully connected graph? i.e. it just returns itself

IAlibay (changes requested):
Couple of small things, otherwise lgtm.

gufe/tests/test_alchemicalnetwork.py
atravitz: yes good idea. I rearranged the fixtures a bit so we can access either one or both of the ligand and complex star networks for the benzene test case. I think this is the clearest way to do this without duplicating too much code.

PR #418, opened 2024-11-21
Add a method for comparing components in ChemicalSystem objects

### LINKED ISSUE #347: ChemicalSystem difference method
## User story

It would be good to have a method built into ChemicalSystem that checks how it differs from another ChemicalSystem and have it return a dict of list of components that are different.

ianmkenney:
Add `component_diff` method to compare ChemicalSystem components.

Implement a method to compare the components of a ChemicalSystem with those of another.

closes #347


ianmkenney:
@IAlibay Is this what you had in mind?

gufe/chemicalsystem.py
```diff
@@ -72,6 +72,44 @@ def components(self) -> dict[str, Component]:
...
+            Only components that differ between the two ChemicalSystem objects will appear in the dictionary.
+
+        Raises
+        ------
+        TypeError
+            If `other` is not an instance of the same class as `self`.
+        """
+
+        # TODO: is this too restrictive?
+        if self.__class__ is not other.__class__:
```
dotsdl, line 97: Can you change this to 
```suggestion
        if not isinstance(other, ChemicalSystem):
```

This would allow for subclasseses of `ChemicalSystem` to use this method.

dotsdl (approved):
Looks great @ianmkenney! Thank you for this!

gufe/chemicalsystem.py
```diff
@@ -72,6 +72,44 @@ def components(self) -> dict[str, Component]:
...
+        ----------
+        other : ChemicalSystem
+            The ChemicalSystem to compare to.
+
+        Returns
+        -------
+        dict[str, tuple[Component | None, Component | None]]
+            A dictionary whose keys are the component labels present in either of the two ChemicalSystem objects.
+            The values are tuples containing the different Component objects from this and the other ChemicalSystem, respectively.
+            Only components that differ between the two ChemicalSystem objects will appear in the dictionary.
```
IAlibay, line 88: It's probably just me.. but I have trouble understanding the intended behaviour from this docstring 🙈 

**IF* it's not a lot of work, could you add in a usage code example a bit lower in the docstring?

gufe/chemicalsystem.py
```diff
@@ -72,6 +72,44 @@ def components(self) -> dict[str, Component]:
...
+
+        expanded_keys = self._components.keys() | other._components.keys()
+
+        difference : dict[str, tuple[Component | None, Component | None]] = {}
+
+        for key in expanded_keys:
+            value_self = self._components.get(key)
+            value_other = other._components.get(key)
+
+            if value_self != value_other:
```
IAlibay, line 108: If I'm corectly interpreting the way we're doing equality, we are saying here - components are equivalent if:

1. They have the same ChemicalSystem key
2. They have gufe equality

The caveat here is that I think we've always somewhat thought that key names (in the ChemicalSystem) were purely arbitrary and more of a user convenience than something we deal with when it comes to equality.

Probably a terrible example of a use case where this falls apart a bit:

1. For some reason I'm reading two SDF files to define the SMCs of my end states
2. The SDFs are not ordered the same, but I don't really care (after all the keys are arbitrary)
3. If I use this in Protocol, I then end up with everything being different, when actually it's just one SMC.

Please don't take my take as "this is the way", but rather "we should discuss this because it's an assumption we've not been making".

IAlibay (changes requested):
Sorry very long winded comment about equality and the arbitrary nature of keys in ChemicalSystems.

The answer probably should have been me coming to the gufe meeting today 🙈 sorry!

gufe/chemicalsystem.py
dotsdl: I like this a lot @ianmkenney after discussion in person. It's a nice alternative to doing `set`-style operations (e.g. `set.difference` or `set.symmetric_difference`) where you only get the difference in a single direction, requiring two calls if you want it in both.

IAlibay: Apologies for not getting back here @ianmkenney ! Yes this proposed solution sounds great, it directly unpacks into "what are my alchemical molecules", I love it!

IAlibay (approved):
lgtm, thanks @ianmkenney !

PR #421, opened 2024-11-22
auto format all the things

mikemhenry:


mikemhenry:
I wanted to make sure CI passed on the auto format, since at one point it was removing trailing white space that was syntactically needed for some file types

Issue #313, opened 2024-04-26
Unpin macos runner


IAlibay:
See: https://github.com/OpenFreeEnergy/openfe/issues/842

PR #424, opened 2024-11-22
switch from macos-12 pin to macos-latest

atravitz:
closes https://github.com/OpenFreeEnergy/gufe/issues/313

minimum supported macos should be 14.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,6 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
-          - "3.13"
         include:
             # Note: we still need to add support for macos-13
             # see https://github.com/OpenFreeEnergy/openfe/issues/842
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -33,10 +33,11 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
```
IAlibay, line 36: Looks like we're not 3.13 compatible yet (because of OpenMM 8.2)

IAlibay (changes requested):

.github/workflows/ci.yaml
atravitz: Yep! Just wanted to check. Removing now.

IAlibay (approved):
please do raise an issue re: macos-13 (or reference in an already existing one)

atravitz:
> please do raise an issue re: macos-13 (or reference in an already existing one)

https://github.com/OpenFreeEnergy/gufe/issues/425

PR #459, opened 2025-01-16
add ExecutionInterrupt (special exception)

### LINKED ISSUE #392: Add special `Exception` subclass that won't be caught by `ProtocolUnit.execute`
Currently, in `ProtocolUnit.execute`, all exceptions other than `KeyboardInterrupt` are caught and handled as `ProtocolUnitFailure`s. This is intended behavior. However, there may be cases where a given `ProtocolUnit` should raise an `Exception` that isn't caught in `ProtocolUnit.execute`, but is instead passed upward in the same way as `KeyboardInterrupt` to the calling executor to handle.

I propose we add an `Exception` subclass specifically for this purpose, which itself can be subclassed downstream if desired by `Protocol` authors to give more specificity to the type of unrecoverable error encountered. Otherwise, the current practice is to simply use `KeyboardInterrupt` for this purpose, but it feels a little misguided to do this.

atravitz:
Resolves #392 

Tips

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/gufe/protocols/protocolunit.py
+++ b/gufe/protocols/protocolunit.py
@@ -20,6 +20,7 @@ from os import PathLike
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+from .errors import ExecutionInterrupt
 from ..tokenization import TOKENIZABLE_REGISTRY, GufeKey, GufeTokenizable
 
 
@@ -328,8 +329,8 @@ class ProtocolUnit(GufeTokenizable):
                 end_time=datetime.datetime.now(),
             )
 
-        except KeyboardInterrupt:
-            # if we "fail" due to a KeyboardInterrupt, we always want to raise
+        except KeyboardInterrupt or ExecutionInterrupt:
+            # we always want to raise in these situations
             raise
         except Exception as e:
             if raise_error:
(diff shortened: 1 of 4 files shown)


atravitz:
pre-commit.ci autofix

gufe/protocols/protocolunit.py
```diff
@@ -328,8 +329,8 @@ def execute(
                 end_time=datetime.datetime.now(),
             )
 
-        except KeyboardInterrupt:
-            # if we "fail" due to a KeyboardInterrupt, we always want to raise
+        except (KeyboardInterrupt, ExecutionInterrupt):
+            # we always want to raise in these situations
```
atravitz, line 333: @dotsdl - I'm pretty sure this is redundant, and that `KeyboardInterrupt` and `ExecutionInterrupt` get raised even if this catch isn't here.

dotsdl: Ah, fair point. I think for the sake of understandability later (and seeing where e.g. `ExecutionInterrupt` actually gets used in the codebase), I'd rather leave it in.

dotsdl (approved):
Looks great @atravitz! Please merge if satisfied!

gufe/protocols/protocolunit.py
atravitz: Works for me, I just added a comment so it's documented for clarity.

PR #461, opened 2025-01-16
Enforce protocol settings at instantiation

### LINKED ISSUE #405: Enforce Protocol settings at instantiation
## Problem

The Settings provided to a Protocol are never validated at runtime and incorrect settings can be passed to a Protocol only to cause failures during execution. Since most protocols will require custom settings objects, the types of these settings should be confirmed during instantiation of the Protocols themselves, catching failures early. 

## Solution

We should add a new class attribute `_settings_cls: type[Settings]` to the Protocol base class. This attribute should be left unbound. In the Protocol's `__init__` method, we can perform the following checks:

1. Check if the `_settings_cls` attribute has been defined in the class
2. Check if the provided settings are an instance of the `_settings_cls`

If either of these checks fail, raise:

1. `NotImplementedError` if `_settings_cls` is not defined
2. `ValueError` if the settings provided do not match the expected type

I've implemented a similar pattern in [stratocaster](https://github.com/OpenFreeEnergy/stratocaster/blob/6964024d830624f6110fb4072a6132c01379ee93/src/stratocaster/base/strategy.py#L45) to validate custom settings at instantiation. I believe this approach would be beneficial for Protocols.

### Additional considerations

Note that this will break backwards compatibility for existing Protocols.

ianmkenney:
Require Protocol authors to define the _settings_cls attribute for their class. __init__ method checks that _settings_cls is define (if not raise NotImplementedError) and that the passed settings are an instance of the _settings_cls (if not raise ValueError).

Update existing Protocols in gufe to reflect this change. Closes #405.

Tips

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -95,6 +95,7 @@ class Protocol(GufeTokenizable):
     """
 
     _settings: Settings
+    _settings_cls: type[Settings]
     result_cls: type[ProtocolResult]
     """Corresponding `ProtocolResult` subclass."""
 
@@ -111,6 +112,15 @@ class Protocol(GufeTokenizable):
         Once the Protocol object is created, the input Settings are frozen,
         so should be finalised before creating the Protocol instance.
         """
+
+        if not hasattr(self.__class__, "_settings_cls"):
+            raise NotImplementedError(
+                f"class `{self.__class__.__qualname__}` must implement the `_settings_cls` attribute."
+            )
+
+        if not isinstance(settings, self._settings_cls):
+            raise ValueError(f"`{self.__class__.__qualname__}` expected a `{self._settings_cls.__qualname__}` instance")
+
         self._settings = settings.frozen_copy()
 
     @property
--- a/gufe/protocols/protocol.py
+++ b/gufe/protocols/protocol.py
@@ -94,12 +94,13 @@ class Protocol(GufeTokenizable):
     - `_default_settings`
 
     Additionally, the `_settings_cls` must be set to the intended
-    subclass of `Settings`. This attribute is validated during the
-    instantiation of the Protocol.
+    subclass of `SettingsBaseModel`. This attribute is validated
+    during the instantiation of the Protocol.
+
     """
 
     _settings: Settings
-    _settings_cls: type[Settings]
+    _settings_cls: type[SettingsBaseModel]
     result_cls: type[ProtocolResult]
     """Corresponding `ProtocolResult` subclass."""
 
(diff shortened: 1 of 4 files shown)


gufe/protocols/protocol.py
```diff
@@ -111,6 +116,17 @@ def __init__(self, settings: Settings):
         Once the Protocol object is created, the input Settings are frozen,
         so should be finalised before creating the Protocol instance.
         """
+
+        if not hasattr(self.__class__, "_settings_cls"):
```
IAlibay, line 120: A thought for another issue, we don't do this for `self._settings` or `results_cls`. Would it be worth eventually cleaning up all these abstract classes with expected attributes by adding `__init_subclass__` methods?

IAlibay (approved):
This works for me, thanks!

gufe/protocols/protocol.py
dotsdl: @IAlibay can you elaborate on this in an issue?

dotsdl:
@ianmkenney please merge `main`. I'd do it, but looks like I can't since you used a branch on a clone instead of making a branch on this repo.

gufe/tests/test_protocol.py
```diff
@@ -814,3 +816,47 @@ def test_settings_readonly():
...
+
+        with pytest.raises(
+            NotImplementedError,
+            match=f"class `{TestEnforcedProtocolSettings.ProtocolMissingSettingsClass.__qualname__}` must implement the `_settings_cls` attribute.",
+        ):
+            TestEnforcedProtocolSettings.ProtocolMissingSettingsClass(
+                TestEnforcedProtocolSettings.ProtocolMissingSettingsClass.default_settings()
+            )
+
+    def test_protol_no_settings(self):
```
dotsdl, line 849:
```suggestion
    def test_protocol_wrong_settings(self):
```

dotsdl (changes requested):
One question requiring @IAlibay input, and one typo to fix. Otherwise looks fine; thanks @ianmkenney!

dotsdl:
Also, @ianmkenney can you add a news entry?

gufe/protocols/protocol.py
```diff
@@ -92,9 +92,14 @@ class Protocol(GufeTokenizable):
...
     - `_gather`
     - `_default_settings`
+
+    Additionally, the `_settings_cls` must be set to the intended
+    subclass of `Settings`. This attribute is validated during the
+    instantiation of the Protocol.
     """
 
     _settings: Settings
+    _settings_cls: type[Settings]
```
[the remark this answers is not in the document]
IAlibay: Ah yes you're right, this should be SettingsBaseModel. There are default variables in Settings that don't make sense for some Protocols, so we don't subclass it.

dotsdl (approved):
Thanks @ianmkenney! Merging!

Issue #330, opened 2024-06-25
user-set partial charge warning


JenkeScheen:
With the current setup in `openfe` every time a molecule gets loaded with user-set partial charges the user receives a warning `Partial charges have been provided, these will preferentially be used instead of generating new partial charges`. This becomes a bit much when we're loading hundreds of molecules in production, especially when we load/de-load the set of molecules multiple times during prep and setup steps ahead of simulations. 

Is there a way to log this more elegantly, or potentially expose a flag for users to disable this warning? 

We've now gotten around this by just ignoring this warning but an `openfe`-side solution would be preferable: https://github.com/asapdiscovery/asapdiscovery/blob/main/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/cli.py

IAlibay:
Just coming back to this, we are definitely not removing the warning, the best we can do is change to a logging.warning instead of a userwarning.

JenkeScheen:
sure - as long as it's mutable then that works.

IAlibay:
@atravitz as part of the CLI 1.x "partial charge assignment early" push, we should try to make the switch to logger.INFO or similar here.

atravitz:
closed by #468

PR #468, opened 2025-01-23
switch user-set partial charges warning to log.info

atravitz:
Addresses #330 

We decided to use `logger.info`, since it is encouraged for users to pre-set partial charges, and we shouldn't be yelling at them for that.

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -83,7 +83,7 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
                 raise ValueError(errmsg)
 
     if np.all(np.isclose(p_chgs, 0.0)):
-        wmsg = f"Partial charges provided all equal to " "zero. These may be ignored by some Protocols."
+        wmsg = "Partial charges provided all equal to zero. These may be ignored by some Protocols."
         warnings.warn(wmsg)
     else:
         message = (
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -92,7 +92,7 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
             "partial charges."
         )
         if logger is None:
-            logger = logging.getLogger()
+            logger = logging.getLogger(__name__)
 
         logger.info(message)
 
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -86,7 +86,7 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
         wmsg = f"Partial charges provided all equal to " "zero. These may be ignored by some Protocols."
         warnings.warn(wmsg)
     else:
-        wmsg = (
+        message = (
             "Partial charges have been provided, these will "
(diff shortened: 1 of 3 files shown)


gufe/components/explicitmoleculecomponent.py
```diff
@@ -85,12 +86,15 @@ def _check_partial_charges(mol: RDKitMol) -> None:
...
     else:
-        wmsg = (
+        message = (
             "Partial charges have been provided, these will "
             "preferentially be used instead of generating new "
-            "partial charges"
+            "partial charges."
         )
-        warnings.warn(wmsg)
+        if logger is None:
```
IAlibay, line 94: Probably needs its own "calling the method directly" test.

IAlibay (approved):
Just the one thing otherwise lgtm

gufe/components/explicitmoleculecomponent.py
atravitz: done!

atravitz:
pre-commit.ci autofix

gufe/components/explicitmoleculecomponent.py
```diff
@@ -82,15 +83,18 @@ def _check_partial_charges(mol: RDKitMol) -> None:
...
+            logger = logging.getLogger()
```
dotsdl, line 95:
```suggestion
            logger = logging.getLogger(__name__)
```

dotsdl: This will scope this logging message to a logger corresponding to this module, instead of just the root logger.

dotsdl (approved):
Just one recommended change; otherwise it is good to merge!

gufe/components/explicitmoleculecomponent.py
```diff
@@ -82,15 +83,18 @@ def _check_partial_charges(mol: RDKitMol) -> None:
                 raise ValueError(errmsg)
 
     if np.all(np.isclose(p_chgs, 0.0)):
-        wmsg = f"Partial charges provided all equal to " "zero. These may be ignored by some Protocols."
+        wmsg = "Partial charges provided all equal to zero. These may be ignored by some Protocols."
         warnings.warn(wmsg)
     else:
-        wmsg = (
+        message = (
```
jthorton, line 89: Should we add something that can be used to identify the SMC this is raised for maybe just adding the `ofe-name` prop to the message?

IAlibay (changes requested):
@atravitz I'm going to put a request changes to add Josh's suggestion if you think it's reasonable.

gufe/components/explicitmoleculecomponent.py
atravitz: we don't require SMCs to have names, is there a better identifier?

jthorton: Good point how about `ofe-name` if present else `self.smiles` then we should always be able to workout which molecule the message refers to.

atravitz: From offline discussion,  we decided that adding `self.smiles` would be helpful, but requires a bit of refactoring that we'd like to not include in this PR. 

https://github.com/OpenFreeEnergy/gufe/issues/473

atravitz:
pre-commit.ci autofix

PR #477, opened 2025-02-05
partial charge checking - throw error for negative net charge diff

atravitz:
Currently, if the net charge on a molecule is negative, the `Sum of partial charges differs` error does not get raised. I updated the check so that it's evaluating the absolute value of the difference between expected and calculated net charge.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/gufe/components/explicitmoleculecomponent.py
+++ b/gufe/components/explicitmoleculecomponent.py
@@ -63,7 +63,8 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
         errmsg = f"Incorrect number of partial charges: {len(p_chgs)} " f" were provided for {mol.GetNumAtoms()} atoms"
         raise ValueError(errmsg)
 
-    if (sum(p_chgs) - Chem.GetFormalCharge(mol)) > 0.01:
+    if abs(sum(p_chgs) - Chem.GetFormalCharge(mol)) > 0.01:
+        import pdb;pdb.set_trace()
         errmsg = (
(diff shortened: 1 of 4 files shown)


gufe/components/explicitmoleculecomponent.py
```diff
@@ -63,7 +63,8 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
         errmsg = f"Incorrect number of partial charges: {len(p_chgs)} " f" were provided for {mol.GetNumAtoms()} atoms"
         raise ValueError(errmsg)
 
-    if (sum(p_chgs) - Chem.GetFormalCharge(mol)) > 0.01:
+    if abs(sum(p_chgs) - Chem.GetFormalCharge(mol)) > 0.01:
```
jthorton, line 66: Good catch!

gufe/components/explicitmoleculecomponent.py
```diff
@@ -63,7 +63,8 @@ def _check_partial_charges(mol: RDKitMol, logger=None) -> None:
...
+        import pdb;pdb.set_trace()
```
jthorton, line 67: is this left over from testing?

jthorton (approved):
Great catch, I would add a news entry for the bug fix then its good to go once tests pass.

gufe/components/explicitmoleculecomponent.py
atravitz: yes 🤦

PR #484, opened 2025-02-13
proteincomponent.py updates

### LINKED ISSUE #479: Incorrect Formal Charges or Missing Ions from ProteinComponent
openfe.ProteinComponent module uses incorrect charges for some ions or does not recognize some ions.
In the script, gufe/components/proteincomponent.py, the following are the only ions defined:
```
negative_ions = ["F", "CL", "BR", "I"]
positive_ions = [
    # +1
    "LI",
    "NA",
    "K",
    "RB",
    "CS",
    # +2
    "BE",
    "MG",
    "CA",
    "SR",
    "BA",
    "RA",
    "ZN"
]
```
Which lacks some ions like MN or CU. 

`protein = openfe.ProteinComponent.from_pdb_file("protein.pdb")
`
So, if a protein structure accommodates a Manganese ion

> ValueError: I don't know this Ion or something really went wrong! 	MN	MN	-343	connectivity0

We can manually alter these lists and put MN inside the positive_ions. But then they take a charge of -1, which is not correct, it should be +2. 

I think, it is because of the GetDefaultValence('int') from rdkit.Chem.GetPeriodicTable() where the charges are taken in proteincomponent.py:
```
        # Formal Charge
        netcharge = 0
        for a in rd_mol.GetAtoms():
            atomic_num = a.GetAtomicNum()
            atom_name = a.GetMonomerInfo().GetName()

            connectivity = sum(_BONDORDER_TO_ORDER[bond.GetBondType()] for bond in a.GetBonds())
            default_valence = periodicTable.GetDefaultValence(atomic_num)

            if connectivity == 0:  # ions:
                # strip catches cases like 'CL1' as name
                if atom_name.strip(string.digits).upper() in positive_ions:
                    fc = default_valence  # e.g. Sodium ions
                elif atom_name.strip(string.digits).upper() in negative_ions:
                    fc = -default_valence  # e.g. Chlorine ions
                else:  # -no-cov-
                    resn = a.GetMonomerInfo().GetResidueName()
                    resind = int(a.GetMonomerInfo().GetResidueNumber())
                    raise ValueError(
                        "I don't know this Ion or something really went "
                        f"wrong! \t{atom_name}\t{resn}\t-{resind}\t"
                        f"connectivity{connectivity}"
                    )
            elif default_valence > connectivity:
```
And then if we check the output from rdkit.Chem.periodicTable,
```
from rdkit import Chem

periodicTable = Chem.GetPeriodicTable()

for atomic_number in range(1,31):
    print(atomic_number,periodicTable.GetElementSymbol(atomic_number),
          periodicTable.GetElementName(atomic_number),periodicTable.GetDefaultValence(atomic_number),
          periodicTable.GetNOuterElecs(atomic_number))
```

> 1 H Hydrogen 1 1
> 2 He Helium 0 2
> 3 Li Lithium 1 1
> 4 Be Beryllium 2 2
> 5 B Boron 3 3
> 6 C Carbon 4 4
> 7 N Nitrogen 3 5
> 8 O Oxygen 2 6
> 9 F Fluorine 1 7
> 10 Ne Neon 0 8
> 11 Na Sodium 1 1
> 12 Mg Magnesium 2 2
> 13 Al Aluminium 3 3
> 14 Si Silicon 4 4
> 15 P Phosphorus 3 5
> 16 S Sulfur 2 6
> 17 Cl Chlorine 1 7
> 18 Ar Argon 0 8
> 19 K Potassium 1 1
> 20 Ca Calcium 2 2
> **21 Sc Scandium -1 3**
> **22 Ti Titanium -1 4**
> **23 V Vanadium -1 5**
> **24 Cr Chromium -1 6**
> **25 Mn Manganese -1 7**
> **26 Fe Iron -1 8**
> **27 Co Cobalt -1 9**
> **28 Ni Nickel -1 10**
> **29 Cu Copper -1 11**
> **30 Zn Zinc -1 2**

Even the ZN which is already in positive_ions list receives an incorrect charge of -1, should be +2.
So, we see that transition metals have a default valence of -1, meaning that rdkit.Chem.periodicTable has a problem with GetDefaultValence(), transition metals valencies are not correct, many have also multiples of different valences. 

However, 
[In rdkit/Code/GraphMol/atomic_data.cpp](https://github.com/rdkit/rdkit/blob/c7169eb17c1e5ab212d5a84926180413ef477204/Code/GraphMol/atomic_data.cpp#L229):

> list of allowed valences, -1 for anything

It is stated in their code, elements with default_valence=-1 actually means any valency. So it can't be used for transition metals.

**To Reproduce**
protein.pdb:

> HETATM 1  CU   CU A 1       5.266  16.224  11.785  1.00  9.96          CU

```
import openfe

protein = openfe.ProteinComponent.from_pdb_file("./protein.pdb")
systemA = openfe.ChemicalSystem({
    'protein': protein,
})
print(systemA.total_charge)
```

**Software versions**
  - MacOS 14.2, $CONDA_SUBDIR=osx-64, openfe==1.2.0, rdkit==2024.03.5
  - confa-forge
  - [conda list](https://justpaste.it/fodhi)

**Output**

> ValueError                                Traceback (most recent call last)
> Cell In[31], line 1
> ----> 1 protein = openfe.ProteinComponent.from_pdb_file("./protein.pdb") ## MG
> 
> File ~/anaconda3/envs/open_env/lib/python3.12/site-packages/gufe/components/proteincomponent.py:125, in ProteinComponent.from_pdb_file(cls, pdb_file, name)
>     109 """
>     110 Create ``ProteinComponent`` from PDB-formatted file.
>     111 
>    (...)
>     122     the deserialized molecule
>     123 """
>     124 openmm_PDBFile = PDBFile(pdb_file)
> --> 125 return cls._from_openmmPDBFile(openmm_PDBFile=openmm_PDBFile, name=name)
> 
> File ~/anaconda3/envs/open_env/lib/python3.12/site-packages/gufe/components/proteincomponent.py:230, in ProteinComponent._from_openmmPDBFile(cls, openmm_PDBFile, name)
>     228         resn = a.GetMonomerInfo().GetResidueName()
>     229         resind = int(a.GetMonomerInfo().GetResidueNumber())
> --> 230         raise ValueError(
>     231             "I don't know this Ion or something really went "
>     232             f"wrong! \t{atom_name}\t{resn}\t-{resind}\t"
>     233             f"connectivity{connectivity}"
>     234         )
>     235 elif default_valence > connectivity:
>     236     fc = -(default_valence - connectivity)  # negative charge
> 
> ValueError: I don't know this Ion or something really went wrong! 	CU	CU	-3	connectivity0

**Expected behavior**
A complete list of ions (including different oxidization states for most common ones such as MN: MN2 and MN3) with correct charges

**Maybe a solution**
Get the formal charges like the following,
```
for a in rd_mol.GetAtoms():
.
.
.

            if connectivity == 0:  # ions:
                    fc = a.GetFormalCharge()
```
This looks at the last column of the PDB for the charges:

Instead of,
> HETATM 1  CU   CU A 1       5.266  16.224  11.785  1.00  9.96          CU,

we can use this notation for ions in PDBs,

> HETATM 1  CU   CU A 1       5.266  16.224  11.785  1.00  9.96          Cu+2

or

> HETATM 1  CU   CU A 1       5.266  16.224  11.785  1.00  9.96          Cu+3
> HETATM 1  MN   MN A 1       5.266  16.224  11.785  1.00  9.96          Mn+2
> HETATM 1  MN   MN A 1       5.266  16.224  11.785  1.00  9.96          Mn+3

Then there is no need to define lists of ions in proteincomponent.py.

Mert-csrg:
[Fixes #479](https://github.com/OpenFreeEnergy/gufe/issues/479 )

Checklist
* [x] Added a ``news`` entry


Mert-csrg:
pre-commit.ci autofix

gufe/components/proteincomponent.py
```diff
@@ -211,25 +266,22 @@ def _from_openmmPDBFile(cls, openmm_PDBFile: PDBFile | PDBxFile, name: str = "")
...
 
-            if connectivity == 0:  # ions:
-                # strip catches cases like 'CL1' as name
-                if atom_name.strip(string.digits).upper() in positive_ions:
-                    fc = default_valence  # e.g. Sodium ions
-                elif atom_name.strip(string.digits).upper() in negative_ions:
-                    fc = -default_valence  # e.g. Chlorine ions
-                else:  # -no-cov-
+            if connectivity == 0:  # ions
+                ion_key = atom_name.strip().upper()
```
jthorton, line 276: Using upper here will give the wrong charge for ions that have multiple possible values which are determined by the name being all upper or a mix of upper and lower like `CE` and `Ce`.  Meaning we need to do a few different checks to cover all of the different possible input cases, maybe something like:
- is the unedited ion key in the list
- is the upper ion key in the list 
- is the upper digit stripped ion key in the list.
We should also add test cases which cover all of these checks.

jthorton (dismissed):
Thanks for the PR @Mert-csrg this is a good start, I think we need to clean up how the ion keys are checked, however,  to make sure we don't introduce more issues. We should also add tests to cover this improved function to make sure we are now getting correct charges for Zinc ions for example and correctly handling cases like `CL1` which were a concern in the old method.

gufe/components/proteincomponent.py
```diff
@@ -51,23 +51,78 @@
...
+    "Tm": 3,
+    "YB2": 2,
+    # Actinides
+    "Th": 4,
+    "U4+": 4,
+    "Pu": 3,
+    # Halogens
+    "F": -1,
+    "CL": -1,
+    "Cl-": -1,
```
atravitz, line 114: @Mert-csrg what is the original source for this list?

Mert-csrg: It is based on an AmberTools library for the atomic ions, I extended it with some common ionic molecules, halogens and some uncommon residue name variations. Original version of that list can be found here,

`$AMBERHOME/dat/leap/lib/atomic_ions.lib
`

As far as I know all the water models call the same library for ions.

atravitz: Thanks @Mert-csrg! This is definitely an improvement over our existing options, but we want to have record of the sources for all of these entries.

Could you provide a github permalink for the items pulled from ambertools, and then either annotate or split into a separate section any ions that you added in addition?

gufe/components/proteincomponent.py
atravitz: I agree and will add this!

atravitz: @jthorton how sure are you that we should be trying to match digit-stripped values? Do you think it could lead to confusion, or are cases like `CL1` pretty common that a user would expect it to be cast to `CL`?

atravitz:
pre-commit.ci autofix

gufe/components/proteincomponent.py
Mert-csrg: [Permalink to Ambertools Lib](https://github.com/Amber-MD/AmberClassic/blob/42e88bf9a2214ba008140280713a430f3ecd4a90/dat/leap/lib/atomic_ions.lib#L1C1-L68C6)

Pull requested script does not accommodate any additional ions. Sorry about that. I have an another script with an extended list that has different oxidation states for transition metals (such as MN2, MN3). But since it is a bit more complicated to use I didn't put it in here, to not cause any more problems.

atravitz: @Mert-csrg -  the list in this PR is great and you don't need to add any more ions. However, we _would_ like to know the original source of each of the ions in this PR.

Could you add comments (similar to how you added comments for alkali metals, transition metals, etc.) that document the original sources, since you mentioned that you extended the list beyond the ambertools lib:

> It is based on an AmberTools library for the atomic ions, I extended it with some common ionic molecules, halogens and some uncommon residue name variations.

Mert-csrg: @atravitz All of them are from that amber library. I apologize for the confusion I made a mistake by saying "I extended the list" as I keep that extended list for my own use and I PR'd the script that is purely and completely based on the original list from Amber.

atravitz: actually, @Mert-csrg - it looks like this list is not identical to the `atomic_ions.lib` you shared. It is missing at least `Tl` and `V2+`.

I will copy over the list from `atomic_ions.lib` unless there is an explicit reason you excluded some of these ions.

Mert-csrg: @atravitz In amber lib, Vanadium in its +2 oxidation state however the most common is +5, I chose not to include that. I guess, I missed Tl but not TL since Thallium has two common ions, +1 and +3 charged, TL is for +1, Tl is for +3 in that library. I think adding those makes no harm.

atravitz:
precommit.ci autofix

gufe/components/proteincomponent.py
atravitz: @Mert-csrg  - I made the following changes to the ions list in this PR to make it better match the amber lib source:

 - alphabetized the list to make comparison to the amber source easier
 - removed ND and SN, which are not listed in the amber source
 - corrected `Pu` to have a charge of 4 (it was listed as 3). I believe this was a typo.

gufe/components/proteincomponent.py
atravitz: I've decided to not support digit stripped values in this PR  - with the idea of starting stricter so that we can be more lax in a follow-up PR if we chose.

jthorton: I don't think its too common but this is a regression as PDBs with CL1 that did work now won't. I am not sure how we make that clear to users or if we should still support those cases.

dotsdl: Although I would normally be partial to your approach @atravitz, I think I agree with @jthorton on this case. Attempting to match digit stripped values would at least be a last resort here.

I think the changes in the PR are a welcome improvement, and reducing surprise from users for which the current behavior was working fine is worth us doing if possible.

dotsdl (changes requested):
Looking great @Mert-csrg and @atravitz! I believe we should take @jthorton's suggestion and add digit stripping matches as a last resort for ions, in order to reproduce current intended behavior at least.

gufe/components/proteincomponent.py
atravitz: I agree with @jthorton as well - I didn't realize it would be a regression. Thanks josh!

atravitz:
pre-commit.ci autofix

dotsdl (approved):
Looks great @atravitz! Thanks for the modifications!

Feel free to merge if satisfied!

PR #489, opened 2025-02-20
Use correct default shell

mikemhenry:
Tips

Checklist
* [ ] Added a ``news`` entry

This makes sure we use a default shell that will throw errors has they happen. This also drops the rdkit pin in the mypy tests since that has been fixed upstream.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -127,7 +127,6 @@ jobs:
       - name: "install checks"
         working-directory: ./dist
         run: |
-          pip check
           python -Ic "import gufe; print(gufe.__version__)"
 
       - name: "run tests"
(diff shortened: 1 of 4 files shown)


mikemhenry:
pre-commit.ci autofix

atravitz (approved):

mikemhenry:
Lol, another error that went unfixed with that default shell

atravitz (approved):

atravitz:
@mikemhenry not sure why mypy is linting within `/lib/python3.10/site-packages` now - is this a known issue?

mikemhenry:
Yes known issue with some rdkit versions, had the same error with openfe as well and just had to pin to the right rdkit version

mikemhenry:
looks like in openfe we also pin the rdkit version in mypy, so we are using the same main pin now (lets see if anything else breaks!)

PR #493, opened 2025-03-03
Update vendored files to match OpenMM 8.2.0 behavior when reading formal chages

LilDojd:
This PR includes changes to the vendored files to properly read and write formal charge field in .pdb files.
Also changes `proteincomponent.py` to use charges from the .pdb, if provided

Part of #410
Depends on #484 

Tips

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/gufe/components/proteincomponent.py
+++ b/gufe/components/proteincomponent.py
@@ -199,9 +199,7 @@ class ProteinComponent(ExplicitMoleculeComponent):
         return cls._from_openmmPDBFile(openmm_PDBFile=openmm_PDBxFile, name=name)
 
     @classmethod
-    def _from_openmmPDBFile(
-        cls, openmm_PDBFile: Union[PDBFile, PDBxFile], name: str = ""
-    ):
+    def _from_openmmPDBFile(cls, openmm_PDBFile: Union[PDBFile, PDBxFile], name: str = ""):
         """Converts to our internal representation (rdkit Mol)
 
         Parameters
@@ -256,9 +254,7 @@ class ProteinComponent(ExplicitMoleculeComponent):
 
         # Set Positions
         rd_mol = editable_rdmol.GetMol()
-        positions = np.array(
-            openmm_PDBFile.positions.value_in_unit(omm_unit.angstrom), ndmin=3
-        )
+        positions = np.array(openmm_PDBFile.positions.value_in_unit(omm_unit.angstrom), ndmin=3)
 
         for frame_id, frame in enumerate(positions):
             conf = Conformer(frame_id)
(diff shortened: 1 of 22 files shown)


LilDojd:
pre-commit.ci autofix

LilDojd:
closed by accident

IAlibay (changes requested):
I'm going to block this on the grounds of "lots of new behaviour and insufficient existing tests". We really need a lot more testing of this, including all the cases with different types of ions and when you have odd ion names (e.g. CL1).

gufe/components/proteincomponent.py
```diff
@@ -210,36 +265,38 @@ def _from_openmmPDBFile(cls, openmm_PDBFile: Union[PDBFile, PDBxFile], name: str
...
+                fc = omm_atom.formalCharge
+            else:
+                atom_name = rd_atom.GetMonomerInfo().GetName().strip()
+                atomic_num = rd_atom.GetAtomicNum()
+
+                connectivity = sum(_BONDORDER_TO_ORDER[bond.GetBondType()] for bond in rd_atom.GetBonds())
+                default_valence = periodicTable.GetDefaultValence(atomic_num)
+
+                if connectivity == 0:  # ions
+                    ion_key = atom_name.strip().upper()
```
jthorton, line 281: We need to make sure we catch cases like CL1 and strip digits from the atom name similarly to the old method.

gufe/components/proteincomponent.py
```diff
@@ -51,23 +51,78 @@
...
-]
+ions_dict = {
+    # Alkali metals
+    "LI": 1,
+    "NA": 1,
+    "K": 1,
+    "RB": 1,
+    "CS": 1,
+    "K+": 1,
+    "Na+": 1,
```
jthorton, line 62: I don't think names which are a combination of upper and lower characters can be matched as we are getting the ion name as upper characters only. So this list needs to all be in upper case.

gufe/tests/test_proteincomponent.py
```diff
@@ -278,7 +303,7 @@ def test_hash_neq_name(self, PDB_181L_path):
     def test_protein_total_charge(self, PDB_181L_path):
         m1 = self.cls.from_pdb_file(PDB_181L_path)
 
-        assert m1.total_charge == 7
+        assert m1.total_charge == 20
```
jthorton, line 306: This is a large difference in the calculated total charge, can we check where this is coming from, ie what was assigned the wrong formal charge in the old version?

LilDojd:
Hi! Will come back to this in a week or so

Please let me know when you have an internal discussion regarding eventually supporting OpenMM 8.2.0. I have realized, that this PR needs much more thought put into it. Additionally, we might want to retain backward compatibility of vendored files with versions of OpenMM before 8.2.0

dotsdl:
@LilDojd thank you for this!

We're going to proceed with #484 to address the incorrect formal charge issues. We will then later consider this PR for OpenMM 8.2.0 compatibility.

atravitz:
Hi @LilDojd - thank you for all your work on this! We ended up addressing the openmm 8.2 support in #484 and #539 with a different approach, so I'm going to go ahead and close this PR.

Please re-open if there's anything you'd like to follow up on!

PR #502, opened 2025-03-06
Protocol validation method

### LINKED ISSUE #412: Add `Protocol.validate` method that takes same inputs as `create`; add a passthrough in `Transformation`
In practice, a `Protocol` can fail during execution due to poor input choices to their `create` methods, such as incompatibilities between `stateA` and `stateB`, `mapping` issues, or conflicts with any of these and `settings` choices. This can trip users up, cost them time in diagnosing problems in (potentially remote) execution, and waste expensive compute on transformations that are certain to fail.

One option is to encourage `Protocol` authors to add validation checks for known limitations like the above to their `_create` methods, but this would slow down execution by remote execution systems every time the `create` method is called.

Instead, we propose adding a distinct `validate` method (with corresponding `_validate` method that `Protocol` authors specify) that features largely the same signature as `create`. This method would allow `Protocol` authors to specify (reasonably) rapid validation code specific to their `Protocol` that can catch obvious issues before performing potentially expensive compute on a set of inputs that are bound to fail.

ianmkenney:
This PR introduces the `_validate` abstract method and the `validate` method to the `Protocol` class. The type signatures largely reflect those of the `create` method as they provide a convenient way to validate the inputs before being passed `create`.

closes #412

Tips

Checklist
* [x] Added a ``news`` entry


ianmkenney:
Since `Transformation` objects only take `extends` when calling `create`, there doesn't appear to be a clean way to validate those values during network creation, as stated in the original issue. This can still be done with a raw `Protocol` object.

gufe/protocols/protocol.py
```diff
@@ -270,6 +268,78 @@ def create(
...
+        stateB: ChemicalSystem,
+        mapping: Optional[Union[ComponentMapping, list[ComponentMapping]]],
+        extends: Optional[ProtocolDAGResult] = None,
+    ):
+        """Method to override in custom `Protocol` subclasses.
+
+        This method should take two `ChemicalSystem`s, and optionally
+        one or more `ComponentMapping` objects and `ProtocolDAGResult`
+        objects used for extensions, and validate them, raising a
+        `ValueError` where validation is not successful.
```
jthorton, line 285: I think we should start pointing people to our new validation errors https://github.com/OpenFreeEnergy/gufe/blob/ab2ca232365e78c5a252bbdfdf4bc03c6277fee6/gufe/protocols/errors.py#L10

gufe/protocols/protocol.py
```diff
@@ -270,6 +268,78 @@ def create(
...
+            stateA and stateB :class:`.ChemicalSystem`.
+        extends : Optional[ProtocolDAGResult]
+            If provided, the :class:`.ProtocolDAGResult` from a
+            previous execution may be validated against the current
+            :class:`.Protocol` object's ``Settings``, depending on the
+            ``Protocol`` implementation.
+
+        Raises
+        ------
+        ValueError
```
jthorton, line 325: same as above lets move to our new errors!

jthorton (changes requested):
The only thing I would consider blocking is to use the new protocol validation errors, otherwise this looks fantastic! It would be good to get others thoughts on providing some basic validation as standard that might be better for another PR however.

gufe/protocols/protocol.py
ianmkenney: Spoken like the author of such an exception! Done!

gufe/tests/test_protocol.py
```diff
@@ -147,6 +147,31 @@ def _create(
...
+                mapping = [mapping]
+
+            for element in mapping:
+                if not isinstance(element, ComponentMapping):
+                    raise ValueError("A non-ComponentMapping object provided as a mapping.")
+
+        if extends:
+            if not isinstance(extends, ProtocolDAGResult):
+                raise ValueError("A non-ProtocolDAGResult object provided as extends.")
+
```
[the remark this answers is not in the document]
ianmkenney: I agree it's probably decent first pass. Curious how others feel about it.

ianmkenney:
@jthorton, given @IAlibay's comment in the issue concerning the timing of this feature, I suspect that this might be parked until v2. We might as well add the default validation if it's a wanted feature in this PR.

ianmkenney:
@jthorton, @atravitz, and @dotsdl, this is good for another review

gufe/tests/test_protocol.py
```diff
@@ -256,6 +257,53 @@ def protocol_dag_broken(self, solvated_ligand, vacuum_ligand, tmpdir):
...
+        with pytest.raises(ProtocolValidationError, match=expected_msg):
+            instance.validate(stateA=solvated_ligand, stateB=vacuum_ligand, mapping="Not a mapping".split(" "))  # type: ignore
+
+        # Test bad extends
+        with pytest.raises(ProtocolValidationError, match="A non-ProtocolDAGResult object provided as extends."):
+            instance.validate(stateA=solvated_ligand, stateB=vacuum_ligand, mapping=mapping, extends="No thank you")  # type: ignore
+
+    def test_author_validation(self, instance: DummyProtocol, solvated_ligand, vacuum_ligand):
+
+        error_msg = "An intentional exception from a very picky author"
```
dotsdl, line 285: Haha

gufe/transformations/transformation.py
```diff
@@ -182,6 +183,13 @@ def __init__(
             of the two states.
         name : str, optional
             A human-readable name for this transformation.
+        validate: bool, optional
```
dotsdl, line 186: I like this exposure.

dotsdl (approved):
Looks great @ianmkenney! As discussed, this shouldn't impact users downstream if merged pre-v2.0.

jthorton (approved):
Great job @ianmkenney this should work very well! 

One thought maybe for another PR,  do we also want validate to run during the protocol create method? My thinking here is that if we are going to encourage protocol authors to move all validation to the validate method to get useful errors for users then should we run it during the create method as well.

dotsdl:
An argument against running validation during `Protocol.create` is that `create` is often run by a downstream executor. In `alchemiscale`'s case, `create` is currently run on a compute service, but in the future may be run on server-side services. Although `validate` should still be cheap compared to `ProtocolDAG` execution, it can very well be more expensive than `create`.

atravitz (approved):

PR #503, opened 2025-03-07
[DNM] Prototype using pablo to load pdb files

jthorton:
This is a quick investigation into using pablo to load pdb files for the protein components.

# Notes
- due to openff topology having a nice interface to openmm topology the changes needed to load pdb files are minimal 
- comparing the topology created by loading from openmm vs pablo are different (this just seems to be extra information like bond orders etc are added by Pablo)
- pdbx and mmcif are not supported yet
- Some tests fail, we should look into if this is a pablo issue or an issue with the files
- we should make sure we can load all of the industry benchmark systems with pablo to have parity with the current loader.

#Questions

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -24,5 +24,8 @@ dependencies:
   - pydata-sphinx-theme
   - sphinx-jsonschema==1.15
   - sphinx <7.1.2
+  # For pablo testing
+  - pyxdg
   - pip:
     - autodoc_pydantic<2.0.0
+    - git+https://github.com/openforcefield/openff-pablo.git@v0.0.1a1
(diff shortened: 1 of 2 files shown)


IAlibay:
Some initial questions: 

- How is the performance? If you load in a large protein, do you get any noticeable differences in performance? 
- What dependencies does Pablo require? Do we need to pull the entire OpenFF toolkit? (including interchange?)
- Is atom and residue order preserved?
- Does it roundtrip well? 
- How easy is it to pass through new definitions for things that aren't in the inside CCD dictionary?
- Does it behave well when you have a specific protonation state? For example protonated GLU.
- How does it recognise CYS bonds? Will it attempt to find them automatically (like. PDBFile) or does it require them to be explicitly defined?

PR #512, opened 2025-03-18
Fix openmm bonds types for `order` and `type` in `to_openmm_topology`

### LINKED ISSUE #501: ProteinComponent.to_openmm_topology() sets wrong bond order
**Describe the bug**

`ProteinComponent.to_openmm_topology()` sets `Bond.order` to a bond type instead (`Single`, `Double`, `Triple`, `Aromatic`  or `Amide`) in the resulting OpenMM Topology.

**To Reproduce**
```python
>>> from gufe import ProteinComponent
>>> prot = ProteinComponent.from_pdb_file("gufe/tests/data/181l.pdb")
>>> prot.to_openmm_topology().bonds().__next__().__repr__() # fastest way to get error

...
File ~/miniforge3/envs/gufedev/lib/python3.12/site-packages/openmm/app/topology.py:507, in Bond.__repr__(self)
    505     s = "%s, type=%s" % (s, self.type)
    506 if self.order is not None:
--> 507     s = "%s, order=%d" % (s, self.order)
    508 s += ")"
    509 return s

TypeError: %d format: a real number is required, not Single
```

**Software versions**
  - gufe (main head)
  - openmm 8.1.2
  - python 3.12

ijpulidos:
Fixes #501 by adding the order information as an `int` or `None` as expected by OpenMM. It additionally adds some `type` information when generating the topology. 

This needed to modify the mapping between `rdkit` types and `openmm` orders. And additionally, a dictionary mapping between `rdkit` types and `openmm` types (not orders) was created.

Tips

Checklist
* [x] Added a ``news`` entry


ijpulidos:
@mikemhenry yes, that's a good point. I think I can try writing some test for it, the only caveat I see is that the OpenMM topology doesn't validate the types, so this is one of those that only fails downstream when you actually try to use the bonds attribute in some way, for example when converting an `openmm` topology to an `mdtraj` one. But I think I can come up with a test for it. Thanks!

jthorton:
Hi @ijpulidos see the example in this issue that should be enough for a test https://github.com/OpenFreeEnergy/gufe/issues/501#issue-2896674094 hope this helps!

ijpulidos:
@jthorton Oh right, yes, that's a good one. Thanks for pointing that out. I'll work on the changes now.

gufe/tests/test_proteincomponent.py
```diff
@@ -239,6 +239,21 @@ def test_to_openmm_topology(self, in_pdb_path):
         assert openmm_top.getNumChains() == gufe_openmm_top.getNumChains()
         assert openmm_top.getNumResidues() == gufe_openmm_top.getNumResidues()
         assert openmm_top.getNumChains() == gufe_openmm_top.getNumChains()
+        # Make sure bond.order is the expected type int or None
+        assert all(isinstance(bond.order, (int, type(None))) for bond in openmm_top.bonds())
+
+    @pytest.mark.parametrize("in_pdb_path", ALL_PDB_LOADERS.keys())
+    def test_protein_component_openmm_bond(self, in_pdb_path):
+        """
+        Test that `to_openmm_topology().bonds()` produces a valid bond object.
```
mikemhenry, line 248:
```suggestion
        Test that `to_openmm_topology().bonds()` produces a valid bond object.
        See https://github.com/OpenFreeEnergy/gufe/issues/501
```

mikemhenry: Just want to make it easier to see how we came up with this

gufe/tests/test_proteincomponent.py
```diff
@@ -239,6 +239,21 @@ def test_to_openmm_topology(self, in_pdb_path):
...
+        """
+        in_pdb_io = ALL_PDB_LOADERS[in_pdb_path]()
+        prot = self.cls.from_pdb_file(in_pdb_io, name="Alice")
+        omm_topology_bonds = prot.to_openmm_topology().bonds()
+
+        with pytest.raises(StopIteration):
+            while True:
+                repr(next(omm_topology_bonds))
```
mikemhenry, line 256: Is there a reason we don't iterate over the iterator and just call `repr` on each object? While this works, it seems odd that do it like this. I think if you do have a good reason for doing it like, I would instead use a `try`/`except` block here, since I tend to use `pytest.raises` when I am testing to see if my code raises an exception that I expect/want to test. Here we just don't care if we exhaust the iterator.

news/fix_to_openmm_topology_bonds.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed behavior of `ProteinComponent.to_openmm_topology` to correctly assign `int` or `None` types to bond order for OpenMM, instead of objects. It now also adds some bond type information as objects, if detected.
```
mikemhenry, line 19:
```suggestion
* Fixed behavior of `ProteinComponent.to_openmm_topology` to correctly assign `int` or `None` types to bond order for OpenMM, instead of objects. It now also adds some bond type information as objects, if detected. (#501)
```

mikemhenry (changes requested):
Thanks! Really just have a question on how we want to iterate over all of the bonds

ijpulidos:
> Do we have a small example that triggers this bug? I'd like to get a regression test added. I think the test you added is good to test our bond order function, but I am surprised we did hit this in our testing

I think there's some confusion here. That test (the first one I wrote) is using openmm's bond order function, and it would fail without the changes in this PR. So, I actually think that should be enough, at least in terms of having valid bond orders.

gufe/tests/test_proteincomponent.py
```diff
@@ -239,6 +239,22 @@ def test_to_openmm_topology(self, in_pdb_path):
         assert openmm_top.getNumChains() == gufe_openmm_top.getNumChains()
         assert openmm_top.getNumResidues() == gufe_openmm_top.getNumResidues()
         assert openmm_top.getNumChains() == gufe_openmm_top.getNumChains()
+        # Make sure bond.order is the expected type int or None
+        assert all(isinstance(bond.order, (int, type(None))) for bond in openmm_top.bonds())
```
ijpulidos, line 243: @mikemhenry  Maybe we should just add a new line that tests for the types of `bond.type` here and it should be enough as regression tests. As in, if we ever go back to NOT having the correct types, these two lines in this test will fail.

ijpulidos: However, adding check for `bond.type` is a bit ugly, since it depends on the custom types/objects defined in openmm (`Single`, `Double`, etc.)

gufe/tests/test_proteincomponent.py
ijpulidos: No special reason other than not being able to be 100% sure that there's always going to be a `TypeError` exception when we use the `Bond.__repr__` method from OpenMM. That is, if I'm understanding you, if we try to catch the `TypeError` exception, there's no guarantee that the `repr(bond)` wouldn't raise any other exception that we would just not catch. Does this make sense?

I agree this is commonly not the way we expect `pytest.raises` to be used, but in this case we are using OpenMM's `Bond.__repr__` as a validator and I want to test for success rather than failure, and `StopIteration` is kind of a success in this case. I may be overthinking this, but I accept suggestions.

news/fix_to_openmm_topology_bonds.rst
ijpulidos: On a second thought, I actually think the part that adds some bond type information is a _change_ rather than a _fix_. Does that part make more sense to add it as a change?

mikemhenry: Hmm, I guess it is kinda a philosophical question but since the thing we were doing was causing an error, I would call it a fix. We had to change things to fix it, but I would argue that they way we were doing was incorrect.

ijpulidos:
@mikemhenry  I added the suggestions and also where we landed from external independent discussions. I think this should be ready to be re-reviewed. Thanks for the feedback!

mikemhenry (approved):

PR #530, opened 2025-04-18
Included concepts and extensible points docs

dotsdl:
Morphing the core concepts doc into a concept doc listing the data models **gufe** includes, and detailing:
- what information each contains
- where they are typically used
- how they relate together

Decided this was the better approach than a separate extensible points doc, which would need to be referenced from this doc as well, requiring too much redirection.


dotsdl:
@atravitz I've added a section for every included model in `gufe` (except perhaps `Settings`, but that might need to get its own treatment as part of the `Protocol` how-to). I've filled in almost all of them, with the exception of:
- `ComponentMapping`
- `AtomMapping`
- `AtomMapper`
- `LigandNetwork`

I don't have much experience working with or on those objects, so I wanted to work with you directly on populating their descriptions. Would you be willing to do this? I'm happy to pair up with you live to do this if you'd prefer as well.

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,256 @@
...
+    The :class:`.Protocol` is an *extensible point* of the library,
+    and is intended to be subclassed to enable new applications.
+    For details on how to create your own :class:`.Protocol` classes, see :ref:`howto-protocol`.
+
+.. _protocoldag:
+
+``ProtocolDAG``
+^^^^^^^^^^^^^^^
+
+A :class:`.ProtocolDAG` is an executable artifact that performs a :ref:`Protocol <protocol>`.
```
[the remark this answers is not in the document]
dotsdl: Switched to "object" for clarity.

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,256 @@
...
+
+A ``Transformation`` functions as a container for all the information needed to obtain an estimate of the free energy difference between its two ``ChemicalSystem``\s.
+
+
+.. _nontransformation:
+
+``NonTransformation``
+---------------------
+
+A :class:`.NonTransformation` represents non-alchemical sampling of a single :ref:`ChemicalSystem <chemicalsystem>`.
```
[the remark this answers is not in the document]
dotsdl: Discussed in call; agreed that current wording sufficiently describes what a `NonTransformation` is, without going into detail as to why we designed it this way.

dotsdl:
@atravitz I've added a new diagram that should help clarify how the pieces of the `Protocol` system work together!

dotsdl:
Looks like neither of the diagrams we use render well in dark mode since they use a transparent background with black foreground elements. Think we should put an explicit white background on these?

dotsdl (commented):
Added some suggestions @atravitz! Mostly trying to avoid the word "stores", since that gives a different connotation to me than how these objects reference each other.

Thank you for adding these!!

docs/_static/gufe_protocol_diagram.svg
atravitz, line 1: thank you for making this! My only concern is that it's going to have the same issue as here: https://github.com/OpenFreeEnergy/gufe/issues/563

how do you feel about me adding an opaque background for the sake of darkmode? and also trimming it down to fit the bounds of the diagram so it doesn't take up more space than necessary.

atravitz:
> Looks like neither of the diagrams we use render well in dark mode since they use a transparent background with black foreground elements. Think we should put an explicit white background on these?

Yes I think that's a good idea, as well as trimming them to fit the bounds of the diagrams.

atravitz:
pre-commit.ci autofix

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
+Data models included in **gufe**
+================================
+
+The core of the **gufe** data model is the :class:`.GufeTokenizable` class,
+but **gufe** features more than just this base data structure.
+
+In order to ensure interoperability,
+**gufe** also defines classes of objects that represent the core chemistry and alchemistry of a free energy pipeline,
+including molecules, chemical systems, and alchemical transformations.
+This provides a shared language that tools in the ecosystem use.
```
atravitz, line 10:
```suggestion
This provides a shared language that is used by tools across the OpenFE ecosystem.
```
Maybe this is more precise?

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
...
+
+.. _component:
+
+``Component``
+-------------
+
+The :class:`.Component` class represents a portion of a system of molecules,
+with a single ``Component`` capable of representing anything from an individual drug-like molecule, an entire protein, or (even the concept of) a solvent with ions.
+
+These are often used to define the *components* of a :ref:`chemicalsystem`, which form the nodes of an :ref:`alchemicalnetwork`.
```
atravitz, line 41:
```suggestion
``Components`` are often used as the building blocks of :ref:`chemicalsystem`, which form the nodes of an :ref:`alchemicalnetwork`.
```
Saying that a component is a component isn't super clear.

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
...
+.. _chemicalsystem:
+
+``ChemicalSystem``
+------------------
+
+A :class:`.ChemicalSystem` represents a complete system of molecules,
+and is often composed of multiple :ref:`Components <component>`.
+
+These are most often used as nodes of an :ref:`alchemicalnetwork`, with pairs of ``ChemicalSystem``\s connected by :ref:`Transformations <transformation>`.
+Because a ``ChemicalSystem`` functions as a kind of container of :ref:`Components <component>`, more than one ``ChemicalSystem`` can feature the same ``Component``\s.
```
atravitz:
```suggestion
Because a ``ChemicalSystem`` functions as a kind of container of references to :ref:`Components <component>`, more than one ``ChemicalSystem`` can feature the same ``Component``\s.
```

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
...
+
+.. _transformation:
+
+``Transformation``
+------------------
+
+A :class:`.Transformation` represents an alchemical transformation between two :ref:`ChemicalSystems <chemicalsystem>`.
+
+``Transformation`` objects are often used as the edges of an :ref:`alchemicalnetwork`.
+In addition to referencing the ``ChemicalSystem``\s it spans,
```
atravitz, line 77:
```suggestion
In addition to referencing the two ``ChemicalSystem``\s it spans,
```

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
...
+
+.. _nontransformation:
+
+``NonTransformation``
+---------------------
+
+A :class:`.NonTransformation` represents non-alchemical sampling of a single :ref:`ChemicalSystem <chemicalsystem>`.
+
+In the context of an :ref:`alchemicalnetwork`, a ``NonTransformation`` is effectively a self-loop, featuring the same ``ChemicalSystem`` at either end.
+Similar to a :ref:`Transformation <transformation>`, it features a :ref:`protocol` used to perform sampling on its ``ChemicalSystem``, but does not feature a :ref:`componentmapping` since none is required for this.
```
atravitz, line 92:
```suggestion
Similar to a :ref:`Transformation <transformation>`, it features a :ref:`protocol` used to perform sampling on its ``ChemicalSystem``, but does not feature a :ref:`componentmapping`, since there is no second ``ChemicalSystem``.
```
it's not just not required, it cannot/should not be defined.

atravitz:
@dotsdl I've added as comments things that are suggestions, but might need more conversation or you can veto. I've also added semi-opaque white backgrounds, so that it looks transparent in light-mode and like a very light grey in dark mode (it's a little gentler on the eyes than stark white against black in dark mode).

I've also added a `LigandNetwork` diagram, trying to match the `AlchemicalNetwork`'s convention as much as possible for clarity.

atravitz:
pre-commit.ci autofix

docs/concepts/included_models.rst
```diff
@@ -0,0 +1,286 @@
...
+    5. :ref:`AtomMapper <atommapper>` : :ref:`howto-atommapper`
+
+
+.. _component:
+
+``Component``
+-------------
+
+The :class:`.Component` class represents a portion of a system of molecules,
+with a single ``Component`` capable of representing anything from an individual drug-like molecule, an entire protein, or (even the concept of) a solvent with ions.
```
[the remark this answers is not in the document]
dotsdl: Works for me! Thank you!

PR #532, opened 2025-04-20
Move JSON (de)serialization to the serialization subpackage

### LINKED ISSUE #516: Move JSON (de)serialization logic into the new `gufe.serialization` subpackage
Since #492 introduced the `gufe.serialization` subpackage to host the MessagePack serialization code, the JSON serialization code in `custom_json.py` and `custom_codecs.py` should be moved there as well. Additionally, it might make sense to merge these two files into a single JSON serialization module.

ianmkenney:
This PR moves the `gufe.custom_json` and `gufe.custom_codecs` modules to the `gufe.serialization.json` subpackage.

This shouldn't be a breaking change unless users are rolling their own `JSONSerializerDeserializer` or using the custom codecs directly.

Checklist
* [x] Added a ``news`` entry

closes #516


gufe/serialization/json.py
```diff
@@ -204,3 +209,147 @@ def serializer(self, obj: Any) -> str:
...
+    return is_custom and dct["pint_unit_registry"] == "openff_units"
+
+
+def is_openff_quantity_dict(dct):
+    expected = ["pint_unit_registry", "magnitude", ":is_custom:", "unit"]
+    is_custom = all(exp in dct for exp in expected)
+    return is_custom and dct["pint_unit_registry"] == "openff_units"
+
+
+PATH_CODEC = JSONCodec(
```
IAlibay, line 269: If this is up for vote, I do generally like having the custom codes be in a separate file, both cleaner / more readable and makes it a lot less of a headache to find where things are.

gufe/serialization/json.py
```diff
@@ -204,3 +209,147 @@ def serializer(self, obj: Any) -> str:
     def deserializer(self, string: str) -> Any:
         """Callable to loads JSON"""
         return self._deserializer(string)
+
+
+# This code is part of OpenFE and is licensed under the MIT license.
+# For details, see https://github.com/OpenFreeEnergy/gufe
+# custom_codecs.py: A place to keep various custom JSONCodec instances
+
+import datetime
```
IAlibay, line 218: if we do keep this here, we probably shouldn't have imports half way through the file (and probably should change the header text above.

ianmkenney: Oops, yes of course. Fixing this now.

gufe/serialization/json.py
ianmkenney: It definitely is up for a vote. Will update once we hear from others!

dotsdl: I'm fine with the JSON components all being in one file. It gives parity with the `msgpack` components also all being in one place, and I'm not sure there's a whole lot of value in the separation we have now.

I often find myself entering the wrong file when I go looking for JSON things, then have to hop to the other one when I realize it.

dotsdl (approved):
Looks excellent @ianmkenney!

Just one question to resolve, then good to merge! I also took the liberty of renaming the test files to match the new layout!

gufe/serialization/json.py
```diff
@@ -1,12 +1,17 @@
-# This file was originally part of OpenPathSampling/SimStore, which is
-# distributed under the MIT license.
-# Portions Copyright (c) 2014-2022 the contributors to OpenPathSampling
-# Permissions are the same as those listed in the gufe LICENSE
```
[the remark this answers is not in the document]
ianmkenney: Yes, the copyright notice needs to be included with standard MIT license. Not sure of the best way to differentiate the portions of the file it applies to after the merge.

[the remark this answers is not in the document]
mikemhenry: If it is contiguous, you can use the line numbers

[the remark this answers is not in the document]
ianmkenney: I think it might be easier to just include the original comment on the top of `json.py` we are appending to the file, which doesn't invalidate the message, but might give more attribution to work done outside of gufe. I think micromanaging line numbers/specific functions will be tedious in the long run. I'll commit it just to wrap this up, but revert/rewrite if you think it's a bad solution.

gufe/serialization/json.py
atravitz: we'll keep it for now given the vote, but if it proves to be a headache we can always refactor.

PR #533, opened 2025-04-21
Unpin off units

IAlibay:
Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,6 @@ dependencies:
   - openff-toolkit >=0.15.1
   - openff-units >=0.3.0
   - pint
-  - openff-models >=0.0.5
   - pip
   - pydantic >1
   - zstandard
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - coverage
   - networkx
   - numpy
-  - openmm <8.2
+  - openmm ==8.2
   - openff-toolkit >=0.15.1
   - openff-units >=0.3.0
   - pint
--- a/environment.yml
+++ b/environment.yml
@@ -1,6 +1,5 @@
 name: gufe
 channels:
-  - jaimergp/label/unsupported-cudatoolkit-shim
   - conda-forge
 dependencies:
   - coverage
@@ -8,8 +7,8 @@ dependencies:
   - numpy
   - openmm <8.2
   - openff-toolkit >=0.15.1
-  - openff-units ==0.2.0  # https://github.com/openforcefield/openff-units/issues/69
-  - pint <0.22  # https://github.com/openforcefield/openff-units/issues/69
+  - openff-units >=0.3.0
+  - pint
   - openff-models >=0.0.5
   - pip
   - pydantic >1
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
   - coverage
   - networkx
   - numpy
-  - openmm ==8.2
+  - openmm <8.2
   - openff-toolkit >=0.15.1
   - openff-units >=0.3.0
   - pint
--- a/gufe/components/solventcomponent.py
+++ b/gufe/components/solventcomponent.py
@@ -2,7 +2,7 @@
 # For details, see https://github.com/OpenFreeEnergy/gufe
 from __future__ import annotations
 
-from typing import Optional, Tuple, TYPE_CHECKING
+from typing import Optional, Tuple
 
 from openff.units import unit, Quantity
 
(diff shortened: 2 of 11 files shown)


environment.yml
```diff
@@ -1,15 +1,14 @@
 name: gufe
 channels:
-  - jaimergp/label/unsupported-cudatoolkit-shim
```
IAlibay, line 3: This no longer does anything.

gufe/components/solventcomponent.py
```diff
@@ -5,6 +5,7 @@
 from typing import Optional, Tuple
 
 from openff.units import unit
+from pint import Quantity
```
IAlibay, line 8: Looks like this is the way (someone double check me), we'll need to switch the imports downstream when we do this.

IAlibay:
@dotsdl I would like to get this into v1.4 if possible. It will help up bump up the rest of the ecosystem.

mikemhenry (approved):
LGTM

gufe/components/solventcomponent.py
dotsdl: I don't think this is necessary at all. It looks like the line above for defining `unit` gives access to `unit.Quantity` for `openff-units` 0.3.0. Mixing imports from `pint` (even if this is literally all `unit.Quantity` is) is confusing at best to me.

What's the motivation?

gufe/components/solventcomponent.py
```diff
@@ -50,7 +51,7 @@ def __init__(
         neutralize : bool, optional
           if the net charge on the chemical state is neutralized by the ions in
           this solvent component.  Default `True`
-        ion_concentration : openff-units.unit.Quantity, optional
+        ion_concentration : openff-units.units.Quantity, optional
```
dotsdl, line 54:
```suggestion
        ion_concentration : openff.units.unit.Quantity, optional
```

dotsdl (changes requested):
See comments; I don't understand why this PR is necessary to support `openff-units` 0.3.0 and above.

gufe/components/solventcomponent.py
IAlibay: This doesn't work - mypy complains when you do that.

IAlibay:
I'm not sure why you say the PR is unnecessary @dotsdl :/

gufe/components/solventcomponent.py
IAlibay: (the above comment is about importing quantity directly from OpenFF.units.units). Effectively as far as I can tell there's a problem with the way Quantity is being typed in that file (since it's not a direct import of just pint.Quantity) which throws a type checking when mypy compares magnitudes.

We could import the `_Quantity` from that file though, which would be effectively like importing Quantity from pint.

mikemhenry:
This is the only spot where we import pint and use it directly,
https://github.com/OpenFreeEnergy/gufe/blob/main/gufe/serialization/msgpack.py#L8
maybe that is the source of the mypy issues you are seeing @IAlibay

gufe/components/solventcomponent.py
mattwthompson: Recommend `from openff.units import Quantity` (or `from openff.toolkit import Quantity` where it's installed)

IAlibay: @mattwthompson importing Quantity from `openff.units` is causing weird behaviour with mypy (see: https://github.com/OpenFreeEnergy/gufe/actions/runs/14599580786/job/40954126788?pr=533). Do you happen to know why that might be?

mikemhenry:
mypy errors:
```
gufe/serialization/msgpack.py:49: error: "Quantity" has no attribute "u"  [attr-defined]
gufe/components/solventcomponent.py:81: error: Unsupported operand types for > ("int" and "Callable[[], float | int | ndarray[Any, Any]]")  [operator]
```

mikemhenry:
If we don't want this to block, lets do what makes the most sense and then what a `# noqa` or whatever to tell mypy to skip the offending lines

IAlibay:
I'm meeting with @mattwthompson tomorrow to discuss this. Given that this is going to be an issue everywhere downstream, I'm not sure that ignoring the problem here is a good idea. At the very least there seems to be an issue upstream and we probably should take the opportunity to report it.

IAlibay:
Ok thanks to @mattwthompson we fixed the initial two issues here: https://github.com/openforcefield/openff-units/pull/117

Unfortunately the new issues that appeared since we pulled openff.models aren't as simple a fix.

IAlibay:
Honestly rather lost on what's going on with the typing now that we've vendored these files @mikemhenry :/

IAlibay:
pre-commit.ci autofix

IAlibay:
@atravitz @mikemhenry @dotsdl 

This should be good to go. The two remaining mypy errors have been fixed upstream and will work once there's a new release of `openff-units`.

gufe/vendor/openff/models/types.py
```diff
@@ -23,7 +23,7 @@ def __getitem__(self, t):
 
 
 if TYPE_CHECKING:
-    FloatQuantity = Quantity
+    FloatQuantity = unit.Quantity
```
IAlibay, line 26: Having to do this is weird typing magic. It's not a great solution but it'll do for now?

mattwthompson:
I made the 0.3.1 release (and merged the feedstock PR). Let me know if you run into any problems

PR #553, opened 2025-05-12
Move data to zenodo

mikemhenry:
Host the data on zenodo as we get 403 errors from github sometimes.

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/tests/conftest.py
+++ b/gufe/tests/conftest.py
@@ -17,38 +17,14 @@ from rdkit.Chem import AllChem
 import gufe
 from gufe.tests.test_protocol import DummyProtocol
 
-try:
-    urllib.request.urlopen("https://google.com")
-except URLError:
-    HAS_INTERNET = False
-else:
-    HAS_INTERNET = True
-
+# Support OpenMM as a soft dep
+# Set it to version 0.0.0 if it isn't installed
 try:
     from openmm import Platform
 
     OPENMM_VERSION = Version(Platform.getOpenMMVersion())
-# Support OpenMM as a soft dep
-# Set it to version 0.0.0 if it isn't installed
 except ModuleNotFoundError:
     OPENMM_VERSION = Version("0.0.0")
-
-
-class URLFileLike:
-    def __init__(self, url, encoding="utf-8"):
-        self.url = url
-        self.encoding = encoding
-        self.data = None
-
-    def __call__(self):
-        if not HAS_INTERNET:  # pragma: no-cover
-            pytest.skip("Skipping because internet seems faulty")
-
-        if self.data is None:
-            req = urllib.request.urlopen(self.url)
-            self.data = req.read().decode(self.encoding)
-
-        return io.StringIO(self.data)
 
 
 def get_test_filename(filename):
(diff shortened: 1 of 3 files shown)


gufe/tests/conftest.py
```diff
@@ -67,18 +69,51 @@ def get_test_filename(filename):
 ]
 
 
-_pl_benchmark_url_pattern = (
-    "https://github.com/OpenFreeEnergy/openfe-benchmarks/blob/main/openfe_benchmarks/data/{name}.pdb?raw=true"
+_pl_benchmark_url_pattern = "https://github.com/OpenFreeEnergy/openfe-benchmarks/tree/f577a88d94b6deae1c0de7eb926edf48dd42b72d/openfe_benchmarks/data/{name}.pdb?raw=true"
```
atravitz, line 72: we don't need this anymore, correct?

gufe/tests/conftest.py
```diff
@@ -67,18 +69,51 @@ def get_test_filename(filename):
...
+        "cmet_protein.pdb": "md5:0a38968fe9c09a49da2f1c57e84196f0",
+        "mcl1_protein.pdb": "md5:a2e57c14a925ee0dd9f158eb7ad5f413",
+        "p38_protein.pdb": "md5:52e1ba80c73c7f710219580bddb71de3",
+        "ptp1b_protein.pdb": "md5:f754c3aa3ea1617d99a4dd21fb158967",
+        "syk_protein.pdb": "md5:0c631f4b88ed7bbd35ee660214426cd2",
+        "thrombin_protein.pdb": "md5:261b8f040b389188e8c0cf14fbab5775",
+        "tnsk2_protein.pdb": "md5:aa13bef540d061ed66ef4240886a39ec",
+        "tyk2_protein.pdb": "md5:48d447290ee637ce8e3255cfa572297a",
+    },
+    retry_if_failed=10,
```
atravitz, line 92: 10 seems aggressive for # of retries - is this so that it goes up to the max of 10s between tries? https://www.fatiando.org/pooch/latest/api/generated/pooch.create.html

gufe/tests/conftest.py
```diff
@@ -67,18 +69,51 @@ def get_test_filename(filename):
...
 )
 
 
-PDB_BENCHMARK_LOADERS = {
-    name: URLFileLike(url=_pl_benchmark_url_pattern.format(name=name)) for name in _benchmark_pdb_names
-}
+# TODO We can either fixtureize this or make it lazy to speed things up
```
atravitz, line 96: I agree this could be improved, but it's probably fine for now. My priority is the current 403 errors.

atravitz (approved):
just a few comments

gufe/tests/conftest.py
mikemhenry: Since I didn't add any locking, we get contention sometimes when 5 pytest instances are trying to download the same file, with my local testing, this amount of retries worked, and with lower amounts it would still fail.

gufe/tests/conftest.py
mikemhenry: Correct -- We should either comment this line out OR remove all the bits that downloaded from github. Do you have a preference? I don't like to leave commented code around since it is confusing, but this also is testing code so I tried to explain it here:
https://github.com/OpenFreeEnergy/gufe/pull/553/files#diff-ef3697965a0e9abdb11db6defdf930637dbc635bf5ac13f9b022f7a88f536099R75-R76
and commented out here:
https://github.com/OpenFreeEnergy/gufe/pull/553/files#diff-ef3697965a0e9abdb11db6defdf930637dbc635bf5ac13f9b022f7a88f536099R110-R112
I am also happy to rip it all out as well

gufe/tests/conftest.py
atravitz: works for me then, thanks!

mikemhenry: I'll add a code comment mentioning this so that in the future it will be obvious

atravitz (approved):

PR #564, opened 2025-06-13
move component properties from AtomMapping up to ComponentMapping

atravitz:
It seems like these properties are intrinsic to any ComponentMapping object. In other words, we always will have `componentA` and `componentB ` asproperties of a `ComponentMapping`, since they must exist by definition.

Tips

Checklist
* [ ] Added a ``news`` entry


gufe/mapping/componentmapping.py
```diff
@@ -21,3 +21,13 @@ def __init__(self, componentA: gufe.Component, componentB: gufe.Component):
 
     def __contains__(self, item: gufe.Component):
         return item == self._componentA or item == self._componentB
+
+    @property
+    def componentA(self) -> gufe.Component:
+        """A copy of the first Component in the mapping"""
```
IAlibay, line 27: It's not really a copy though right? It's just giving you the object itself.

atravitz: True! fixed.

dotsdl:
Looks great @atravitz!

Issue #579, opened 2025-07-02
How to handle logging in a protocol


jthorton:
When writing a protocol how should logging be handled? In the SepTop PR I noticed a pattern where we repeatedly check if we are verbose before logging[ like so](https://github.com/OpenFreeEnergy/openfe/pull/1057#discussion_r2166365090) which seems awkward and error prone, as we could forget to check before logging. Would it not make sense to allow protocol authors to log anything and everything they want and then control the logging level at the execution of the protocol, taking away the decision from the author of what should count as verbose.

atravitz:
@dotsdl is this something you can address in https://github.com/OpenFreeEnergy/gufe/issues/520?

PR #594, opened 2025-07-23
Added Protocol how-to doc

### LINKED ISSUE #520: Add how-to doc for defining a new `Protocol`
As one of the extensible points (#499) in `gufe`, we should include a how-to doc on how to define entirely new `Protocol`s. This should highlight:
- subclassing `Protocol` and `ProtocolResult`
- defining custom `ProtocolUnit` subclasses
- practical advice for how to structure `Protocol.create()` to yield a `ProtocolDAG` that is parallelizable
- interfaces between `Component`s that are useful for `Protocol`s, such as methods for going to and from `openmm`, `openff`, and `rdkit` objects

This should reference the corresponding concept doc on `Protocol`s, which covers what `Protocol`s do and how they are structured.

dotsdl:
Closes #520.

- **Seeded protocol how-to**
- **Finished document, with substantial help from Claude**

### DIFF
--- a/docs/how-tos/protocol.rst
+++ b/docs/how-tos/protocol.rst
@@ -6,84 +6,532 @@ How to define a new ``Protocol``
 The **gufe** :ref:`Protocols <protocol>` system is designed as an *extensible point* of the library,
 allowing developers to write their own methods for performing free energy calculations in a form that can take full advantage of the OpenFE ecosystem.
 
-Our recommendation in this how-to is to start simple; we will call out areas where you can iterate further on making your ``Protocol`` more sophisticated.
+Our recommendation in this how-to is to start simple;
+we will call out areas where you can iterate further on making your ``Protocol`` more sophisticated.
 
 .. image:: ../_static/gufe_protocol_diagram.svg
     :alt: The ``gufe`` protocol system.
 
 
-Designing and defining your ``ProtocolDAG``
--------------------------------------------
+Overview: What makes a ``Protocol``
+-----------------------------------
 
-First, subclass :class:`.Protocol`, and create the :meth:`.Protocol._create` method:
+A complete ``Protocol`` implementation requires several interconnected components:
+
+1. **Settings class**: A Pydantic model defining configuration parameters
+2. **Protocol class**: The main class that creates computational workflows
+3. **ProtocolUnit classes**: Individual computational steps in the workflow
+4. **ProtocolResult class**: Container for aggregated results from multiple runs
+
+The ``Protocol`` creates a :ref:`ProtocolDAG <protocoldag>` (directed acyclic graph) of :ref:`ProtocolUnit <protocolunit>` objects that define the computational workflow.
+Multiple ``ProtocolDAG`` executions can be aggregated into a single :ref:`ProtocolResult <protocolresult>` to provide statistical estimates.
+
+
+Step 1: Define your Settings
+-----------------------------
+
+First, create a settings class that inherits from :class:`.Settings`.
+This defines all the configuration parameters your protocol needs:
 
 .. code-block:: python
 
-    from gufe import ChemicalSystem, Protocol, ComponentMapping, ProtocolDAGResult, ProtocolUnit
-    
-    class MyProtocol(Protocol):
+    from gufe.settings import Settings
+    import pydantic
+    from openff.units import unit
 
+    class MyProtocolSettings(Settings):
+        # Required: number of independent simulations
+        n_repeats: int = 5
+        
+        # Protocol-specific parameters
+        simulation_length: pydantic.Field(
+            default=10.0 * unit.nanosecond,
+            description="Length of each simulation"
+        )
+        
+        equilibration_length: pydantic.Field(
+            default=1.0 * unit.nanosecond, 
+            description="Equilibration time before production"
+        )
+
+
+Step 2: Define your ProtocolResult
+----------------------------------
+
+Create a result class that knows how to compute estimates and uncertainties from your protocol's outputs:
+
+.. code-block:: python
+
+    from gufe import ProtocolResult
+    from openff.units import unit
+    import numpy as np
+
+    class MyProtocolResult(ProtocolResult):
+        
+        def get_estimate(self) -> unit.Quantity:
+            """Calculate the free energy estimate from all runs."""
+            # Extract the key results from all completed runs
+            free_energies = []
+            for run_result in self.data["free_energies"]:
+                free_energies.append(run_result)
+            
+            # Return the mean as our best estimate
+            mean_dg = np.mean(free_energies) * unit.kilocalorie_per_mole
+            return mean_dg
+        
+        def get_uncertainty(self) -> unit.Quantity:
+            """Calculate the uncertainty from all runs."""
+            free_energies = self.data["free_energies"]
+            if len(free_energies) < 2:
+                return 0.0 * unit.kilocalorie_per_mole
+            
+            std_err = np.std(free_energies) / np.sqrt(len(free_energies))
+            return std_err * unit.kilocalorie_per_mole
+
+
+Step 3: Define your ProtocolUnits
+----------------------------------
+
+Create the computational units that will perform the actual work.
+Each unit should inherit from :class:`.ProtocolUnit` and implement a ``_execute`` method.
+
+.. important :: 
+
+   Use ``ctx.shared`` for large objects that need to be passed between units.
+   This avoids serialization issues and improves performance by keeping file paths in the return objects instead of the large objects themselves.
+
+.. code-block:: python
+
+    from gufe import ProtocolUnit
+
+    class SetupUnit(ProtocolUnit):
+        """Prepare the system for simulation."""
+        
+        @staticmethod
+        def _execute(ctx, *, stateA, stateB, mapping, settings, **inputs):
+            """Set up the alchemical system."""
+            import pickle
+            from pathlib import Path
+            
+            # ctx provides scratch and shared directories
+            # Use ctx.shared to write files that other units will need
+            shared_dir = Path(ctx.shared)
+            
+            # Your setup logic here...
+            # - Create alchemical system from stateA/stateB
+            # - Apply the atom mapping 
+            # - Set up force field parameters
+            prepared_system = ...  # Your setup code here
+            topology = ...         # Your topology creation
+            coordinates = ...      # Your coordinate preparation
+            
+            # Write large objects to shared directory instead of returning them
+            system_file = shared_dir / "system.pkl"
+            topology_file = shared_dir / "topology.pkl" 
+            coords_file = shared_dir / "initial_coords.pkl"
+            
+            with open(system_file, 'wb') as f:
+                pickle.dump(prepared_system, f)
+            with open(topology_file, 'wb') as f:
+                pickle.dump(topology, f)
+            with open(coords_file, 'wb') as f:
+                pickle.dump(coordinates, f)
+            
+            return {
+                "system_file": str(system_file),
+                "topology_file": str(topology_file),
+                "initial_coordinates_file": str(coords_file),
+                "log": "System setup completed"
+            }
+
+    class SimulationUnit(ProtocolUnit):
+        """Run an individual simulation."""
+        
+        @staticmethod 
+        def _execute(ctx, *, setup_result, window, settings, **inputs):
+            """Execute a single alchemical window simulation."""
+            import pickle
+            from pathlib import Path
+            
+            # Load large objects from files written by setup unit
+            with open(setup_result.outputs["system_file"], 'rb') as f:
+                system = pickle.load(f)
+            with open(setup_result.outputs["topology_file"], 'rb') as f:
+                topology = pickle.load(f)
+            with open(setup_result.outputs["initial_coordinates_file"], 'rb') as f:
+                coordinates = pickle.load(f)
+            
+            # Your simulation logic here...
+            # - Run equilibration
+            # - Run production simulation
+            # - Calculate free energy contribution
+            dg_contribution = ...  # Your free energy calculation
+            final_coords = ...     # Your final coordinates
+            
+            # Write output files to shared directory
+            shared_dir = Path(ctx.shared)
+            final_coords_file = shared_dir / f"final_coords_window_{window}.pkl"
+            
+            with open(final_coords_file, 'wb') as f:
+                pickle.dump(final_coords, f)
+            
+            return {
+                "free_energy_contribution": dg_contribution,
+                "final_coordinates_file": str(final_coords_file),
+                "window": window,
+                "log": f"Simulation completed for window {window}"
+            }
+
+    class AnalysisUnit(ProtocolUnit):
+        """Analyze results from all simulations."""
+        
+        @staticmethod
+        def _execute(ctx, *, simulation_results, settings, **inputs):
+            """Combine results from all simulation windows."""
+            import pickle
+            from pathlib import Path
+            
+            # simulation_results will be a list of ProtocolUnitResult objects
+            total_free_energy = 0.0
+            logs = []
+            final_coords = {}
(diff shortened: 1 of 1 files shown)


docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+we will call out areas where you can iterate further on making your ``Protocol`` more sophisticated.
+
+.. image:: ../_static/gufe_protocol_diagram.svg
+    :alt: The ``gufe`` protocol system.
+
+
+Overview: What makes a ``Protocol``
+-----------------------------------
+
+A complete ``Protocol`` implementation requires several interconnected components:
```
atravitz, line 19:
```suggestion
A complete ``Protocol`` implementation requires the following interconnected components:
```

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+
+1. **Settings class**: A Pydantic model defining configuration parameters
+2. **Protocol class**: The main class that creates computational workflows
+3. **ProtocolUnit classes**: Individual computational steps in the workflow
+4. **ProtocolResult class**: Container for aggregated results from multiple runs
```
atravitz, line 24: would it make sense to mirror the order below and link the sections?

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+.. image:: ../_static/gufe_protocol_diagram.svg
+    :alt: The ``gufe`` protocol system.
+
+
+Overview: What makes a ``Protocol``
+-----------------------------------
+
+A complete ``Protocol`` implementation requires several interconnected components:
+
+1. **Settings class**: A Pydantic model defining configuration parameters
```
atravitz, line 21:
```suggestion
1. **Settings class**: A `Pydantic <https://docs.pydantic.dev/latest/>`_ model defining configuration parameters
```

atravitz: a user probably is familiar with pydantic, but I think adding a link on first mention of an external library is good practice.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+        equilibration_length: pydantic.Field(
+            default=1.0 * unit.nanosecond, 
+            description="Equilibration time before production"
+        )
+
+
+Step 2: Define your ProtocolResult
+----------------------------------
+
+Create a result class that knows how to compute estimates and uncertainties from your protocol's outputs:
```
atravitz, line 61:
```suggestion
Create a result class that inherits from :ref:`ProtocolResult <protocolresult>`and defines how to compute estimates and uncertainties from your protocol's outputs:
```

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+        # Protocol-specific parameters
+        simulation_length: pydantic.Field(
+            default=10.0 * unit.nanosecond,
+            description="Length of each simulation"
+        )
+        
+        equilibration_length: pydantic.Field(
+            default=1.0 * unit.nanosecond, 
+            description="Equilibration time before production"
+        )
```
atravitz, line 55: are `n_repeats`, `simulation_length`, and `equilibration length` required, or is this just an example of the type of extension one might want to do?

docs/how-tos/protocol.rst
atravitz: I think it's helpful to explicitly link here because it shows that both of these methods must be defined.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+            
+            std_err = np.std(free_energies) / np.sqrt(len(free_energies))
+            return std_err * unit.kilocalorie_per_mole
+
+
+Step 3: Define your ProtocolUnits
+----------------------------------
+
+Create the computational units that will perform the actual work.
+Each unit should inherit from :class:`.ProtocolUnit` and implement a ``_execute`` method.
```
atravitz, line 96:
```suggestion
Each unit should inherit from :ref:`.ProtocolUnit <protocolunit>` and implement an ``_execute`` method.
```

I think I prefer to link to the extensible points doc, since you can get from extensible points -> python api easily, but not vice-versa.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+            shared_dir = Path(ctx.shared)
+            
+            # Your setup logic here...
+            # - Create alchemical system from stateA/stateB
+            # - Apply the atom mapping 
+            # - Set up force field parameters
+            prepared_system = ...  # Your setup code here
+            topology = ...         # Your topology creation
+            coordinates = ...      # Your coordinate preparation
+
```
atravitz, line 127: maybe i'm misreading, but this is a bit confusing to me. this docstring says to set up force field parameters, but where does that happen?

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+                "total_free_energy": total_free_energy,
+                "analysis_file": str(analysis_file),
+                "log": "Analysis completed"
+            }
+
+
+Step 4: Implement your Protocol class
+-------------------------------------
+
+Now create your main ``Protocol`` class that ties everything together:
```
atravitz, line 234:
```suggestion
Now create your main custom ``Protocol`` class that inherits from :ref`Protocol <protocol>` and ties everything together:
```

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+-----------------------
+
+1. **Start simple**: Begin with a minimal working implementation and add complexity gradually.
+
+2. **Handle errors gracefully**: Use ``try``/``except`` in ``_execute`` methods and return meaningful error information.
+
+3. **Use the context effectively**: The ``ctx`` parameter provides ``scratch`` (temporary, persists over execution of a single ``ProtocolUnit``) and ``shared`` (persists over execution of the ``ProtocolDAG``) directories.
+   Use ``ctx.shared`` for large objects that need to pass between units; store file paths in return objects, not the objects themselves.
+
+4. **Test thoroughly**: Write unit tests for your ``ProtocolUnit`` classes independently.
```
atravitz, line 491:
```suggestion
4. **Test thoroughly**: Write unit tests for your ``ProtocolUnit`` classes early in development.
```

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,536 @@
...
+
+5. **Document your settings**: Use Pydantic's ``Field`` with descriptions to document what each setting does.
```
atravitz, line 493:
```suggestion
5. **Document your settings**: Use Pydantic's `Field() function <https://docs.pydantic.dev/latest/concepts/fields/>`_ with descriptions to document what each setting does.
```

atravitz (changes requested):
I took a first pass at reviewing this, but I think feedback from protocol developers (@IAlibay, @ijpulidos, and/or @hannahbaumann) would be valuable.

docs/how-tos/protocol.rst
dotsdl: How all this happens is going to be very dependent on the tools and methods one is using. I think this is about as far as we can go with recommendations here.

dotsdl (commented):
Thanks for the review @atravitz! I believe I've addressed all your comments!

atravitz (approved):

atravitz:
thanks @dotsdl ! I'd still consider this blocked until at least one protocol developer reviews.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+    from openff.units import unit
+
+    class MyProtocolSettings(Settings):
+        """Settings for an example Protocol.
+
+        """
+        # the number of independent simulations to perform
+        n_repeats: int = 5
+        
+        simulation_length: pydantic.Field(
```
jthorton, line 52: I think we should add more details on the unit typing system we have here and introduce things like `FloatQuantity["nanosecond"]`, something like for settings which require units you can use FloatQuantity and provide a default unit which the value will be converted to this also ensures that the provided unit is compatible with what you intend.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+First, create a settings class that inherits from :class:`.Settings`.
+This defines all the configuration parameters your protocol needs:
+
+.. code-block:: python
+
+    from gufe.settings import Settings
+    import pydantic
+    from openff.units import unit
+
+    class MyProtocolSettings(Settings):
```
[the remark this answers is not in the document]
jthorton: Following this it might be good to introduce the idea of grouping like settings into separate classes, like we do in the other protocols, to make it easier for the user to find a setting or change a whole section of settings.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+    from openff.units import unit
+    import numpy as np
+
+    class MyProtocolResult(ProtocolResult):
+        
+        def get_estimate(self) -> unit.Quantity:
+            """Calculate the free energy estimate from all runs."""
+            # Extract the key results from all completed runs
+            free_energies = []
+            for run_result in self.data["free_energies"]:
```
jthorton, line 80: Why doesn't this already have a unit on it?

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+            with open(topology_file, 'wb') as f:
+                pickle.dump(topology, f)
+            with open(coords_file, 'wb') as f:
+                pickle.dump(coordinates, f)
+            
+            return {
+                "system_file": str(system_file),
+                "topology_file": str(topology_file),
+                "initial_coordinates_file": str(coords_file),
+                "log": "System setup completed"
```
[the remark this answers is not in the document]
jthorton: See this related issue and the example in SepTop https://github.com/OpenFreeEnergy/gufe/issues/579

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+----------------------------------
+
+You can add custom validation to check that inputs are compatible with your protocol:
+
+.. code-block:: python
+
+    class MyProtocol(Protocol):
+        # ... other methods ...
+        
+        def _validate(
```
jthorton, line 344: Also point to the default validate method as it might not be appropriate for this protocol or they might impliment redundant validation.

docs/how-tos/protocol.rst
atravitz: as discussed in gufe dev call, `FloatQuantity` will no longer be used as we move to vendoring interchange and using pydantic v2.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+        equilibration_length: pydantic.Field(
+            default=1.0 * unit.nanosecond, 
+            description="Equilibration time before production"
+        )
+
+
+Step 2: Define your ProtocolResult
+----------------------------------
+
+Create a result class that inherits from :ref:`ProtocolResult <protocolresult>` and defines how to compute estimates and uncertainties from your protocol's outputs:
```
hannahbaumann, line 66: Maybe define this more broadly, e.g. the ProtocolResult could also return file paths/... ; e.g. the MD protocol does not return estimates (returns None), but filepaths, could be other things.

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+            final_coords = ...     # Your final coordinates
+            
+            # Write output files to shared directory
+            shared_dir = Path(ctx.shared)
+            final_coords_file = shared_dir / f"final_coords_window_{window}.pkl"
+            
+            with open(final_coords_file, 'wb') as f:
+                pickle.dump(final_coords, f)
+            
+            return {
```
hannahbaumann, line 183: Potentially add explanation of ProtocolUnitResult?

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+                "total_free_energy": total_free_energy,
+                "analysis_file": str(analysis_file),
+                "log": "Analysis completed"
+            }
+
+
+Step 4: Implement your Protocol class
+-------------------------------------
+
+Now create your main custom ``Protocol`` class that inherits from :ref:`Protocol <protocol>` and ties everything together:
```
hannahbaumann, line 239: Maybe add explicitly what it "ties together" before showing it in the code below.

hannahbaumann (commented):
Thanks, this looks great, just left a few comments!

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
 
 How to define a new ``Protocol``
 ================================
+
+The **gufe** :ref:`Protocol <protocol>` system is designed as an *extensible point* of the library,
+allowing developers to write their own methods for performing free energy calculations in a form that can take full advantage of the OpenFE ecosystem.
```
IAlibay, line 7:
```suggestion
allowing developers to write their own methods for performing calculations in a form that can take full advantage of the OpenFE ecosystem.
```
Not all Protocols implement free energy methods - see MD protocols or Boltz-2, maybe this is explained elsewhere though.

docs/how-tos/protocol.rst
dotsdl: I've updated the `Settings` example to use the new typing system!

docs/how-tos/protocol.rst
dotsdl: Thanks for this! I've incorporated this into the notes within step 2!

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+
+.. code-block:: python
+
+    from gufe import ProtocolResult
+    from openff.units import unit
+    import numpy as np
+
+    class MyProtocolResult(ProtocolResult):
+        
+        def get_estimate(self) -> unit.Quantity:
```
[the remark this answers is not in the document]
dotsdl: I've added this as an inline comment!

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+1. **Start simple**: Begin with a minimal working implementation and add complexity gradually.
+
+2. **Handle errors gracefully**: Use ``try``/``except`` in ``_execute`` methods and return meaningful error information.
+
+3. **Use the context effectively**: The ``ctx`` parameter provides ``scratch`` (temporary, persists over execution of a single ``ProtocolUnit``) and ``shared`` (persists over execution of the ``ProtocolDAG``) directories.
+   Use ``ctx.shared`` for large objects that need to pass between units; store file paths in return objects, not the objects themselves.
+
+4. **Test thoroughly**: Write unit tests for your ``ProtocolUnit`` classes early in development.
+
+5. **Document your settings**: Use Pydantic's `Field() function <https://docs.pydantic.dev/latest/concepts/fields/>`_ with descriptions to document what each setting does.
```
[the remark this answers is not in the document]
atravitz: autodoc pydantic also handles `Field()` - I think we can have a suggestion, but ultimately just encourage good documentation

docs/how-tos/protocol.rst
```diff
@@ -2,3 +2,541 @@
...
+                # Extract numerical results directly
+                total_free_energy += sim_result.outputs["free_energy_contribution"]
+                logs.append(sim_result.outputs["log"])
+                
+                # Load coordinate files if needed for analysis
+                window = sim_result.outputs["window"]
+                coords_file = sim_result.outputs["final_coordinates_file"]
+                with open(coords_file, 'rb') as f:
+                    coords = pickle.load(f)
+                final_coords[window] = coords
```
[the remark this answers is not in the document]
dotsdl: I've reworked this example to use lambda windows instead, and then combining reduced potentials from each window and feeding to MBAR in the `AnalysisUnit`. I think this is originally what Claude was going for, but agree it didn't exactly add up.

PR #595, opened 2025-07-23
add tests for settings models

atravitz:
ahead of migrating to pydantic v2, we should have some more thorough tests of the settings models.

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/tests/test_models.py
+++ b/gufe/tests/test_models.py
@@ -78,6 +78,7 @@ def test_json_round_trip(all_settings_path, tmp_path):
 
     assert settings == Settings.parse_raw(settings_from_file)
 
+
 def test_default_settings():
     my_settings = Settings.get_defaults()
     my_settings.thermo_settings.temperature = 298 * unit.kelvin
@@ -85,7 +86,7 @@ def test_default_settings():
     my_settings.schema_json(indent=2)
 
 
-class TestOpenMMSystemGeneratorFFSettings:
+class TestSettingsValidation:
     @pytest.mark.parametrize(
         "value,valid,expected",
         [
@@ -97,7 +98,7 @@ class TestOpenMMSystemGeneratorFFSettings:
             (None, True, None),
         ],
     )
-    def test_constraints_validation(self, value, valid, expected):
+    def test_openmmff_constraints(self, value, valid, expected):
         if valid:
             s = OpenMMSystemGeneratorFFSettings(constraints=value)
             assert s.constraints == expected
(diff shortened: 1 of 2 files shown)


atravitz:
pre-commit.ci autofix

gufe/tests/test_models.py
```diff
@@ -41,23 +69,135 @@ def test_default_settings():
...
-            _ = OpenMMSystemGeneratorFFSettings(constraints=value)
+class TestSettingsValidation:
+    @pytest.mark.parametrize(
+        "value,valid,expected",
+        [
+            ("parsnips", False, None),  # shouldn't be allowed
+            ("hbonds", True, "hbonds"),
+            ("hangles", True, "hangles"),
+            ("allbonds", True, "allbonds"),  # allowed options
+            ("HBonds", True, "HBonds"),  # check case insensitivity TODO: cast this to lower?
```
mikemhenry, line 80: Good question, I think b/c settings are checked at instantiation time and normally right at the start of any workflow, _I_ think we should be more strict about things like the case of `HBonds` and platform names `CUDA` even though we could decide on `.lower()` everything. I think for 2.0 we should decide on how much we coerce things versus reject them.

gufe/tests/test_models.py
```diff
@@ -41,23 +69,135 @@ def test_default_settings():
...
+            assert s.constraints == expected
+        else:
+            with pytest.raises(ValueError):
+                _ = OpenMMSystemGeneratorFFSettings(constraints=value)
+
+    @pytest.mark.parametrize(
+        "value,valid,expected",
+        [
+            (1.0 * unit.nanometer, True, 1.0 * unit.nanometer),
+            (1.0, True, 1.0 * unit.nanometer),  # should cast float to nanometer
```
mikemhenry, line 96: Another example of behavior we should change in 2.0, I don't think we should assume input units

gufe/tests/test_models.py
```diff
@@ -41,23 +69,135 @@ def test_default_settings():
...
+            ("1", True, 1.0),
+        ],
+    )
+    def test_thermo_redox(self, value, valid, expected):
+        if valid:
+            s = ThermoSettings(redox_potential=value)
+            assert s.redox_potential == expected
+        else:
+            with pytest.raises(ValueError):
+                _ = ThermoSettings(redox_potential=value)
```
mikemhenry, line 200: In a future PR, we should think about how we can re-factor this pattern into some testing method where we pass in the class name and setting as well so it will be easy to add more without having to copy this bit

mikemhenry (approved):
LGTM!

atravitz:
Thanks @mikemhenry! I agree with all these comments and we can address them in the 2.0 migration.

PR #608, opened 2025-09-03
Add `ChemicalSystem.isin` method

### LINKED ISSUE #566: ChemicalSystem "isin" method to check if a Component belongs to a ChemicalSystem
**Is your feature request related to a problem? Please describe.**

Sometimes it's useful to check if a Component belongs to a ChemicalSystem (for example is componentA or componentB of an AtomMapping belong to a given ChemicalSystem). An `isin` or simillar method that compares the Components would be useful.

**Describe alternatives you've considered**

We can just do an any call on the dictionary values.

jthorton:
Fixes #566 by implementing an isin method on a ChemicalSystem which can check if an instance of type of a Component are present in the system and optionally return any matches. 

Tips

Checklist
* [x] Added a ``news`` entry


jthorton:
pre-commit.ci autofix

gufe/chemicalsystem.py
```diff
@@ -126,6 +126,45 @@ def total_charge(self):
...
+        for comp in self._components.values():
+            if isinstance(item, Component):
+                if comp == item:
+                    matches.append(comp)
+            elif issubclass(item, Component):
+                if isinstance(comp, item):
+                    matches.append(comp)
+
+        if return_matches:
+            return len(matches) > 0, matches
```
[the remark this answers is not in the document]
atravitz: as discussed offline, I prefer this solution of returning an empty list if there are no matches.

gufe/chemicalsystem.py
```diff
@@ -126,6 +126,34 @@ def total_charge(self):
...
+        matches = []
+        for comp in self._components.values():
+            if isinstance(item, Component):
+                if comp == item:
+                    matches.append(comp)
+            elif issubclass(item, Component):
+                if isinstance(comp, item):
+                    matches.append(comp)
+
+        return matches
```
IAlibay, line 155: O I thought we were going to keep both the bool return and just matches if we asked for matches. 

If we're keeping matches, would `get` be better than `isin`? The latter usually makes you expect a bool by default.

jthorton: Ahh I thought based on what Alyssa wrote above, we wanted this. Yeah `get` might make more sense based on the return type we now have.

jthorton:
pre-commit.ci autofix

gufe/chemicalsystem.py
```diff
@@ -126,6 +126,54 @@ def total_charge(self):
...
+        item : Component or type of Component
+            The Component instance or class to check for.
+
+        Returns
+        -------
+        bool
+            Returns True if the item is found in the ChemicalSystem, False otherwise.
+        """
+        # check the input types
+        if not (isinstance(item, Component) or (isinstance(item, type) and issubclass(item, Component))):
```
atravitz, line 143: Tests still pass if you remove the `(isinstance(item, type)` check, and off the top of my head I can't think of an edge case where you would need it. Could you either remove this or add a test for an applicable edge case?
```suggestion
        if not (isinstance(item, Component) or issubclass(item, Component))):
```

gufe/tests/test_chemicalsystem.py
```diff
@@ -132,6 +133,39 @@ def test_sorting(solvated_complex, solvated_ligand):
...
+
+def test_get_components_of_type(solvated_ligand):
+    # check for present types
+    matches = solvated_ligand.get_components_of_type(ProteinComponent)
+    assert matches == []
+
+    matches = solvated_ligand.get_components_of_type(SmallMoleculeComponent)
+    assert matches == [solvated_ligand.components["ligand"]]
+
+    matches = solvated_ligand.get_components_of_type(SolventComponent)
```
atravitz, line 160: I'd like the sanity-check that `solvated_ligand.get_components_of_type(Component)` returns all components.

news/add_contains_method.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Added method ``ChemicalSystem.contains()`` to check if a ``Component`` or ``Component`` type is present in the ``ChemicalSystem`` (`PR #608 <https://github.com/OpenFreeEnergy/gufe/pull/608>`_).
+
+* Added method ``ChemicalSystem.get_components_of_type()`` to return a list of ``Component``s that match the ``Component`` type in the ``ChemicalSystem`` (`PR #608 <https://github.com/OpenFreeEnergy/gufe/pull/608>`_).
```
atravitz, line 5: excellent news entry 🚀

gufe/chemicalsystem.py
```diff
@@ -126,6 +126,54 @@ def total_charge(self):
...
+
+        Parameters
+        ----------
+        item : Component or type of Component
+            The Component instance or class to check for.
+
+        Returns
+        -------
+        bool
+            Returns True if the item is found in the ChemicalSystem, False otherwise.
```
atravitz, line 140: not blocking, but bonus points for adding an `Examples` section in the spirit of gufe usability drive  (below also)

gufe/chemicalsystem.py
```diff
@@ -126,6 +126,54 @@ def total_charge(self):
...
+        Returns
+        -------
+        list of Component
+            A list of all Components in the ChemicalSystem that are instances of the specified type.
+            If no Components of the specified type are found, an empty list is returned.
+        """
+        if not (isinstance(item, type) and issubclass(item, Component)):
+            raise TypeError("`item` must be a subclass of `Component`")
+
+        matches = [comp for comp in self._components.values() if isinstance(comp, item)]
```
atravitz, line 173: Non-blocking thought:
Am I correct in thinking that all items will be unique here? I'm considering @ijpulidos's preference for sets (which tends to be my preference as well), but I don't think it makes much of a difference here. Plus, returning a list might make it clearer that we're *not* pulling just the unique items from a list that contains duplicates. 

Third idea just to make things harder: maybe a set would be appropriate here, since it suggests immutability?

atravitz (changes requested):
Looks great @jthorton! Just a couple things to address and some non-blocking thoughts. And optionally get @hannahbaumann and @ijpulidos feedback since they'll be users.

PR #645, opened 2025-10-14
user-facing clarity fixes for custom quantities

atravitz:
- **types.py -> typing.py**
- **updating docstrings**

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,11 @@ repos:
   - id: trailing-whitespace
     exclude: '\.(pdb|gro|top|sdf|xml|cif|graphml)$'
 
+- repo: https://github.com/tox-dev/pyproject-fmt
+  rev: "v2.8.0"
+  hooks:
+    - id: pyproject-fmt
+
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.13.3
   hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,36 +1,79 @@
 [build-system]
-requires = [
-	"setuptools>=77.0.3",
-	"versioningit",
-]
 build-backend = "setuptools.build_meta"
+requires = [
+  "setuptools>=77.0.3",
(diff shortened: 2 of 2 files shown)


IAlibay:
> > LGTM -- If we make this a separate package then it is really easy to add more types as needed without requiring a GUFE release.
> 
> My current solution is to put any openfe-specific reused custom types in openfe, and encourage protocol users to do the same since it's quite simple to do (docs for that are in-progress): [OpenFreeEnergy/openfe#1575](https://github.com/OpenFreeEnergy/openfe/pull/1575)

Aren't a lot of the things that end up being in openfe just "types other protocol users might want"? If so, I think it makes more sense to centralize them?

IAlibay:
> The problem with `TimeQuantity` is that earlier we decided that we want to convert any compatible unit to the unit specified by the type: https://github.com/OpenFreeEnergy/gufe/pull/584/files/096b537b2d06f3816c3c722a60203b02a472e0c8#r2237893007
> 
> switching from `NanosecondQuantity` to `TimeQuantity` would mean switching from converting `picosecond`->`nanosecond`, which would be an API break.
> 
> @IAlibay @mikemhenry do we need to have a conversation about this before v1.7.0?

Let's just stick with what we have now. Regarding the proliferation of types, at worst we just don't use NanosecondQuantity often. This relates to my point in the comment above - it's better if lots of options are exposed in gufe (or any other centralized typing place) and then those that don't get used eventually get removed.

mikemhenry:
We can talk about this in our meeting today but after thinking about this some more, I don't think we can ship this without adding a backwards compatibility shim in the `__init__.py` so people can still import this stuff from `types`. We can then add a deprecation warning that in 2.0 this shim will be removed.

I did do a quick search on github and it doesn't look like anyone (outside `openfe` forks) is using this BUT I still think we need to adhere to SemVer.

mikemhenry:
I was wrong! This was never released, sorry!

PR #669, opened 2025-10-29
expose GufeTokenizableTestsMixin

atravitz:
we should expose this to make it clear it's for users to take advantage of if they write custom GufeTokenizables.

Tips

Checklist
* [ ] Added a ``news`` entry


atravitz:
> In protocols we've just directly imported from the relevant module file, but this seems useful.

yeah, as we were writing the documentation, I felt that it's more intuitive to expose it here rather than encourage users to import from the module.

PR #687, opened 2025-12-03
Add `AlchemicalArchive`

### LINKED ISSUE #323: Define `AlchemicalArchive` object for use as archival artifact
In https://github.com/openforcefield/alchemiscale/issues/246, we identified the need to easily produce archival-quality extracts from `alchemiscale`. The proposed solution would package up `AlchemicalNetwork`s, `ProtocolDAGResult`s, and user metadata into an `AlchemicalArchive` object, which could itself be easily serialized into a form that is archivable on FAIR data repositories, such as Zenodo.

So that other tools in the OpenFE ecosystem can produce and process `AlchemicalArchive`s, it makes most sense to define this object in `gufe`. At its most basic, this object is effectively a `tuple` of the form:

```python
Tuple[AlchemicalNetwork, Dict[Transformation, List[ProtocolDAGResult]], Dict[str, Any]]
```
where the second element is a dict with `Transformation`s as keys, a list of `ProtocolDAGResult`s as values, and the third element a metadata dict that is user-defined.

ianmkenney:
This PR introduced the `AlchemicalArchive` for serializing an `AlchemicalNetwork` along with it's transformation results. Closes #323.


ianmkenney:
pre-commit.ci autofix

IAlibay (approved):

atravitz (changes requested):

atravitz (approved):

PR #708, opened 2026-01-15
Closes #583: ProtocolUnits list missing units in Protocol._create is ok?

### LINKED ISSUE #583: `ProtocolUnit`s list missing units in `Protocol._create` is ok?
**Describe the bug**
In the NonEquilibrium cycling protocol in feflow I've just encountered a discrepancy in what the return of the `._create` method is and what the actual units of the procotol are. While the [`_create` method isn't returning the `setup` unit](https://github.com/OpenFreeEnergy/feflow/blob/ed5b9edddbd45e952e008a85ae6dde8482261b31/feflow/protocols/nonequilibrium_cycling.py#L1037), the unit is definitely being successfully executed.

I wonder if this hints towards an unspecified and maybe undesired behavior in how we are building `ProtocolDAG`s.

**To Reproduce**
We have run the `NonEquilibriumCyclingProtocol` successfully with the current state of the code. I guess we could create a simple and short example of a hypothetical protocol that doesn't include one of its units to  try to reproduce this at a more basic level, if needed.

**Expectation**
Expectation is that lacking the actual setup unit from the contents of the returned list of `ProtocolUnit`s in the `_create` method should affect the behavior of the protocol, probably error out at some point. But it's being executed successfully.

dotsdl:
Closes #583.

Tips

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/gufe/protocols/__init__.py
+++ b/gufe/protocols/__init__.py
@@ -3,6 +3,7 @@
 from .errors import (
     GufeProtocolError,
     MissingUnitResultError,
+    ProtocolDAGError,
     ProtocolDAGResultError,
     ProtocolUnitExecutionError,
     ProtocolUnitFailureError,
--- a/gufe/protocols/errors.py
+++ b/gufe/protocols/errors.py
@@ -15,6 +15,10 @@ class ProtocolUnitExecutionError(GufeProtocolError):
     """Error when executing a protocol unit."""
 
 
+class ProtocolDAGError(GufeProtocolError):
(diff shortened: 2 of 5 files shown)


dotsdl:
pre-commit.ci autofix

atravitz (approved):

ijpulidos (approved):
This looks great. Just wanted to note that this addresses the problem from a protocol dev point of view, because the list of units returned by the create method would be fed into the constructor. I just left a non-blocking comment to consider.

PR #710, opened 2026-01-15
docs: add information on using keyed chain for serialization to disk

ethanholz:
This PR adds information about how you can leverage `to_keyed_chain()` as a way for serializing to disk.

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/docs/concepts/tokenizables.rst
+++ b/docs/concepts/tokenizables.rst
@@ -180,31 +180,21 @@ Deduplication on disk
 ~~~~~~~~~~~~~~~~~~~~~
 
 Deduplication when writing a GufeTokenizable to disk can be handled by :ref:`to_keyed_chain() <keyed_chain>`, which serializes and de-duplicates an entire chain of tokenizable objects to disk in a single file.
-However, **gufe** provides no tooling for deduplicating across chains stored to disk and it is up to the storage system to implement its own registry for handling this.
 
-The main idea is to use the ``GufeKey`` to ensure uniqueness, and to use it as a label for the object's serialized representation.
-Additionally, the ``GufeKey``, which is simply a string, can be used as a stand-in for the object.
-When an outer ``GufeTokenizable`` contains an inner ``GufeTokenizable``, the outer can store the ``GufeKey`` in place of the inner object.
-
-
-To convert a ``GufeTokenizable`` ``obj`` into a dictionary that references inner ``GufeTokenizable``\s by ``GufeKey``, use ``obj.to_keyed_dict()``.
-That method replaces each ``GufeTokenizable`` by a ``dict`` with a single key, ``':gufe-key:'``, mapping to the ``GufeKey`` of the object.
+**gufe** provides no tooling for deduplicating across chains stored to disk and it is up to the storage system to implement its own registry for handling this.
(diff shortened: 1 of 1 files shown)


atravitz:
thanks for this addition! requesting review from @dotsdl  and/or @dwhswenson  in case they have any wisdom to impart.

ethanholz:
From both sets of comments, I think we can note that using a `KeyedChain` to serialize directly to disk does work so long as you store the whole chain as a single file. I think that makes sense in a documentation context. Most users who serialize to disk would generally just dump a JSON file to disk. What we need to highlight is that to store to disk, you need something that is consistently serializable which in most cases is using something like `to_json`/`from_json` (or another format) which dumps everything to a repeatable format. 

I think I was too focused on how we have to handle this in Warehouse, rather than how users _should_ do this.

docs/concepts/tokenizables.rst
atravitz: I would open here with what you said in the PR comment

> , I think we can note that using a KeyedChain to serialize directly to disk does work so long as you store the whole chain as a single file.

And then say that de-duplicating _across_ files is out of scope of gufe, and requires the storage system to implement its own registry.

atravitz (changes requested):

docs/concepts/tokenizables.rst
```diff
@@ -179,8 +179,8 @@ In practice, that leads to the following behavior, where ``Foo()`` is representa
 Deduplication on disk
 ~~~~~~~~~~~~~~~~~~~~~
 
-Deduplication on disk storage is fundamentally the responsibility of the specific storage system, which falls outside the scope of **gufe**.
-However, **gufe** provides some tools to facilitate implementation of a storage system.
+Deduplication on on disk can be handled by :ref:`keyed chain <keyed_chain>` by serializing and deduplicating an entire chain of tokenizable objects to disk in a single file.
```
atravitz:
```suggestion
Deduplication when writing a GufeTokenizable to disk can be handled by :ref:`keyed chain <keyed_chain>`, which serializes and de-duplicates an entire chain of tokenizable objects to disk in a single file.
```

docs/concepts/tokenizables.rst
```diff
@@ -189,8 +189,19 @@ That is, we can store by reference to the ``GufeKey``.
...
+However, you take advantage of **gufe** functionality to facilitate system-specific deduplication.
 
+You can use :ref:`keyed dict <keyed_dict>` to represent``GufeTokenizable`` ``obj`` as a ``dict`` that references inner ``GufeTokenizable``\s by their``GufeKey``\s, and repeat this recursively to store (with the disk storage method of your choosing) all objects by reference.
+Then, to get a list of all objects, use :func:`.get_all_gufe_objs` on the outermost ``obj``\s.
+
+
+If you don't need the level of granularity that keyed dict representation offers, :ref:`keyed chain <keyed_chain>` does this recursive unpacking and handles the correct serialization of all nested objects.
+
+
+``obj.to_keyed_chain()`` does exactly that. It handles the correct serialization of all sub-objects.
```
atravitz:
```suggestion

```

Also delete since this is redundant now.

atravitz (changes requested):
couple more edits

atravitz (approved):

PR #712, opened 2026-01-21
docs: Context and Storage

ethanholz:
This PR closes #695, providing documentation on `ExternalStorage` and `Context`.

Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #718, opened 2026-01-27
feat: return storage handles from StorageManager

ethanholz:
This PR blocks #712 by solving a logical error in how we handle passing around storage handles between `ProtocolUnits`, previously users had no way to actually do this, now they do. The only thing that we may need to adjust is that this contains no metadata on which storage manager it is stored in (ie permanent vs shared), not sure if you really need it though.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/storage/storagemanager.py
+++ b/gufe/storage/storagemanager.py
@@ -78,7 +78,7 @@ class StorageManager:
         Returns
         -------
         str
-            The globaly namespaced path used by other units
+            The globally namespaced path to be used by other units
         """
         self.registry.add(filename)
         return self.append_to_namespace(self.namespace, filename)


gufe/storage/storagemanager.py
```diff
@@ -65,15 +67,21 @@ def append_to_namespace(namespace: str, filename: str) -> str:
...
 
         Parameters
         ----------
         filename : str
             The filename to register for transfer.
+
+        Returns
+        -------
+        str
+            The globaly namespaced path used by other units
```
atravitz:
```suggestion
            The globally namespaced path to be used by other units
```

atravitz (approved):

PR #741, opened 2026-02-19
Move the rest of kartograf viz bits into gufe migration viz epic

mikemhenry:
Tips

Checklist
* [ ] Added a ``news`` entry


mikemhenry:
I forgot I was using pre-commit locally so that is why we have 1 commit with a bad commit message

mikemhenry:
I've tested this downstream and it works https://github.com/OpenFreeEnergy/kartograf/pull/138

environment.yml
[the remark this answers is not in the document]
atravitz: I think a weekly runner for gufe-min isn't a bad idea - I'd like to catch any issues before it goes to conda forge. I'm imagining weird package resolution quirks.

gufe/tests/test_mapping_visualization.py
```diff
@@ -243,3 +245,25 @@ def test_draw_unhighlighted_molecule_integration_smoke(benzene_transforms):
     # integration test/smoke test to catch errors if the upstream drawing
     # code changes
     draw_unhighlighted_molecule(benzene_transforms["benzene"].to_rdkit())
+
+
+def test_score_mappings_rmsd(stereo_chem_mapping) -> None:
```
atravitz: nitpick - this means that `stereo_chem_mapping` will be created even if this test doesn't run, correct? Not the end of the world, but would be nice if we could avoid.

gufe/visualization/mapping_visualization.py
```diff
@@ -394,3 +404,161 @@ def _get_max_dist_in_x(atom_mapping) -> float:
...
     estm = float(np.round(max(max_d), 1))
     return estm if (estm > 5) else 5
+
+
+@requires_package("py3Dmol")
+def display_mapping_3d(
+    mapping: LigandAtomMapping,
+    spheres: bool | None = True,
+    show_atomIDs: bool | None = False,
+    style: str | None = "stick",
```
atravitz: do we need `| None` if default values _arent_ `None`?

atravitz (commented):
couple questions!

gufe/tests/test_mapping_visualization.py
atravitz: wonderful, thank you!  I saw "session" scoped and thought "this might take some time, let's not construct it unnecessarily"

atravitz (approved):
lgtm! I responded about docs style guide, but it's non-blocking.

gufe/visualization/mapping_visualization.py
mikemhenry: sweet, I thought the same thing!

PR #742, opened 2026-02-20
Remove imports into gufe namespace that are unnecessary 

mikemhenry:
Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/__init__.py
+++ b/gufe/__init__.py
@@ -3,7 +3,6 @@
 
 from importlib.metadata import version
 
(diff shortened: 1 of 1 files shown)


mikemhenry:
I don't think this needs a news since it wouldn't change how anyone would import these modules in their own code (user script or downstream package) BUT I could be wrong.

atravitz (approved):
I tested it with openfe `main` and I agree that this shouldn't affect imports. good to merge.

PR #753, opened 2026-02-27
Epic: context & storage

atravitz:
Changes to how we handle contexts and storage.

This introduces *breaking changes* that will require updates to openfe protocols and exorcist (as described in openfe #1833).

Whether we do this in 2.0 or a highly-advertised 1.x is TBD.

Implemented by @ethanholz - this epic is to track the work as a long-lived branch.

resolves https://github.com/OpenFreeEnergy/gufe/pull/753 

Tips

Checklist
* [ ] Added a ``news`` entry


docs/concepts/context.rst
```diff
@@ -0,0 +1,175 @@
...
+Existing protocols adopted a variety of implicit conventions around those attributes.
+Follow this checklist when migrating old protocols:
+
+1. **Swap file paths to StorageManager APIs.**  Calls like ``ctx.shared /
+   "filename"`` should be replaced with the helper methods offered by
+   :class:`StorageManager`.  For example:
+
+.. code-block:: python
+
+    path = ctx.shared.scratch_dir / "myfile.dat"
```
dotsdl, line 143:
```suggestion
    path = ctx.shared / "myfile.dat"
```

docs/concepts/context.rst
```diff
@@ -0,0 +1,175 @@
...
+2. **Avoid storing heavy objects in Python outputs.**  Older protocols often
+   returned raw ``Path`` objects pointing at scratch files.  Instead, register
+   the file with the storage manager and return the storage key (a string) from
+   ``_execute`` as shown in the example above.  Downstream units can then call
+   :meth:`.StorageManager.load`.
+
+3. **Handle ctx.permanent.**  There was no equivalent in the legacy API.
+   Decide which results must persist between DAG executions and write them via
+   ``ctx.permanent``.  For migration you can start by mirroring whatever used
+   to live in ``ctx.shared`` and refine later.
```
dotsdl, line 161: I don't think we should recommend that protocol authors switch from `shared` to `permanent` by default, as this could result in large in-DAG intermediate files getting held by execution engines for no good reason.

The recommendation here should be to switch to `permanent` any files that must survive DAG execution for the purposes of continuation or understanding results later; the default should not be *keep everything*.

dotsdl: To elaborate: in **alchemiscale**, we plan to retain `permanent` files by default (unless opted out of by the `Task` creator) for DAGs, and only retain `shared` during the duration of DAG execution (for continuation if interrupted) *or* retain `shared` indenfinitely if opted into by the `Task` creator (for debugging problematic cases).

dotsdl (commented):
This is looking excellent! Did a quick review of the new concept docs; just a couple notes I think should be addressed.

Will give the rest of the PR a thorough review at a later time! Can't wait to use this in `alchemiscale` to finally support result file retention!

PR #761, opened 2026-03-12
Add missing files to MANIFEST.in

mikemhenry:
Since we build on conda-forge with a tarball and not in a git repo, `setuptools_scm` doesn't know what files to include via tracking, and then falls back to default behavior, so we need to include these files in the `MANIFEST.in`. We should investigate if there are other ways to do this.

Tips

Checklist
* [ ] Added a ``news`` entry


atravitz (approved):
lgty!

PR #771, opened 2026-03-26
Run pyupgrade --py311-plus since python 3.11 is the min

mikemhenry:
Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/gufe/network.py
+++ b/gufe/network.py
@@ -2,9 +2,9 @@
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
 from collections.abc import Generator, Iterable
+from typing import Self  # Self is included in typing as of python 3.11
 
 import networkx as nx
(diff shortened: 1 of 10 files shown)


mikemhenry:
Need to make sure the docs build

mikemhenry:
If we do decide to merge this in, we should add the commit to `.git-blame-ignore-revs`

gufe/network.py
```diff
@@ -2,9 +2,9 @@
 # For details, see https://github.com/OpenFreeEnergy/gufe
 
 from collections.abc import Generator, Iterable
+from typing import Self  # Self is included in typing as of python 3.11
```
atravitz:
```suggestion
from typing import Self
```

atravitz (approved):
just one comment clean up, but yes I like this, and I agree it should be added to the ignore dotfile.

PR #789, opened 2026-04-24
Add metadata attribute to TransformationBase, Transformation, and NonTransformation

### LINKED ISSUE #788: Add a metadata field to transformations
**Is your feature request related to a problem? Please describe.**

I would like to add extra information to a transformation so that I can map it to experimental results more easily. In the past I have added annotations to the ligand atom mapping taken by a transformation for hybrid topology protocol runs, but some protocols, like SepTop don't need a mapping. This came up in https://github.com/OpenFreeEnergy/openfe-benchmarks/pull/115

**Describe the solution you'd like**

To add a metadata field to transformations which acts like the annotations field on a ligand atom mapping.

**Describe alternatives you've considered**

**Additional context**

ianmkenney:
Add metadata attribute to TransformationBase, Transformation, and NonTransformation. Fixes #788. 

Checklist
* [x] Added a ``news`` entry

- [X] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.


gufe/transformations/transformation.py
```diff
@@ -16,24 +18,25 @@
...
         Parameters
         ----------
-        protocol : Protocol
+        protocol: Protocol
             The sampling method to use for the transformation.
-        name : str, optional
+        name: str, optional
             A human-readable name for this transformation.
+        metadata : dict[str, Any], optional
+            Metadata of the Transformation. ``None`` produces and
```
atravitz:
```suggestion
            Metadata of the Transformation. ``None`` produces an
```

src/gufe/transformations/transformation.py
```diff
@@ -49,6 +52,11 @@ def name(self) -> str | None:
         """
         return self._name
 
+    @property
+    def metadata(self) -> dict[str, Any]:
+        """Metadata associated with the transformation."""
+        return copy.deepcopy(self._metadata)
```
atravitz, line 58: could this inadvertently undermine the flyweight pattern if someone puts a huge `GufeTokenizable` in `metadata` without understanding that it will be deepcopied? If so, I think this should be a note or warning in the docs at minimum.

atravitz (changes requested):

src/gufe/transformations/transformation.py
ianmkenney: Implementing `__deepcopy__` for `GufeTokenizable` might address this.

```python
from copy import deepcopy
from typing import Self

class Example:

    def __deepcopy__(self, memo) -> Self:
        return self


container = {"my object": Example()}
copied = deepcopy(container)

assert container["my object"] is copied["my object"]
assert container is not copied
```

ianmkenney: Added it in ef328f3, can revert if not needed

jthorton (approved):
This is perfect thanks @ianmkenney!

atravitz (approved):

PR #792, opened 2026-06-02
move to src layout

atravitz:
Tips

Checklist
* [ ] Added a ``news`` entry


.github/workflows/check_for_api_break.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       id: check
       run: |
         pip install griffe
-        griffe check "gufe" --verbose -a origin/main
+        griffe check "src/gufe" --verbose -a origin/main
```
mikemhenry: Looking at https://mkdocstrings.github.io/griffe/guide/users/checking/#using-git

I think this needs to be:

```suggestion
        griffe check -s "src/gufe" gufe --verbose -a origin/main
```

mikemhenry (changes requested):
We will either want to update the [docs](https://github.com/OpenFreeEnergy/gufe/blob/main/docs/conf.py#L16) with something like `sys.path.insert(0, os.path.abspath("../src"))` or just remove the line since we can just expect the user (we already do this on RTD) to install the package before installing the docs.

atravitz:
> We will either want to update the [docs](https://github.com/OpenFreeEnergy/gufe/blob/main/docs/conf.py#L16) with something like `sys.path.insert(0, os.path.abspath("../src"))` or just remove the line since we can just expect the user (we already do this on RTD) to install the package before installing the docs.

Removed!

mikemhenry (approved):
LGTY!

Issue #750, opened 2026-02-25
RDKIT Pickling Roundtrip Error (but only on MacOS?)


apayne97:
Thanks to @ijpulidos and @chrisiacovella for finding and troubleshooting this.

[2026.02.20_test_failure.ipynb](https://github.com/user-attachments/files/25549804/2026.02.20_test_failure.ipynb)

It manifests as the loss of the names in a LigandNetwork in the edges (but not the nodes?) such that `edge.componentA.name` and `edge.componentB.name` are ''. 

This warning appears: "UserWarning: RDKit does not preserve Mol properties when pickled by default, which may drop e.g. atom charges; consider setting `Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)`

When we set this property in a few different places in our code, it fixes the bug but only for ubuntu/linux. On MacOS, the warning and error still appear.

The drugforge error is here:
https://github.com/choderalab/drugforge/actions/runs/22386574178/job/64798431670?pr=132

And you can see it passes on Ubuntu here:
https://github.com/choderalab/drugforge/actions/runs/22386574178/job/64798431676?pr=132#step:8:138

On the other hand, if you run this code on MacOS, it works just fine(below)
```
from rdkit import Chem
from rdkit.Chem import Descriptors
import pickle
Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)

smiles = "CN1C=NC2=C1C(=O)N(C(=O)N2C)C"
mol = Chem.MolFromSmiles(smiles)

mol.SetProp("_Name", "Caffeine")
mol.SetProp("Source", "Coffee Beans")
mol.SetProp("MolecularWeight", f"{Descriptors.MolWt(mol):.2f}")

# Pickle the molecule
pickled_mol = pickle.dumps(mol)

print(f"Molecule '{mol.GetProp('_Name')}' has been pickled.")
print(f"Pickle size: {len(pickled_mol)} bytes")

# Unpickle to demonstrate data persistence
new_mol = pickle.loads(pickled_mol)
print(f"\nRecovered Name: {new_mol.GetProp('_Name')}")
print(f"Recovered MW: {new_mol.GetProp('MolecularWeight')}")
```
This suggests there's a way that `Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)` is not persisting through the code but only for MacOS.

[...]

PR #795, opened 2026-06-04
Make sure to save rdkit props when pickling happens

mikemhenry:
related to https://github.com/OpenFreeEnergy/gufe/issues/750

Tips

Checklist
* [ ] Added a ``news`` entry


gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -461,6 +462,39 @@ def test_prop_preservation(ethane, target, dtype):
...
+            lambda smc: SmallMoleculeComponent.from_dict(smc.to_dict()),
+            id="dict",
+        ),
+        pytest.param(
+            lambda smc: SmallMoleculeComponent.from_msgpack(content=smc.to_msgpack()),
+            id="msgpack",
+        ),
+    ],
+)
+def test_equality_after_round_trip(roundtrip):
```
atravitz: can this be generalized and live in [ExplicitMoleculeComponentMixin](https://github.com/OpenFreeEnergy/gufe/blob/8f5dfa03fe44e6d47ed7004d81149b2cd40986ae/gufe/tests/test_explicitmoleculecomponent.py#L4) so that ProteinComponents and any future subclasses are tested?

I'm noting the existing pickling test there, and that `assert smc is new_smc` will pass for `msgpack` and `dict` but *not* `pickle`, which might be something we need to include in docs.

mikemhenry:
I think two main things to talk about at our gufe meeting:

If anyone has been using pickle to save these components, then this will break loading them back in. I'm fine with this since they just need to use an older version of gufe to load them and then re-save them as a dict or message pack. Also we have never encouraged people to use pickle as a storage method so it isn't a use case we really support.

Another idea I had is, should this live up "higher" in our object inheritance hierarchy?

gufe/tests/test_smallmoleculecomponent.py
```diff
@@ -461,6 +462,39 @@ def test_prop_preservation(ethane, target, dtype):
...
+    Chem.AllChem.Compute2DCoords(mol)
+    mol.SetProp("_Name", "ethane")
+
+    smc = SmallMoleculeComponent(rdkit=mol)
+
+    new_smc = roundtrip(smc)
+
+    assert new_smc == smc
+
+    new_mol = new_smc.to_rdkit()
```
atravitz:
```suggestion
    new_mol = new_smc.to_rdkit()
    smc.to_rdkit() != new_smc.to_rdkit()
```

we discussed this and I think it's expected, but adding it in the test might be good to document this as expected behavior.

atravitz: offline @mikemhenry and I agreed this isn't necessary.

mikemhenry:
FYI this test does work to fix the issue on osx, see https://github.com/OpenFreeEnergy/openfe/pull/1861

mikemhenry:
So these tests will slow things down a bit with all the to/from
`24 passed, 916 deselected in 85.54s (0:01:25)`

I can combine the equality check + the name check into one test to speed things up

mikemhenry:
But first I want to see if things all pass after I re-worked the tests (and want to make sure other perts of the testing suite are unaffected).

mikemhenry:
Okay faster now as one test:
`12 passed, 916 deselected in 47.92s`

src/gufe/components/explicitmoleculecomponent.py
```diff
@@ -46,6 +46,7 @@ class ExplicitMoleculeComponent(Component):
     _name: str
 
     def __init__(self, rdkit: RDKitMol, name: str = ""):
+        Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
```
IAlibay, line 49: RDKit has changed a lot since 2019 - it might be good to just quickly ping the RDKit devs to check if the side-effects with this option really are just the two outlined in https://github.com/rdkit/rdkit/issues/2470

IAlibay: I'm wondering if it wouldn't be safer to temporarily set/unset (if not already set) in `__getstate__` / `__setstate__` ? That way "other" rdkit behaviour remains the way users expect it to be ?

mikemhenry: @IAlibay I did that originally but we couldn't come up with what that "other" behavior would be. I just commented on that issue to see if anything changed.

mikemhenry: Sounds like we are good!

> Both of those statements are true. There is one caveat, some properties will actually be “double pickled” with this setting. At some point we should remove this behavior and make property pickling the default and have it be opt out. I have always found the default behavior surprising

mikemhenry:
Superseded by #797

PR #577, opened 2025-06-19
[WIP] adding a try/except for pathlib._local import

atravitz:
Tips
* Comment "pre-commit.ci autofix" to have pre-commit.ci atomically format your PR.
  Since this will create a commit, it is best to make this comment when you are finished with your work.

Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [ ] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #796, opened 2026-06-05
Support Python 3.14

mikemhenry:
Tips

Checklist
* [ ] Added a ``news`` entry


mikemhenry:
Getting these errors now:
```
FAILED gufe/tests/test_models.py::TestBoxVectors::test_valid_box_quantity[value0-m_expected0] - pydantic_core._pydantic_core.ValidationError: 1 validation error for BoxSettingsModel
box_vectors
  Extra inputs are not permitted [type=extra_forbidden, input_value=<Quantity([[0. 0. 1.]
```
We are just missing `ConfigDict(extra='ignore')` in our dummy settings.

mikemhenry:
Ah our old friend:
```
FAILED gufe/tests/test_serialization_json.py::TestPathCodec::test_default - AssertionError: assert {':is_custom:...h': 'foo/bar'} == {':is_custom:...h': 'foo/bar'}
  
  Omitting 3 identical items, use -vv to show
  Differing items:
  {'__module__': 'pathlib'} != {'__module__': 'pathlib._local'}
  
  Full diff:
    {
        ':is_custom:': True,
        '__class__': 'PosixPath',
  -     '__module__': 'pathlib._local',
  ?                           -------
  +     '__module__': 'pathlib',
        'path': 'foo/bar',
    }
```

atravitz:
> Ah our old friend:
> 
> ```
> FAILED gufe/tests/test_serialization_json.py::TestPathCodec::test_default - AssertionError: assert {':is_custom:...h': 'foo/bar'} == {':is_custom:...h': 'foo/bar'}
>   
>   Omitting 3 identical items, use -vv to show
>   Differing items:
>   {'__module__': 'pathlib'} != {'__module__': 'pathlib._local'}
>   
>   Full diff:
>     {
>         ':is_custom:': True,
>         '__class__': 'PosixPath',
>   -     '__module__': 'pathlib._local',
>   ?                           -------
>   +     '__module__': 'pathlib',
>         'path': 'foo/bar',
>     }
> ```

First battle with this was here: https://github.com/OpenFreeEnergy/gufe/pull/577/changes

If I understand correctly, `pathlib._local` is only a python 3.13 thing? In which case we should be able to change [this line](https://github.com/OpenFreeEnergy/gufe/blob/e4ee296a7ffb9458a13113d9a6b20024380591b2/gufe/tests/test_serialization_json.py#L246) to `if sys.version_info[:2] == (3, 13):`

mikemhenry:
You are correct, we need to fix that test since this is the current situation:


Runtime writing JSON | observed module
-- | --
Python ≤3.12 | pathlib
Python 3.13 | pathlib._local
Python 3.14 | pathlib

For the test, we can just get the module we need at runtime:

```
self.dcts = [
    {
        "__class__": "PosixPath",
        "__module__": pathlib.PosixPath.__module__,
        ":is_custom:": True,
        "path": "foo/bar",
    }
]
```

We can also set up the [codec](https://github.com/OpenFreeEnergy/gufe/blob/e4ee296a7ffb9458a13113d9a6b20024380591b2/gufe/serialization/json.py#L273-L278) to just use the public module name, something like:

```
PATH_CODEC = JSONCodec(
    cls=pathlib.PosixPath,
    to_dict=lambda p: {
        "__module__": "pathlib",
        "__class__": "PosixPath",
        ":is_custom:": True,
        "path": str(p),
    },
    from_dict=lambda dct: pathlib.PosixPath(dct["path"]),
    is_my_dict=is_legacy_path_dict,
)
```

Then the reader will accept `("pathlib", "pathlib._local")` but we will always write `pathlib` even on python 3.13

atravitz:
> Then the reader will accept ("pathlib", "pathlib._local") but we will always write pathlib even on python 3.13

We originally decided not to do this because we anticipated `pathlib._local` to be the behavior for the future and for `pathlib` to be legacy, but that turned out not to be the case.

I'm typically inclined to keep the code as-is so that we're not arbitrarily changing things, but since the 3.13 behavior was immediately reverted in 3.14, I'm fine with updating the code to treat 3.13 as an anomaly that we're accounting for and update the codec.

tl;dr: yeah do it 🚀

mikemhenry:
Sweet, looking at the issues and mailing lists, it looks like it was a mistake that happened during a re-factor which is why 3.13 is an anomaly

tl;dr :rocket:

mikemhenry:
The API check bot is having an issue, will fix later `ImportError: With sys.path = ['/tmp/griffe-worktree-gufe-origin-main-e4757g4_/origin-main/src/gufe'], accessing 'gufe' raises ModuleNotFoundError: No module named 'gufe'`

pyproject.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT"
 license-files = [ "LICENSE" ]
 authors = [ { name = "The OpenFE developers", email = "openfe@omsf.io" } ]
-requires-python = ">=3.11,<3.14"
+requires-python = ">=3.11"
```
atravitz: do we want to argue about upper pins again? I'd rather pin until we check 3.15 beta, but I won't block this PR over it.

src/gufe/serialization/json.py
```diff
@@ -272,7 +272,12 @@ def is_legacy_path_dict(dct: dict) -> bool:
...
     cls=pathlib.PosixPath,
-    to_dict=lambda p: {"path": str(p)},
+    to_dict=lambda p: {
+        "__module__": "pathlib",
+        "__class__": "PosixPath",
+        ":is_custom:": True,
+        "path": str(p),
+    },
     from_dict=lambda dct: pathlib.PosixPath(dct["path"]),
     is_my_dict=is_legacy_path_dict,
```
atravitz: `is_legacy_path_dict` should be renamed to `is_python_313_path_dict` or similar.

src/gufe/serialization/json.py
```diff
@@ -272,7 +272,12 @@ def is_legacy_path_dict(dct: dict) -> bool:
 
 PATH_CODEC = JSONCodec(
     cls=pathlib.PosixPath,
-    to_dict=lambda p: {"path": str(p)},
+    to_dict=lambda p: {
+        "__module__": "pathlib",
+        "__class__": "PosixPath",
+        ":is_custom:": True,
```
atravitz: it's not clear to me why we need to add these:

```
 "__class__": "PosixPath",
 ":is_custom:": True,
```

src/gufe/settings/models.py
```diff
@@ -4,6 +4,8 @@
 Pydantic models used for storing settings.
 """
 
+from __future__ import annotations
```
atravitz, line 7: this makes our annotations evaluated eagerly? https://docs.python.org/3/whatsnew/3.14.html#new-features

atravitz (changes requested):
some questions

pyproject.toml
mikemhenry: Sure, I was working on that with https://github.com/conda-forge/python-feedstock/pull/878 but that might take some work so we can add `<3.15` (very excited for lazy imports!).

src/gufe/settings/models.py
mikemhenry: This makes our annotations act like they did in python `<=3.13` since they changed in 3.14. So this actually uses the "make annotations strings" for analysis.

src/gufe/serialization/json.py
mikemhenry: I'll check if we can drop them!

src/gufe/serialization/json.py
```diff
@@ -257,7 +257,7 @@ def is_openff_quantity_dict(dct):
     return is_custom and dct["pint_unit_registry"] == "openff_units"
 
 
-def is_legacy_path_dict(dct: dict) -> bool:
+def is_python_313_path_dict(dct: dict) -> bool:
     """This supports the case where python 3.12 needs to load python 3.13
     TODO: remove when 3.12 support is dropped
     """
```
atravitz:
```suggestion
    """This supports the python 3.13 convention
    TODO: remove when 3.13 support is dropped
    """
```

atravitz (approved):
just one docstring correction, then good to merge!

PR #800, opened 2026-06-22
turn off api break check (for now)

atravitz:
We're having some false positives from the runners failing, I'd like to turn this check off until we address it.

Tips

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/check_for_api_break.yaml
+++ b/.github/workflows/check_for_api_break.yaml
@@ -1,8 +1,7 @@
 name: Check for API breaks
 
 on:
-  pull_request_target:
-    branches:
+  workflow_dispatch:
 
 jobs:
   check:


IAlibay (approved):
As per other PR - maybe raise issue as a reminder to re-enable?

.github/workflows/check_for_api_break.yaml
mikemhenry, line 28:
```suggestion
        griffe check "gufe" -s src --verbose -a origin/main
```
This will match how we do it on openfe

mikemhenry: Not sure which is "correct" but I just noticed we are doing it differently. If you want to turn both off (since a false positive or a false negative here is bad) then that makes sense as well!

mikemhenry (approved):

.github/workflows/check_for_api_break.yaml
mikemhenry: I will play with this locally to try and fix it.

PR #794, opened 2026-06-02
[WIP] fixes for pydantic>2.12

atravitz:
following @mattwthompson's lead at: https://github.com/openforcefield/openff-interchange/pull/1501/changes

Tips
* Comment "pre-commit.ci autofix" to have pre-commit.ci atomically format your PR.
  Since this will create a commit, it is best to make this comment when you are finished with your work.

Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [ ] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

PR #808, opened 2026-06-30
support pydantic 2.13

### LINKED ISSUE #807: support pydantic >=2.13
see:
- https://github.com/openforcefield/openff-units/pull/178
- https://github.com/openforcefield/openff-interchange/pull/1501

atravitz:
resolves #807 

Note that this will push us to drop python 3.11, which is out of SPEC0.

Tips

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: no

## Checklist
* [ ] Added a ``news`` entry


atravitz:
currently, these are ways we _could_ fix things, but the existing `model_dump()` behavior is a problem IMO.

for example, in the test `src/gufe/tests/test_models.py(340)test_frozen_equality_changed()`

```
s1.model_dump()
  > {'forcefield_settings': {}, 'thermo_settings': {'temperature': {'val': 300, 'unit': 'kelvin'}, 'pressure': None, 'ph': None, 'redox_potential': None}}`
```

(note `'forcefield_settings':{}`)
vs 

```
s1.forcefield_settings.model_dump()
 > {'constraints': 'allbonds', 'rigid_water': True, 'hydrogen_mass': 3.0, 'forcefields': ['amber/ff14SB.xml', 'amber/tip3p_standard.xml', 'amber/tip3p_HFE_multivalent.xml', 'amber/phosaa10.xml', 'amber/lipid17_merged.xml'], 'small_molecule_forcefield': 'openff-2.2.1', 'nonbonded_method': 'PME', 'nonbonded_cutoff': {'val': 0.9, 'unit': 'nanometer'}}
```

I think we should be able to resolve this in model settings or something similarly upstream, but I haven't sorted it out yet.

atravitz:
pull in [this comment](https://github.com/OpenFreeEnergy/gufe/pull/794#issuecomment-4654287344) from @mattwthompson :


I'm not sure why the serialized representation changes - whether or not the _default value_ of `"nonbonded_cutoff"` is serialized might not be important, but I've seen problems[^1] in which a serialized non-default value is ignored and set to the default value[^2].

The failures due to dependencies not being installed is just from the switch to the Conda package to the development install - `cachetools`, `python-constraint`, and `xmltodict` are probably all you need

[^1]: https://github.com/openforcefield/openff-interchange/pull/1501#issuecomment-4545442911
[^2]: Bad!

src/gufe/settings/typing.py
```diff
@@ -60,7 +59,6 @@ def _is_array(quantity) -> Quantity:
 # TODO: enforce that this is *not* a list?
 GufeQuantity = Annotated[
     Quantity,
-    Field(validate_default=True),  # fail fast up front if the default isn't valid.
```
atravitz, line 63: This was not actually doing anything (only works when defining specific fields is my understanding). Pydantic 2.13 raised a warning about it.

src/gufe/tests/test_models.py
```diff
@@ -108,6 +109,7 @@ def test_openmmffsettings_schema():
             },
             "nonbonded_method": {"default": "PME", "title": "Nonbonded Method", "type": "string"},
             "nonbonded_cutoff": {
+                "default": '{"magnitude": 0.9, "units": "nanometer"}',
```
IAlibay, line 112: TBD - I wonder if we should drop the default here.

news/pydantic_2.13.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Added support for **Pydantic 2.13** and **Python 3.14**.
+
+**Changed:**
+
+* For all ``gufe.settings`` models, ``model_dump()`` uses ``polymorphic_serialization``, meaning that sub-models will be included in the output.
```
atravitz: double check whether this is user-facing change.

IAlibay (approved):

src/gufe/vendor/openff/interchange/pydantic.py
```diff
@@ -15,7 +15,7 @@ class _BaseModel(BaseModel):
     )
 
     def model_dump(self, **kwargs) -> dict[str, Any]:
-        return super().model_dump(serialize_as_any=True, **kwargs)
+        return super().model_dump(polymorphic_serialization=True, **kwargs)
 
     def model_dump_json(self, **kwargs) -> str:
-        return super().model_dump_json(serialize_as_any=True, **kwargs)
+        return super().model_dump_json(polymorphic_serialization=True, **kwargs)
```
mattwthompson, line 21: I don't recall experimenting with this method and we don't use it[^1], but if I understand the docs correctly I'm a little confused why it's necessary here but not in our code. Is it because you're subclassing and re-defining a fresh schema[^2]? The functions in that make up that annotations look pretty familiar, but if important details differ then it does make sense to subclass

[^1]: https://github.com/search?q=org%3Aopenforcefield+polymorphic_serialization&type=code)
[^2]: https://github.com/OpenFreeEnergy/gufe/blob/b29dfe45b99ae24672653467066fc5905feece6a/src/gufe/settings/typing.py#L61-L68

atravitz: From my testing, it seems that we need either `polymorphic_serialization` or `serialize_as_any` in order for our subclassed model, [OpenMMSystemGeneratorFFSettings](https://github.com/OpenFreeEnergy/gufe/blob/48f70bbff01c95bd7afcf39ea115539592960bd3/src/gufe/settings/models.py#L138) to be included in the output of `model_dump_json`. 

- Without either option, thermo_settings default is not included, e.g. `thermo_settings: {}`.  
  - This is bad.
- Using `serialize_as_any`, openff.units remain, e.g. `'thermo_settings': {'temperature': <Quantity(298, 'kelvin')>`.
  - This changes existing behavior, but might be preferred. I don't _think_ this was possible to do prior to Pydantic v2.13.
- Using `polymorphic_serialization`,  openff.units are serialized, e.g .`'temperature': {'val': 298, 'unit': 'kelvin'}`
  -  This reproduces existing behavior, but may not be preferred long-term. This is what I chose to do with the mindset of "let's get Pydantic 2.13 working while breaking as little as possible."
  
  
We originally defined a `GufeQuantity` to meet gufe's serialization requirements when moving to Pydantic v2. I vaguely remember that I had to do validation slightly differently based on gufe's serialization. 

It might be worth it at some point for me to revisit how we've vendored and adapted the openff pydantic typing logic, but my priority here was to get things working with minimal changes.

PR #826, opened 2026-08-11
Ligand mapping helper functions

jthorton:
Adding functions discussed in https://github.com/OpenFreeEnergy/openfe/pull/2122 here as they might be generally useful outside of the validation function. I think the heavy atom mapping is the most important the mapping ratio I am still not sure about. 

Tips

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: yes / no
If yes, please provide details here: No

## Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/src/gufe/mapping/ligandatommapping.py
+++ b/src/gufe/mapping/ligandatommapping.py
@@ -316,7 +316,15 @@ class LigandAtomMapping(AtomMapping):
         mol_a = self.componentA.to_rdkit()
         mol_b = self.componentB.to_rdkit()
 
-        num_unique_heavy_atoms_a = sum(1 for atom in mol_a.GetAtoms() if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.keys())
-        num_unique_heavy_atoms_b = sum(1 for atom in mol_b.GetAtoms() if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.values())
+        num_unique_heavy_atoms_a = sum(
+            1
+            for atom in mol_a.GetAtoms()
+            if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.keys()
+        )
+        num_unique_heavy_atoms_b = sum(
+            1
+            for atom in mol_b.GetAtoms()
+            if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.values()
+        )
 
         return (num_unique_heavy_atoms_a + num_unique_heavy_atoms_b) / num_mapped_heavy_atoms
--- a/src/gufe/mapping/ligandatommapping.py
+++ b/src/gufe/mapping/ligandatommapping.py
@@ -314,6 +314,11 @@ class LigandAtomMapping(AtomMapping):
         """
         heavy_atom_mapping = self.heavy_atom_componentA_to_componentB
         num_mapped_heavy_atoms = len(heavy_atom_mapping)
+
+        # quick return if there are no mapped atoms to avoid zero division errors
+        if num_mapped_heavy_atoms == 0:
+            return 0
+
         mol_a = self.componentA.to_rdkit()
         mol_b = self.componentB.to_rdkit()
 
--- a/src/gufe/mapping/ligandatommapping.py
+++ b/src/gufe/mapping/ligandatommapping.py
@@ -87,7 +87,7 @@ class LigandAtomMapping(AtomMapping):
     @property
     def heavy_atom_componentA_to_componentB(self) -> dict[int, int]:
         """
-        Return the heavy atoms mapped between componentA and componentB, heavy to hydrogen mappings are also removed.
+        Return a subset of the ``componentA_to_componentB`` mapping where the atoms in both components are heavy (non-hydrogen) atoms.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


jthorton:
pre-commit.ci autofix

src/gufe/mapping/ligandatommapping.py
```diff
@@ -278,3 +278,53 @@ def get_alchemical_charge_difference(self) -> int:
         charge_a = Chem.rdmolops.GetFormalCharge(self.componentA.to_rdkit())
         charge_b = Chem.rdmolops.GetFormalCharge(self.componentB.to_rdkit())
         return charge_a - charge_b
+
+    def get_heavy_atom_componentA_to_componentB(self) -> dict[int, int]:
```
IAlibay: Can we make this a property and move it up where all the other mapping properties are?

src/gufe/mapping/ligandatommapping.py
```diff
@@ -278,3 +278,53 @@ def get_alchemical_charge_difference(self) -> int:
...
+        -------
+        float:
+            The ratio of alchemical heavy atoms to mapped heavy atoms.
+        """
+        heavy_atom_mapping = self.get_heavy_atom_componentA_to_componentB()
+        num_mapped_heavy_atoms = len(heavy_atom_mapping)
+        mol_a = self.componentA.to_rdkit()
+        mol_b = self.componentB.to_rdkit()
+
+        num_unique_heavy_atoms_a = sum(
```
IAlibay: Sorry a little bit confused - why do we need to do all of this if we're using the heavy atom mapping?

IAlibay (changes requested):
Initial quick review.

jthorton:
pre-commit.ci autofix

IAlibay (approved):
It looks good to me, but I would like @atravitz to also have a look too - since I'm partly responsible for the suggestion to do this.

src/gufe/mapping/ligandatommapping.py
```diff
@@ -84,6 +84,30 @@ def _defaults(cls):
...
+    @property
+    def heavy_atom_componentA_to_componentB(self) -> dict[int, int]:
+        """
+        Return the heavy atoms mapped between componentA and componentB, heavy to hydrogen mappings are also removed.
+
+        Returns
+        -------
+        dict[int, int]:
+            A dictionary mapping heavy atom indices from componentA to componentB.
+        """
```
atravitz: I think a clearer docstring here would help. 

From my understanding, `heavy_atom_componentA_to_componentB` returns a _subset_ of `componentA_to_componentB`, keeping only the mapping pairs where _both_ members are heavy atoms (heavy atom defined as anything other than a hydrogen).

Is this correct?

src/gufe/mapping/ligandatommapping.py
```diff
@@ -278,3 +302,35 @@ def get_alchemical_charge_difference(self) -> int:
...
+
+        num_unique_heavy_atoms_a = sum(
+            1
+            for atom in mol_a.GetAtoms()
+            if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.keys()
+        )
+        num_unique_heavy_atoms_b = sum(
+            1
+            for atom in mol_b.GetAtoms()
+            if atom.GetAtomicNum() != 1 and atom.GetIdx() not in heavy_atom_mapping.values()
```
atravitz: If `.keys()` and `.values()` are switched in this logic, the tests still pass,  since it's a benzene -> phenol and so the heavy atom mapping is symmetric (`{0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 5: 5}`)

Is there a more complex mapping in the test suite we can use instead to be safer?

atravitz: also relevant to this chunk of code - is getting the alchemical atoms of each component something worth splitting into a separate function so it's defined clearly?

Non-blocking, and I'd recommend it start as a private function, but could be useful and make the code more legible.

atravitz (dismissed):
These functionalities seem useful! I just want to make sure the docstrings are super clear and the tests rigorous since these are immediately being exposed as public properties.

src/gufe/mapping/ligandatommapping.py
IAlibay: Not sure what you mean re: alchemical atoms (since they are all technically alchemical) - we have a method that gives us the unique atoms, which we could technically call here, is that what you mean?

src/gufe/mapping/ligandatommapping.py
jthorton: Correct, updated the docstring based on your description here!

src/gufe/mapping/ligandatommapping.py
jthorton: > If .keys() and .values() are switched in this logic, the tests still pass,

Ah nice observation, thats actually a property of the metric it should be symmetrical to reversing the mapping order so I have added a test for that, in terms of more complicated, it seems like most of the tests use beznene I can add a new set of molecules if needed though. 

For the alchemical atom helper function, we do have `comppnentA/B_unique` but we have to then filter these lists for non-hydrogen atoms and it reads similar to the current code, I have swapped to using something from rdkit which should be much cleaner.

jthorton:
pre-commit.ci autofix

src/gufe/mapping/ligandatommapping.py
jthorton: Added a more complicated non-symmetrical test!

jthorton:
pre-commit.ci autofix

src/gufe/mapping/ligandatommapping.py
IAlibay: Re: @atravitz second suggestion, @jthorton can you move it to a new issue?

PR #827, opened 2026-08-15
make file storage create directory upon instantiation

atravitz:
Makes FileStorage behave more similarly to expected Pathlib behavior.
This also enables better error handling and guardrails for  FileSystemWarehouse, as mentioned in https://github.com/OpenFreeEnergy/openfe/pull/2155#issuecomment-5297327437.

As written, this PR breaks existing behavior. Setting `exist_ok=True` by default would keep existing behavior. In my opinion, since ExternalStorage isnt used elsewhere yet, we should stick with the safer `exist_ok=False` option, but I can be convinced otherwise.

** LLM / AI generated code disclosure **

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: no
If yes, please provide details here:

## Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/src/gufe/tests/storage/test_externalresource.py
+++ b/src/gufe/tests/storage/test_externalresource.py
@@ -129,6 +129,19 @@ class TestFileStorage:
         assert set(storage) == expected
         assert set(storage) == set(storage.iter_contents())
 
+    def test_exist_ok(self, tmp_path):
+        store_dir = tmp_path / "store_dir"
+
+        _ = FileStorage(store_dir, exist_ok=False)
+        assert store_dir.is_dir()
+
+        file_store = FileStorage(store_dir, exist_ok=True)
+        assert file_store.root_dir == store_dir
+
+        with pytest.raises(FileExistsError, match=str(store_dir)):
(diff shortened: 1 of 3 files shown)


src/gufe/tests/storage/test_externalresource.py
```diff
@@ -121,7 +122,7 @@ def test_iter(self, tmp_path):
             with open(path, "wb") as f:
                 f.write(b"")
 
-        storage = FileStorage(tmp_path)
+        storage = FileStorage(tmp_path, exist_ok=True)
```
ethanholz, line 125: We need a negative test for this case as well. This should roughly be `exist_ok` is `False` but the directory exists.

ethanholz (changes requested):
This looks good to me, no problems here.

src/gufe/tests/storage/test_externalresource.py
atravitz: good point, done!

ethanholz (approved):

PR #830, opened 2026-08-27
Fix archive version parsing

### LINKED ISSUE #758: alchemical archive breaks if a letter is in the version name
**Describe the bug**

seen when releasing an alpha version (`v1.9.0.a0`), which we should be able to handle.

Thanks @mikemhenry for flagging this and suggesting https://packaging.pypa.io/en/latest/version.html#packaging.version.Version as a possible solution.

**Software versions**
  - Which operating system and version did you use? (e.g. ubuntu 22.04.5)
  - Which method did you use to install this package? (e.g. conda-forge)
  - Copy/paste the output of `conda list` (or the equivalent for your package manager):

atravitz:
resolves https://github.com/OpenFreeEnergy/gufe/issues/758

Tips

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: yes / no
If yes, please provide details here:

## Checklist
* [ ] Added a ``news`` entry (not needed, only affects tests)

### DIFF
--- a/src/gufe/tests/test_archival.py
+++ b/src/gufe/tests/test_archival.py
@@ -36,12 +36,15 @@ class TestArchival(GufeTokenizableTestsMixin):
         # fixture will generate correct version
         assert instance.version_gufe == gufe.__version__
 
+    # only warn for a difference in major or minor version,
+    # do not warn for difference in patch or dev version
     @pytest.mark.parametrize(
         "archive_ver,current_ver,should_warn",
         [
             ("1.12.0", "1.13.0", True),
             ("1.12.0", "1.12.1", False),
             ("1.12.0a0", "1.12.0", False),
+            ("1.12.0-a0", "1.12.0", False),
             ("1.12.1", "1.13", True),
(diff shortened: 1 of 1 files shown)


ianmkenney (changes requested):
Looks much better! I would add a comment on the relationship between version differences and `should_warn` above the decorator. Additionally, the issue states a version `X.Y.Z.alpha`. I think the convention is to use `X.Y.Z-meta`, so I think `1.12.0a0` would normally be written as `1.12.0-a0`.

atravitz:
> Looks much better! I would add a comment on the relationship between version differences and `should_warn` above the decorator. Additionally, the issue states a version `X.Y.Z.alpha`. I think the convention is to use `X.Y.Z-meta`, so I think `1.12.0a0` would normally be written as `1.12.0-a0`.

both good points, done! I kept `1.12.0a0` because that was the failure case we saw before, but I've included `1.12.0-a0` to cover how we _should_ do pre-release versioning.

ianmkenney (approved):

